PR #28, opened 2019-09-22
Remove "Created" property section.

msporny:
This PR removes the created property as I don't think anyone is using it. It also belongs in a DID Resolver response and not the DID Document itself.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-spec/pull/28.html" title="Last updated on Sep 26, 2019, 2:31 PM UTC (c27d3df)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-spec/28/0c41898...c27d3df.html" title="Last updated on Sep 26, 2019, 2:31 PM UTC (c27d3df)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1557,56 +1557,6 @@ considerations regarding authentication service endpoints.
 
     <section>
       <h2>
-Created (Optional)
-      </h2>
-
-      <p>
-Standard metadata for identifier records includes a timestamp of the
-original creation. The rules for including a creation timestamp are:
-      </p>
-
-      <ol start="1">
-        <li>
-A DID Document MUST have zero or one property representing a
-creation timestamp. It is RECOMMENDED to include this property.
-        </li>
-
-        <li>
-The key for this property MUST be created.
-        </li>
-
-        <li>
-The value of this key MUST be a valid XML datetime value as
-defined in section 3.3.7 of <a href="https://www.w3.org/TR/xmlschema11-2/">W3C XML Schema Definition
-Language (XSD) 1.1 Part 2: Datatypes</a> [[XMLSCHEMA11-2]].
-        </li>
-
-        <li>
-This datetime value MUST be normalized to UTC 00:00 as indicated
-by the trailing "Z".
-        </li>
-
-        <li>
-Method specifications that rely on DLTs SHOULD require time
-values that are after the known <a href="https://github.com/bitcoin/bips/blob/master/bip-0113.mediawiki">"median
-time past" (defined in Bitcoin BIP 113)</a>, when the DLT supports
-such a notion.
-        </li>
-      </ol>
-
-      <p>
-Example:
-      </p>
-
-      <pre class="example nohighlight">
-{
-  "created": "2002-10-10T17:00:00Z"
-}
-</pre>
-    </section>
-
-    <section>
-      <h2>
 Updated (Optional)
       </h2>
 


dmitrizagidulin:
Unlike the Updated timestamp, I can see arguments for including a Created timestamp in the DID itself. That could be a potentially important piece of metadata (possibly affecting trust metric calculations, etc), and it seems like the contents of the DID Document are likely to have more integrity-protection mechanisms (signatures, hashlink binding the DID to the contents of its document, etc) than would the Resolver metadata.

talltree:
I agree with @dmitrizagidulin here. This could be very useful info in a DID document if the corresponding DID method specification requires that it be validated by the DID registry. Not all DID methods may take advantage of this, but for those that do, this is useful, in part because it is integral to the DID document and not separated out in resolver metadata.

TallTed:
Why should `created` belong in Resolver response, if it's about the DID Document?

Being optional, nobody is required to include it, but I can think of many scenarios where it would be useful, regardless of whether anyone is using it today, so I don't understand the motivation to remove it.

TallTed:
Further, there must have been reasons that the property was added in the first place.

It follows that all of these should be shown to no longer be relevant, before the property should be removed. "I don't think anyone is using it" does not seem sufficient, even if we cannot easily identify any such user.

msporny:
> Why should created belong in Resolver response, if it's about the DID Document?

... because that's where information /about/ a DID Document should be placed. Putting this sort of information requires developers and resolvers to do more complex things that are error prone.

For example, a resolver should be able to determine when a DID Document was created and updated, and it should include that information in the metadata for a resolver response.

If this information is also allowed to go into a DID Document, then a resolver must cross-check the information w/ the actual ledger/registry... what happens if the times don't match up? Do you trust the DID Document or the resolver? Do you throw an error? A warning?

There should be one way to do this if we can get there... and since it's metadata /about/ the DID Document, it belongs in a resolver response, which contains the DID Document *and* metadata about the DID Document (and the query).

iherman:
(Sorry if this was discussed before in the cg days.)

>> Why should created belong in Resolver response, if it's about the DID Document?
>
> ... because that's where information /about/ a DID Document should be placed. Putting this sort of information requires developers and resolvers to do more complex things that are error prone.

I have seen this argument of @msporny in several issues now: that the content of the DID document and the metadata _about_ the DID document should be separated from one another. And I agree with this argument. However, when we say "belong in Resolver response": what does it mean? The current document has a (fairly short) section on [DID resolvers](https://w3c.github.io/did-spec/#did-resolvers) which does not include a formal specification of the response metadata. Doesn't this mean that we should have a separate set of metadata definition for the resolver, a bit like we have the definition of the HTTP response headers?

TallTed:
> If this information is also allowed to go into a DID Document, then a resolver must cross-check the information w/ the actual ledger/registry...

Well, *if* this particular DID Document is backed by ledger/registry entries. Rather than justifying your structure of where things belong, it seems that this might be an argument against ledger/registry backing; that a different verification pattern may be superior (in at least some cases).

peacekeeper:
@iherman :

> However, when we say "belong in Resolver response": what does it mean? The current document has a (fairly short) section on [DID resolvers](https://w3c.github.io/did-spec/#did-resolvers) which does not include a formal specification of the response metadata.

The DID Resolution spec (in the CCG) has a work-in-progress section on a "DID Resolution Result", which contains the DID Document plus metadata: https://w3c-ccg.github.io/did-resolution/#did-resolution-result.

iherman:
> The DID Resolution spec (in the CCG) has a work-in-progress section on a "DID Resolution Result", which contains the DID Document plus metadata: https://w3c-ccg.github.io/did-resolution/#did-resolution-result.

Thanks @peacekeeper, I did not realize that.

But if, depending where the discussion goes, it may be that some portion of that document should be folded into the spec. At the minimum, we may have to re-publish that document as a WG Note, otherwise things look a bit incomplete...

dlongley:
@dmitrizagidulin,

> 1. All DID methods will be ledger-based.
> 2. All ledgers implement timestamp tracking (definitely not the case) and crypto signatures (probably the case?)

I disagree. I think the only assumptions are that the resolution layer is separate from the DID Document. The argument that "created" and "updated" should remain would be to encourage DID methods that do not separate these concerns -- expressing all of the information in the DID Document itself.

I want to point out that the way that we've avoided the HTTP-Range-14 argument (which we should absolutely continue to do) is by deciding that you can, for most practical purposes, conflate a DID Document and a DID subject (they have the same identifier). There's a danger that we may lose this simplicity by encouraging expressing information in a way that stretches the limits of that conflation.

That being said, we may actually be talking about two different things anyway. There is a notion of when the DID Document (or DID subject) was created -- i.e., when the "profile" or "entity" that is the subject was created -- and when the DID Document was recognized as created on the DID network (if the DID method supports this). The latter is clearly, in my view, a piece of resolution information. The former need not be.

dmitrizagidulin:
(Discussion continued in separate issue, https://github.com/w3c/did-spec/issues/65 )

msporny:
Waiting on resolution to #65.

iherman:
This issue was discussed in [a meeting](https://www.w3.org/2019/did-wg/Meetings/Minutes/2019-12-10-did#section6).

- No actions or resolutions
meaning of “Created”<br/>
Brent Zundel: <a href="https://github.com/w3c/did-core/pull/28">https://github.com/w3c/did-core/pull/28</a><br/>
Drummond Reed: @selfissued RE the authority component, we have full rights to add syntax there in the DID URI scheme, but not in query parameters<br/>
Manu Sporny: <a href="https://github.com/w3c/did-core/issues/65">https://github.com/w3c/did-core/issues/65</a><br/>
Manu Sporny: wants to try to focus the discussion on what we should talk about in this WG<br/>
… there are 3 aspects. The first one is when the DID document was created an updated.<br/>
… that topic is being debated in the resolution subgroup that @markus_sabadello is leading<br/>
… so the other options are: 1) an assertion on the ledger itself about when the DID document was created/issued<br/>
Dmitri Zagidulin: there is already a field for ‘the ledger is asserting the creation date’, separate from ‘created’<br/>
Manu Sporny: and 2) the DID subject’s assertion of when the DID document was created/updated<br/>
Dmitri Zagidulin: so we’re only talking about 2<br/>
Manu Sporny: manu believes that it is the latter<br/>
Dave Longley: wouldn’t it be when the DID subject was created?<br/>
Oliver Terbu: We should also consider that some did methods do not have an easy way to assert created/updated time<br/>
Dmitri Zagidulin: we are not talking about the ‘ledger is asserting the DID was created’ date, because that’s a different parameter<br/>
… so the DID document contains: 1) the DID, 2) DID metadata, 3) resolution metadata<br/>
Jonathan Holt: For specific DID methods, you must refer to the ledger, so the only option is the author’s assertion<br/>
Markus Sabadello: Notes that “created” and “updated” correspond to two operations required by a DID method<br/>
… so it’s okay that the DID document contains the author-asserted timestamps and resolution produces the DID registry-based timestamps<br/>
Joe Andrieu: wants to comment on terminology<br/>
Oliver Terbu: some did methods are not ledger-based but there are some did methods such as did:ethr that are ledger-based but don’t require a ledger to be created. so, no information can be put into the did document for the creation time.<br/>
Dave Longley: agree with Joe … a DID document is controlled by a DID controller and says things about the DID subject<br/>
Joe Andrieu: it’s not the “DID subject” that sets the date, it’s the “DID controller”<br/>
Markus Sabadello: +1 to JoeAndrieu , we should have all said controller instead of subject<br/>
Joe Andrieu: wants to make sure we’re using the right language<br/>
Brent Zundel: seeing wide agreement that the datestamps in the DID document are asserted by the DID controller<br/>
Jonathan Holt: probably even more specific would be the DID controller software<br/>
Manu Sporny: agrees with Joe on terminology<br/>
… so the question is: security issues around using datestamps asserted by the DID controller<br/>
… so Manu would like to understand why self-asserted datestamps are important<br/>
Brent Zundel: so this moves us to the topic that if dates are DID-controller-asserted, are they needed?<br/>
Dave Longley: If the timestamps are in a DID document, they should be authoritative.<br/>
Ivan Herman: +1 to dlongley<br/>
Orie Steele: +1 to dlongley<br/>
Dave Longley: so they should be describing the DID subject, not the DID document<br/>
Oliver Terbu: -1<br/>
Dave Longley: “DID subject createdDocument date” would make more sense<br/>
Dave Longley: but still messy.<br/>
Dmitri Zagidulin: the name of the property is ambiguous, so the proposal is to change it to didDocumentCreated and didDocumentUpdated<br/>
Orie Steele: why?<br/>
Brent Zundel: so the question is now whether this is needed<br/>
Oliver Terbu: -1 because some ppl don’t create the did document at the created time<br/>
Orie Steele: what is the use case for knowing when the document was created, by looking at the document?<br/>
Oliver Terbu: some did methods don’t store the did doc on a ledger<br/>
Dmitri Zagidulin: oliver - I’m not sure I understand<br/>
Dmitri Zagidulin: oliver - the timestamp is not about the ledger. it’s only about when the document was created (the ‘created on the ledger’ or whatever is a separate field)<br/>
Jonathan Holt: asks if PGP keys in the web-of-trust are self-asserted with regard to datestamps?<br/>
Brent Zundel: is anyone a PGP key expert that can weigh in?<br/>
Orie Steele: I’m not sure anything PGP does regarding embedding meta data like this is good.<br/>
Brent Zundel: brent has only been to one PGP key party<br/>
Jonathan Holt: looking at his own PGP keys, he attests to their datestamps, so he wants to get the same info about others<br/>
Manu Sporny: PGP does allow you to assert datastamps for keys, but that does not make for good security<br/>
Orie Steele: +1 to manu, there is no solid use case.<br/>
Oliver Terbu: dmitriz - let’s have this discussion in a github issue<br/>
Daniel Buchner: Seems like added complexity for little benefit<br/>
Samuel Smith: q<br/>
Yancy Ribbens: +1 to manu<br/>
Manu Sporny: but we are not hearing a solid use case on why we would need “createdDocument” and “updatedDocument” as asserted by the DID controller<br/>
Daniel Buchner: one can make an argument for throwing tons of things in the docs, I think we should be super stingy about it<br/>
Jonathan Holt: can look at his own keys and the timestamps associated with each<br/>
… even though it is self-asserted timestamp, it is still a way of keeping order<br/>
Daniel Buchner: why can’t you save them with an added stamp post resolution, when you save it to your machine<br/>
Manu Sporny: agree with daniel – we should be very careful about the data we standardize in DID Document.<br/>
Dave Longley: agree with daniel/manu so far.<br/>
Samuel Smith: from a security perspective, the only way the DID controller can make an authoritative assertion about the DID document is to link it to a hash of the event that existed at that point in time<br/>
… any other ordering can be self-asserted, forged, etc.<br/>
… so that would be the cryptographic way of verifying the authoritative ordering<br/>
Yancy Ribbens: blockheight can determine time ordering?<br/>
Dave Longley: what is “authoritative” is up to the DID method, but agree we should be using crypto here to extend trust<br/>
Dave Longley: (when we can)<br/>
Manu Sporny: agrees with Sam. If a createdDocument property is in the DID document, it needs to be cryptographically verifiable<br/>
Orie Steele: you can always just embed a PGP key in your did document which contains this (and other) meta data.<br/>
Dave Longley: +1 to Orie<br/>
Joe Andrieu: @orie that would only cover the key, not the entire document<br/>
Dmitri Zagidulin: I’m totally fine with pushing down the ‘documentCreated’ metadata field down to method-specific land.<br/>
Manu Sporny: the case of just self-assertion of metadata about a DID document for the author can be handled in other ways<br/>
Yancy Ribbens: +1 to metadata not going in diddoc<br/>
Manu Sporny: but all of those ways are outside of a DID document<br/>
Daniel Buchner: or use Epoc seconds as your IDs in the values<br/>
Manu Sporny: so Manu is not hearing a strong use case for having these properties in a DID document<br/>
Daniel Buchner: epoch*<br/>
Manu Sporny: and believes we should take them out of the spec until there is a strong use case<br/>
Dave Longley: if you want to express when you registered a DID with some registry – you can add that to your DID doc perhaps under a “registration” property with more information about what you registered … i would think if you have anchored your DID to other blockchains you’d want something like that anyway for more than just one ledger/blockchain<br/>
Manu Sporny: daniel, yes, that would be a fine thing to do… epoch in id itself, and hopefully, asserted by the ledger.<br/>
Dmitri Zagidulin: dlongley: there’s a separate field for when it was registered, tho<br/>
Joe Andrieu: believes that it’s too early to call for lack of use cases when the conversation is just getting started<br/>
Dave Longley: dmitriz: this is about the DID controller saying something<br/>
Joe Andrieu: there are lots of examples of claims that are not directly tied to the subject<br/>
Daniel Buchner: The DID controller saying something is not substrate-authoritative<br/>
Joe Andrieu: this is the only place where the DID controller can assert these properties for purposes of redundancy and sniff tests<br/>
Orie Steele: SamSmith this is an attempt to formalize inception event<br/>
Dave Longley: so it seems the cases here are entirely around self-assertion, in which case such an “anchor” should have<br/>
Daniel Buchner: it’s speculative data, not something remotely falsifiable<br/>
Dave Longley: semantics that make it clear that its self-asserted info from the DID controller<br/>
Joe Andrieu: +1 to more specific property<br/>
Manu Sporny: +1 to daniel<br/>
Irene Adamski: @manu: unrelated to current discussion - who can I contact about more infos on the f2f in Amsterdam?<br/>
Justin Richer: the fundamental question is whether the DID document should be internally self-describing<br/>
Daniel Buchner: If it’s just an assertion by a party external to the trust-minimizing anchoring system, then there’s 0 reason we can’t make that a separate witness statement outside of the doc<br/>
Dmitri Zagidulin: great way to put it, Justin<br/>
Justin Richer: there are others seeing the DID document will be in a context that has other metadata about it<br/>
Daniel Buchner: if you want it bound to a doc revision, we could just allow a single prop for witness data that can be included via hash reference<br/>
Justin Richer: so these two views need to be reconciled<br/>
Dmitri Zagidulin: agrees with Justin_R<br/>
Joe Andrieu: @daniel, it’s a statement by the controller, just like everything else in the DID Document. All of which lack falsifiability.<br/>
Daniel Buchner: That way you can stuff a freaking DVD worth of witness data in there if you want, just via hash link<br/>
Dmitri Zagidulin: Dmitri believes that DID documents need to be self-describing<br/>
… but if it is self-asserted, then it can be kicked down to DID method specs<br/>
Daniel Buchner: Can someone argue against the hash-link approach?<br/>
Dmitri Zagidulin: (also standalone, yes! has implications on portability / exportability of the DID)<br/>
Jonathan Holt: conflating the need for a DID registry to add the security around the created or updated timestamp is a different issue<br/>
Dave Longley: a DID document has information about the DID subject … if there will be “self-describing” information, then the DID controller should state that the DID subject has a DID Document with properties X, Y, Z.<br/>
Jonathan Holt: but for a DID document to be able to standalone and self-describing is still important<br/>
… there is probably a room for a compromise, where the property is optional<br/>
Daniel Buchner: { witness: HASH_OF_N_OTHER_FIELDS/DATA }<br/>
Jonathan Holt: what jonathan currently does is using Open Timestamps<br/>
… so thinks there is room to compromise<br/>
Samuel Smith: +1 to self describing. The DID Controller needs to be able to make authoritative statements in the DID Doc. But I think that it may be method specific<br/>
Dmitri Zagidulin: +1 to manu’s proposal.<br/>
Joe Andrieu: -1<br/>
Dave Longley: we should remember that the root subject of the DID document is the DID subject<br/>
Manu Sporny: to speak to that compromise, he proposes to take it out of the core DID spec, and let method specs decide about how to use it<br/>
Yancy Ribbens: +1<br/>
Daniel Buchner: +42<br/>
Dmitri Zagidulin: daniel - that’s + too many :)<br/>
Manu Sporny: however there is already dissent, so no action on that proposal yet<br/>
Dave Longley: so something must say “DID subject has a DID document”, and introduce a new object with properties about that DID document.<br/>
Brent Zundel: closes the call with thanks to the scribes<br/>
Dave Longley: (so there’s no confusion over what the properties apply to DID subject vs. DID doc)<br/>
</details>

msporny:
Closing based on discussion at DID F2F meeting: https://www.w3.org/2019/did-wg/Meetings/Minutes/2020-01-29-did#section11

I'll open a new PR eventually to address this metadata issue.

PR #54, opened 2019-09-25
Minor, essentially editorial changes

iherman:
Having read the document I have noted some minor issues and I went ahead and changed the document...

- There were a number of normative references that were marked as informative (e.g., references to RFCs when defining the DID syntax). I have changed them to normative references. (Note that, for reasons that I do not really understand, I had to explicitly mark those sections as normative, otherwise respec was complaining. Go figure.)
- The reference to VC was still the FPWD, because it appeared as an explicit reference in the js file. I removed that, and let respec pick up the correct reference (ie, currently a Proposed Rec)
- I changed the issue references to the current repo, so that respec could make the proper link and did not complain
- Minor spelling mistakes

I hope this helps

### DIFF
--- a/common.js
+++ b/common.js
@@ -27,19 +27,6 @@ var ccg = {
       status: "FPWD",
       publisher: "Verifiable Claims Working Group"
     },
-    "VC-DATA-MODEL": {
-      title: "Verifiable Credentials Data Model",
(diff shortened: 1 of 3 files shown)


iherman:
It would be clearly better to merge it before the other PR-s to avoid merge issues...

PR #60, opened 2019-10-05
Add "key-type" DID URL matrix parameter.

peacekeeper:
Re-creating PR from CCG repo: https://github.com/w3c-ccg/did-spec/pull/193. Please consider earlier discussions there.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/60.html" title="Last updated on Nov 25, 2019, 3:08 AM UTC (e5cd766)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/60/0c41898...e5cd766.html" title="Last updated on Nov 25, 2019, 3:08 AM UTC (e5cd766)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -725,6 +725,13 @@ the DID Document that was valid for a DID at a certain time). Note: This
 parameter may not be supported by all DID methods.
             </td>
           </tr>
+<code>key-type</code>
+            </td>
+            <td>
+Identifies a set of keys from the DID Document by key type.
+            </td>
+          </tr>
+
         </tbody>
       </table>
 
--- a/index.html
+++ b/index.html
@@ -725,6 +725,8 @@ the DID Document that was valid for a DID at a certain time). Note: This
 parameter may not be supported by all DID methods.
             </td>
           </tr>
+          <tr>
+            <td>
 <code>key-type</code>
             </td>
             <td>


peacekeeper:

Description: Identifies a set of keys from the DID Document by key type.

Example: `did:example:1234;key-type=Ed25519VerificationKey2018`

dlongley:
-1 to merging as is because I don't know what the purpose of this parameter is and because it seems like it would tend to produce a set of keys whilst losing their authorized purpose (which could be dangerous). It is also "key" specific, not allowing for other types (i.e., why not "type="?).

msporny:
> did:example:1234;key-type=Ed25519VerificationKey2018

I don't understand the use case here... why can't you just write a function to go through the DID Document and fetch all of these keys? 

I'm trying to understand where we'd put a value like this -- `did:example:1234;key-type=Ed25519VerificationKey2018` ... like, we'd put it in a database table because we'd want to get the set of all `Ed25519VerificationKey2018`s, but I'm struggling to figure out why we'd put this value in a database table vs. just write a function to extract what we want from the DID Document?

OR13:
I believe the consensus on this was that its not a good idea, and we should close this PR.

msporny:
> I believe the consensus on this was that its not a good idea, and we should close this PR.

Agreed... @peacekeeper can we close this PR?

OR13:
@peacekeeper can we close this?

OR13:
@msporny @brentzundel @burnburn can we get labels on call the open PRs related to changes that have matrix parameters in them, and mark them as "pending-close"... we have a lot of PRs open... PRs should not remain open unless they are being worked on / going to be merged within a week or 2.... we are loosing the ability to focus reviews on things that need reviews...

kdenhartog:
Marked pending close after the resolution #1 on the 4/7/2020 call.

peacekeeper:
> Marked pending close after the resolution 1 on the 4/7/2020 call.

I agree this should be closed, but the reason for closing is that there has been consensus that there is no need for having `key-type` as a DID parameter in the DID URL. The resolution on the [2020-04-07 call](https://www.w3.org/2019/did-wg/Meetings/Minutes/2020-04-07-did) was about the syntax of DID parameters, which is independent of the question which DID parameters we want.

msporny:
Let's take this up on the next special topic call since we're still discussing the general area of DID Parameters.

kdenhartog:
> 
> 
> Let's take this up on the next special topic call since we're still discussing the general area of DID Parameters.

Sounds good, I'll hold off on taking any other action until I see confirmation from an editor or chair and will instead prompt others to take action if I see it becomes stale.

PR #61, opened 2019-10-05
Add "content-type" and "content-id" DID URL matrix parameters.

peacekeeper:
Re-creating PR from CCG repo: https://github.com/w3c-ccg/did-spec/pull/195. Please consider earlier discussions there.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/61.html" title="Last updated on Nov 25, 2019, 3:13 AM UTC (46b710c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/61/0c41898...46b710c.html" title="Last updated on Nov 25, 2019, 3:13 AM UTC (46b710c)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -712,6 +712,11 @@ Identifies a service from the DID Document by service id.
 Identifies a specific version of a DID Document to be resolved (the version ID
 could be sequential, or a UUID, or method-specific). Note: This parameter may
 not be supported by all DID methods.
+            </td>
+          </tr>
+
+          <tr>
+            <td>
 <code>content-type</code>
             </td>
             <td>
@@ -731,6 +736,9 @@ the DID Document that was valid for a DID at a certain time). Note: This
 parameter may not be supported by all DID methods.
             </td>
           </tr>
+
+          <tr>
+            <td>
 <code>content-id</code>
             </td>
             <td>
--- a/index.html
+++ b/index.html
@@ -712,6 +712,12 @@ Identifies a service from the DID Document by service id.
 Identifies a specific version of a DID Document to be resolved (the version ID
 could be sequential, or a UUID, or method-specific). Note: This parameter may
 not be supported by all DID methods.
+<code>content-type</code>
+            </td>
+            <td>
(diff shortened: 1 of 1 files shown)


peacekeeper:
This adds two concrete DID URL matrix parameters.

Description: At Rebooting-the-Web-of-Trust 8 in Barcelona, a use case was described by @talltree and @kenebert to use DID URL syntax for referencing objects in a DID target system that are not DID Documents. See https://github.com/WebOfTrustInfo/rwot8-barcelona/blob/master/topics-and-advance-readings/DID-Content-References.md.

Example: `did:example:1234;content-type=schema;content-id=z9y8x7w6`

dlongley:
> to use DID URL syntax for referencing objects in a DID target system that are not DID Documents.

It seems like a DID URL is the wrong tool for that -- or that those things should actually be given their own DIDs so they can be referenced directly. My understanding is that a minimal DID Document could just have an `id` (that is a DID) -- which should allow for just about anything to be expressed.

msporny:
> My understanding is that a minimal DID Document could just have an id (that is a DID) -- which should allow for just about anything to be expressed.

Right, why doesn't this solve the problem that these two matrix parameters are attempting to solve? This also raises the question, can DID Documents have alternate content-types, and if so, should that be retrieved via an HTTP API, or are we going to add the concept of content-types to DID Registries (the latter being a really scary expansion in scope).

peacekeeper:
Regarding these matrix parameters, the original idea was to be able to construct DID URLs that can identify arbitrary method- or application-specific resources that are not part of or related to the DID document. I believe there are at least two better ways of accomplishing this:

- Use the path and/or query component in DID URLs for identifying such resources.
- Assign separate, dedicated DIDs to such resources.

Therefore, unless @ken-ebert or @talltree or @brentzundel see a use case for the `content-id` and/or `content-type` matrix parameters, I propose to close this PR.

brentzundel:
I am fine with closing this PR

peacekeeper:
Closing due to community consensus.

PR #62, opened 2019-10-05
Add "service-type" DID URL matrix parameter.

peacekeeper:
Re-creating PR from CCG repo: https://github.com/w3c-ccg/did-spec/pull/191. Please consider earlier discussions there.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/62.html" title="Last updated on Nov 25, 2019, 3:10 AM UTC (d3314a9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/62/0c41898...d3314a9.html" title="Last updated on Nov 25, 2019, 3:10 AM UTC (d3314a9)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -725,6 +725,13 @@ the DID Document that was valid for a DID at a certain time). Note: This
 parameter may not be supported by all DID methods.
             </td>
           </tr>
+<code>service-type</code>
+            </td>
+            <td>
+Identifies a set of services from the DID Document by service type.
+            </td>
+          </tr>
+
         </tbody>
       </table>
 
--- a/index.html
+++ b/index.html
@@ -725,6 +725,9 @@ the DID Document that was valid for a DID at a certain time). Note: This
 parameter may not be supported by all DID methods.
             </td>
           </tr>
+
+          <tr>
+            <td>
 <code>service-type</code>
             </td>
             <td>


peacekeeper:

Description: Identifies a set of services from the DID Document by service type.

Example: `did:example:1234;service-type=agent`

dlongley:
The purpose/use case for this parameter needs to be expressed. It's unclear why referencing a set via a URL would be more useful than simply passing a plain DID to an application and allowing it to make some kind of (more) informed selection of an appropriate service after dereferencing the DID Document.

msporny:
> Example: did:example:1234;service-type=agent

Same questions as in https://github.com/w3c/did-spec/pull/60#issuecomment-540296240 ... and I'd expect that question to be true of any type-based matrix parameter.

peacekeeper:
Regarding a potential `service-type` parameter, it may be interesting to note that there is a media type `text/uri-list` ([RFC2483](https://tools.ietf.org/html/rfc2483#section-5)). In other words, it wouldn't be completely unreasonable to argue that dereferencing a URI could result in a list of other URLs.

OR13:
@peacekeeper any update on this after the F2F?

msporny:
> @peacekeeper any update on this after the F2F?

@peacekeeper is working with @jandrieu to create use cases for matrix parameters. The use cases for matrix parameters weren't clear (as in, not clearly articulated and there were a number of people that were arguing that the matrix parameter use cases could be solved via other means, like resolver parameters).

Next step is to get use cases for matrix params, which have been renamed to "DID Parameters", and then see if we need special syntax for that or if we can just use query parameters.

OR13:
This PR is stale, has conflicts and is blocked pending discussion of DID Parameters. I'd prefer to close it for now, to focus the attention of the WG on PRs that are not blocked. @peacekeeper thoughts?

kdenhartog:
Marked pending close after the resolution #1 on the 4/7/2020 call.

peacekeeper:
> Marked pending close after the resolution 1 on the 4/7/2020 call.

I agree this should be closed, but the reason for closing is that there has been consensus that there is no need for having `service-type` as a DID parameter in the DID URL. The resolution on the [2020-04-07 call](https://www.w3.org/2019/did-wg/Meetings/Minutes/2020-04-07-did) was about the syntax of DID parameters, which is independent of the question which DID parameters we want.

msporny:
Let's take this up on the next special topic call since we're still discussing the general area of DID Parameters.

Issue #46, opened 2019-09-23
It would be useful to have `services` as a mapping instead of an `array`


brentzundel:
@aaitor moved from CCG (https://github.com/w3c-ccg/did-spec/issues/224)

dlongley:
Part of this will depend on https://github.com/w3c/did-spec/issues/47 -- whether or not IDs are required, so we must wait on a decision over there. While it would be useful to have `services` be a mapping, I suspect we want to support services that do not require identifiers.

dlongley:
As #47 was merged, we should close this.

PR #66, opened 2019-10-09
Make "id" optional for services and public keys.

### LINKED ISSUE #47: Should "id" be mandatory for "service" and "publicKey"?
@peacekeeper moved from CCG (https://github.com/w3c-ccg/did-spec/issues/247)

peacekeeper:
Fixes #47 (see discussion there).

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-spec/pull/66.html" title="Last updated on Oct 18, 2019, 10:17 PM UTC (99d9e76)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-spec/66/ee57740...99d9e76.html" title="Last updated on Oct 18, 2019, 10:17 PM UTC (99d9e76)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1184,7 +1184,8 @@ Cryptographic Suite Registry</a>.
         <li>
 Each public key SHOULD include an <code>id</code> property. The array
 of public keys MUST NOT contain multiple entries with the same
-<code>id</code>.
+<code>id</code>. A DID document processor MUST produce an error in that
+case.
         </li>
 
         <li>
@@ -1484,7 +1485,8 @@ of service endpoints.
         <li>
 Each service endpoint SHOULD include an <code>id</code> property.
 The array of service endpoints MUST NOT contain multiple entries
-with the same <code>id</code>.
+with the same <code>id</code>. A DID document processor MUST
+produce an error in that case.
         </li>
 
         <li>
--- a/index.html
+++ b/index.html
@@ -1182,7 +1182,7 @@ Cryptographic Suite Registry</a>.
         </li>
 
         <li>
-Each public key MAY include an <code>id</code> property. The array
+Each public key SHOULD include an <code>id</code> property. The array
 of public keys MUST NOT contain duplicate entries with the same
 <code>id</code>.
         </li>
--- a/index.html
+++ b/index.html
@@ -1183,7 +1183,7 @@ Cryptographic Suite Registry</a>.
 
         <li>
 Each public key SHOULD include an <code>id</code> property. The array
-of public keys MUST NOT contain duplicate entries with the same
+of public keys MUST NOT contain multiple entries with the same
 <code>id</code>.
         </li>
 
@@ -1483,7 +1483,7 @@ of service endpoints.
 
         <li>
 Each service endpoint SHOULD include an <code>id</code> property.
-The array of service endpoints MUST NOT contain duplicate entries
+The array of service endpoints MUST NOT contain multiple entries
 with the same <code>id</code>.
         </li>
 
--- a/index.html
+++ b/index.html
@@ -1189,7 +1189,7 @@ of public keys MUST NOT contain multiple entries with the same
 
         <li>
 Each public key MUST include a <code>type</code> property
-and exactly one value property, and MAY include additional properties.
+and exactly one public key value, and MAY include additional properties.
         </li>
 
         <li>
--- a/index.html
+++ b/index.html
@@ -1482,7 +1482,7 @@ of service endpoints.
         </li>
 
         <li>
-Each service endpoint MAY include an <code>id</code> property.
+Each service endpoint SHOULD include an <code>id</code> property.
 The array of service endpoints MUST NOT contain duplicate entries
 with the same <code>id</code>.
         </li>
--- a/index.html
+++ b/index.html
@@ -1182,9 +1182,14 @@ Cryptographic Suite Registry</a>.
         </li>
 
         <li>
-Each public key MUST include <code>id</code> and <code>type</code>
-properties, and exactly one value property. The array of public keys
-MUST NOT contain duplicate entries with the same <code>id</code>.
+Each public key MAY include an <code>id</code> property. The array
+of public keys MUST NOT contain duplicate entries with the same
+<code>id</code>.
+        </li>
+
+        <li>
+Each public key MUST include a <code>type</code> property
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1182,9 +1182,14 @@ <h2>
         </li>
 
         <li>
-Each public key MUST include <code>id</code> and <code>type</code>
-properties, and exactly one value property. The array of public keys
-MUST NOT contain duplicate entries with the same <code>id</code>.
+Each public key MAY include an <code>id</code> property. The array
```
```suggestion
Each public key SHOULD include an <code>id</code> property. The array
```

msporny: I'm on the fence... SHOULD or MAY? Aren't identifiers for everything in a DID Document a good idea... otherwise, it becomes difficult to refer to them... so, shouldn't we default to "hey, this is a good idea to give identifiers to all of your things"... but allow people to not do it.

index.html
```diff
@@ -1182,9 +1182,14 @@ <h2>
...
+of public keys MUST NOT contain duplicate entries with the same
```
msporny: This means that DID Document processors must throw an error when they hit multiple entries with the same `id`?

index.html
```diff
@@ -1182,9 +1182,14 @@ <h2>
...
+<code>id</code>.
+        </li>
+
+        <li>
+Each public key MUST include a <code>type</code> property
+and exactly one value property, and MAY include additional properties.
```
```suggestion
and exactly one public key value, and MAY include additional properties.
```

index.html
```diff
@@ -1477,9 +1482,15 @@ <h2>
         </li>
 
         <li>
-Each service endpoint MUST include <code>id</code>,
-<code>type</code>, and <code>serviceEndpoint</code> properties, and
-MAY include additional properties.
+Each service endpoint MAY include an <code>id</code> property.
```
```suggestion
Each service endpoint SHOULD include an <code>id</code> property.
```

index.html
```diff
@@ -1477,9 +1482,15 @@ <h2>
...
+The array of service endpoints MUST NOT contain duplicate entries
```
msporny: Again, should processors throw errors if they find duplicate entries? We should probably define a conformance class here for a DID Document processor and restate this as "A DID Document Processor MUST throw an error if there are duplicate entries with the same id." ... or something to that effect.

index.html
```diff
@@ -1477,9 +1482,15 @@ <h2>
...
+with the same <code>id</code>.
+        </li>
+
+        <li>
+Each service endpoint MUST include <code>type</code> and
+<code>serviceEndpoint</code> properties, and MAY include additional
```
msporny: We need to go through the spec and talk about property-value pairs. `properties` may be confusing to readers... because they're typically the "key" part of a "key-value" pair.

index.html
peacekeeper: Agreed..

index.html
peacekeeper: I'd say yes.. Even though it wouldn't really be a JSON-LD error, it should still be thrown as a DID Document error.

index.html
peacekeeper: It's an RDF term, right? And it's used in many places throughout the spec. So how about we don't address it in this specific PR, but instead raise a new (editorial) issue to see if we can find a better term.

msporny: Not RDF... that's predicate-object (which will just confuse everyone, so we shouldn't use it). "key-value" is concerning because of the prevalence of "public keys" in the spec. We try to define things using JSON-ish terminology, but we should probably avoid the phrase "key-value" and prefer "property-value" to refer to the data model concept of an associative array... but your suggestion works for me, let's deal with it in another PR.

index.html
msporny: Agreed.

selfissued:
The JSON terminology is "member": "value".

index.html
```diff
@@ -1182,9 +1182,14 @@ <h2>
         </li>
 
         <li>
-Each public key MUST include <code>id</code> and <code>type</code>
-properties, and exactly one value property. The array of public keys
-MUST NOT contain duplicate entries with the same <code>id</code>.
+Each public key SHOULD include an <code>id</code> property. The array
+of public keys MUST NOT contain duplicate entries with the same
```
```suggestion
of public keys MUST NOT contain multiple entries with the same
```
... because *duplicate* would be redundant, not conflicting.

index.html
```diff
@@ -1477,9 +1482,15 @@ <h2>
         </li>
 
         <li>
-Each service endpoint MUST include <code>id</code>,
-<code>type</code>, and <code>serviceEndpoint</code> properties, and
-MAY include additional properties.
+Each service endpoint SHOULD include an <code>id</code> property.
+The array of service endpoints MUST NOT contain duplicate entries
```
```suggestion
The array of service endpoints MUST NOT contain multiple entries
```
... as above, because *duplicate* would be redundant, not conflicting.

index.html
peacekeeper: Yes, I'd say throw an error in this case, as above. Regarding a "DID Document Processor" conformance class, do you want to do that here in this PR, or would that first have to be introduced/defined elsewhere?

msporny: > Regarding a "DID Document Processor" conformance class, do you want to do that here in this PR, or would that first have to be introduced/defined elsewhere?

Ideally, we'd define a DID Document Processor conformance class first, merge that in, then fix this one and merge it in.

Alternatively, and I think this would be better, just raise an issue stating that we need to define a DID Document Processor conformance class, fix this PR so it throws an error, we'll merge this, and get to the DID Document Processor conformance class later. Better to get this PR merged than have it hang out there for too long.

msporny:
> The JSON terminology is "member": "value".

Ugh, so it is... I was thinking of Javascript's object properties -- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/getOwnPropertyNames

Wonder why JSON deviated from that language...

index.html
peacekeeper: Just added text about producing errors in https://github.com/w3c/did-spec/pull/66/commits/bb57e32755d621a03a9d5442d8f90786589c25d5. Also raised issue https://github.com/w3c/did-spec/issues/80.

peacekeeper:
If this is merged, I think it also makes #46 obsolete (see discussion there).

msporny:
Discussed during telecon, multiple positive reviews, suggestions applied, no objections, merging.

Issue #2, opened 2019-09-17
Rewrite Authentication section - controller is wrong


msporny:
The following section in the spec is wrong, because it conflates authentication with the ability to update the DID Document:

> Authentication is the mechanism by which the controller(s) of a DID can cryptographically prove that they are associated with that DID. See Section § 9.3 Binding of Identity . Note that Authentication is separate from Authorization because the controllers may wish to enable others to update their DID Document (for example, to assist with key recovery as discussed in Section § 9.8 Key Revocation and Recovery ) without enabling them to prove control (and thus be able to impersonate the controllers).

dlongley:
The rules for how a DID Document can be updated (or how to establish the appropriate authority to do so) are method-specific. The `authentication` property is a relationship between a subject and a set of verification methods (e.g., public keys). Its meaning is that a subject, identified by the DID, has authorized some set of verification methods (as referenced/specified via the value of the `authentication` property) for the *purpose* of authentication.

This is an important statement for "authentication verifiers" whatever they may be. It does not make any statement about what type of verifier is requiring `authentication` (e.g., it does not talk about a DID ledger using `authentication` to enable updating a DID document, it does not talk about a website, it does not talk about a point of sale device).

This statement is useful to *any* "authentication verifier" that needs to check to see if an entity that is attempting to authenticate is, in fact, presenting a valid proof of authentication. If a verifier receives some data (in some protocol-specific format) that contains a proof that was made *for the purpose of "authentication"* that says that an entity is identified by the DID, then that verifier should check to ensure that the proof can be verified using a verification method (e.g., public key) listed under `authentication` in the DID Document.

This is how "proof purposes" with DID Documents work; the *purpose* of a proof must both be stated in a given proof and it must be expressed as a property of the DID Document that establishes a relationship between the entity identified by the DID and its authorized verification methods *for that purpose*. This enables verifiers, *whatever they may be*, to verify proofs for specific purposes. It is then up to the software that accepts these proofs to decide what authority they confer.

Now, a particular DID method *may* use a simple `authentication` proof to enable a DID Document to be updated. But that is DID method specific. What is done *after* an authentication check and how sufficient that is to establish some authority is *out of scope* for this mere `authentication` relationship. Some DID methods, like Veres One (v1) do *not* allow updates by merely presenting authenticating proofs (which are, again, verified by using a verification method specified via `authentication`).

[...]

Issue #22, opened 2019-09-20
[Authentication] Is Authentication the correct mechanism for specifying the ownership of a Thing? 


brentzundel:
@mwherman2000 moved from CCG (https://github.com/w3c-ccg/did-spec/issues/148)

msporny:
This may be a duplicate of issue #2.

msporny:
This issue will be closed soon unless there is an objection to the changes made to address the issue.

msporny:
The authentication section was expanded to address the issue raised. No further requests have been made to change the section. Closing.

PR #77, opened 2019-10-17
Ensure DID Controller is used consistently. Fix #37.

### LINKED ISSUE #37: Add "DID controller" to "Terminology" section.
@peacekeeper moved from CCG (https://github.com/w3c-ccg/did-spec/issues/205)

msporny:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-spec/pull/77.html" title="Last updated on Oct 23, 2019, 1:23 AM UTC (df93787)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-spec/77/99a0873...df93787.html" title="Last updated on Oct 23, 2019, 1:23 AM UTC (df93787)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1403,14 +1403,14 @@ Authentication
       </h2>
 
       <p>
-Authentication is the mechanism by which <a>DID controller(s)</a can
+Authentication is the mechanism by which the <a>DID controller(s)</a> can
 cryptographically prove that they are associated with that DID.
 See Section <a href="#binding-of-identity"></a>. Note
-that Authentication is separate from Authorization because the 
-<a>DID controller(s)</a> may wish to enable others to update their DID Document 
+that Authentication is separate from Authorization because the
+<a>DID controller(s)</a> may wish to enable others to update their DID Document
 (for example, to assist with key recovery as discussed in Section
 <a href="#key-revocation-and-recovery"></a>) without enabling them to prove
-control (and thus be able to impersonate the <a>DID controller</a>).
+control (and thus be able to impersonate the <a>DID controller(s)</a>).
       </p>
 
       <p>
@@ -2684,7 +2684,7 @@ Privacy Considerations
     <p>
 It is critically important to apply the principles of Privacy by Design
 to all aspects of decentralized identifier architecture, because DIDs
-and DID Documents are &mdash; by design &mdash; administered directly by 
+and DID Documents are &mdash; by design &mdash; administered directly by
 the <a>DID controller(s)</a>.
 There is no registrar, hosting company, or other intermediate service
 provider to recommend or apply additional privacy safeguards. The
--- a/index.html
+++ b/index.html
@@ -1403,14 +1403,14 @@ Authentication
       </h2>
 
       <p>
-Authentication is the mechanism by which the controller(s) of a <a>DID</a> can
-cryptographically prove that they are associated with that <a>DID</a>.
+Authentication is the mechanism by which <a>DID controller(s)</a can
+cryptographically prove that they are associated with that DID.
 See Section <a href="#binding-of-identity"></a>. Note
-that Authentication is separate from Authorization because the controllers
-may wish to enable others to update their <a>DID document</a> (for
-example, to assist with key recovery as discussed in Section
+that Authentication is separate from Authorization because the 
+<a>DID controller(s)</a> may wish to enable others to update their DID Document 
+(for example, to assist with key recovery as discussed in Section
 <a href="#key-revocation-and-recovery"></a>) without enabling them to prove
-control (and thus be able to impersonate the controllers).
+control (and thus be able to impersonate the <a>DID controller</a>).
       </p>
 
       <p>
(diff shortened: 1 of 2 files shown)


msporny:
> At some point there were discussions about the difference between "controller of a DID" and "controller of a DID document".

That's true, and we still need to deal with that... waiting on @jandrieu to provide some guidance there as I thought he had the beginnings of a plan the last time this came up.

peacekeeper:
Also, in some cases this PR changes plural language "controller(s) of a DID" to singular "DID controller".. Is that a problem?

TallTed:
> Also, in some cases this PR changes plural language "controller(s) of a DID" to singular "DID controller".. Is that a problem?

I think, yes, because it will suggest to some readers that there can only be a single controller, which interpretation was intentionally avoided by including the `(s)`.

index.html
```diff
@@ -1367,14 +1367,14 @@ <h2>
       </h2>
 
       <p>
-Authentication is the mechanism by which the controller(s) of a DID can
+Authentication is the mechanism by which the <a>DID controller</a> can
```
```suggestion
Authentication is the mechanism by which the <a>DID controller(s)</a> can
```

index.html
```diff
@@ -1367,14 +1367,14 @@ <h2>
...
 cryptographically prove that they are associated with that DID.
 See Section <a href="#binding-of-identity"></a>. Note
-that Authentication is separate from Authorization because the controllers
-may wish to enable others to update their DID Document (for
-example, to assist with key recovery as discussed in Section
+that Authentication is separate from Authorization because the
+<a>DID controller</a> may wish to enable others to update their DID Document
```
```suggestion
<a>DID controller(s)</a> may wish to enable others to update their DID Document
```

index.html
```diff
@@ -1367,14 +1367,14 @@ <h2>
...
+(for example, to assist with key recovery as discussed in Section
 <a href="#key-revocation-and-recovery"></a>) without enabling them to prove
-control (and thus be able to impersonate the controllers).
+control (and thus be able to impersonate the <a>DID controller</a>).
```
```suggestion
control (and thus be able to impersonate the <a>DID controller(s)</a>).
```

index.html
```diff
@@ -2641,7 +2641,8 @@ <h1>
     <p>
 It is critically important to apply the principles of Privacy by Design
 to all aspects of decentralized identifier architecture, because DIDs
-and DID Documents are &mdash; by design &mdash; administered directly by their controllers.
+and DID Documents are &mdash; by design &mdash; administered directly by
+the <a>DID controller</a>.
```
```suggestion
the <a>DID controller(s)</a>.
```

TallTed: Github is doing something weird. The `<span ... /span>` markup doesn't belong; only the embedded `(s)`

jandrieu:
First, note that this intersects with #2 but that shouldn't keep this from merging.

Second, I'm not sure how the ReSpec rules work, but will Ted's proposed <a>controller(s)</a> still link to the controller definition?

It might be more effective to use the indefinite article "a" instead of the definite article "the", as in 
> Authentication is the mechanism by which a DID controller can cryptographically prove that they are associated with that DID

instead of 
> Authentication is the mechanism by which the DID controller can cryptographically prove that they are associated with that DID

I believe that shift will also address the plurality question, as we no longer need to specify the cardinality of controller(s).

Original
> control (and thus be able to impersonate the <a>DID controller</a>).

Ted's suggestion
> control (and thus be able to impersonate the <a>DID controller(s)</a>).

My suggestion
> control (and thus be able to impersonate a <a>DID controller</a>).

rhiaro:
> Second, I'm not sure how the ReSpec rules work, but will Ted's proposed controller(s) still link to the controller definition?

L75 of terms.html should be updated to add `controller(s)` to the matching terms so the link will work.

> It might be more effective to use the indefinite article "a" instead of the definite article "the"

Potential ambiguity where "a DID controller" could mean the controller of any DID when we want to mean the controller of the specific DID in question. Only likely to be called out by someone deliberately misunderstanding, so might be worth the tradeoff.

TallTed:
Unfortunately, in addition to @rhiaro's point about the potential ambiguity, `a` doesn't make plain that multiple controllers are possible and permitted. The tweak to terms.html is tiny, is likely to be useful in other places, and is now in PR #83.

jandrieu:
Ok. This reads fine to me, modulo the fixes we'll need to both the authentication section (controller != authenticator) and the issue of what do we call the parties involve. Let's get this in.

dlongley:
> Ok. This reads fine to me, modulo the fixes we'll need to both the authentication section (controller != authenticator) and the issue of what do we call the parties involve. Let's get this in.

Yes, once this PR is in I plan on writing up a PR that should help address `controller != authenticator` to resolve issues #2 and #22.

msporny:
Applied suggestions from @TallTed and @rhiaro which seem to address concerns raised by @peacekeeper and @jandrieu. I think we have consensus, editorial, merging.

PR #81, opened 2019-10-18
Fix DID query term definition

peacekeeper:
(looks like an earlier copy&paste mistake)

### DIFF
--- a/terms.html
+++ b/terms.html
@@ -139,7 +139,7 @@ A DID registry is a type of verifiable data registry (see [[VC-DATA-MODEL]]).
 
   <dd>
 The portion of a <a>DID URL</a> that follows the first question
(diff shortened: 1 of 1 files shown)


terms.html
```diff
@@ -139,7 +139,7 @@
 
   <dd>
 The portion of a <a>DID URL</a> that follows the first question
-mark character (<code>?</code>). A DID path uses the identical syntax as a URI path.
+mark character (<code>?</code>). A DID query uses the identical syntax as a URI query.
```
TallTed:
```suggestion
mark character (<code>?</code>). DID query syntax is identical to URI query syntax.
```

msporny:
Editorial, commenter suggestion integrated, merging.

PR #82, opened 2019-10-18
Add "DID resolver" to terminology

peacekeeper:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-spec/pull/82.html" title="Last updated on Oct 23, 2019, 2:07 AM UTC (c1dbcfe)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-spec/82/a45cd90...c1dbcfe.html" title="Last updated on Oct 23, 2019, 2:07 AM UTC (c1dbcfe)">Diff</a>

### DIFF
--- a/terms.html
+++ b/terms.html
@@ -20,7 +20,7 @@ ledger.
   </dd>
 
 
-  <dt><dfn data-lt="DID|DIDs|decentralized identifiers">decentralized identifier</dfn> (DID)</dt>
+  <dt><dfn data-lt="DID|DIDs">decentralized identifier</dfn> (DID)</dt>
 
   <dd>
 A globally unique identifier that does
@@ -76,7 +76,7 @@ a position to control the private keys. A dependent becomes an
   </dd>
 
 
-  <dt><dfn data-lt="did controller|did controllers|did controller(s)">DID controller</dfn></dt>
+  <dt><dfn data-lt="did controller(s)">DID controller</dfn></dt>
 
   <dd>
 The entity, or a group of entities, in control of a DID and/or <a>DID document</a>.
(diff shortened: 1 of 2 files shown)


terms.html
```diff
@@ -143,6 +143,13 @@
   </dd>
 
 
+  <dt><dfn data-lt="">DID resolver</dfn></dt>
```
TallTed:
```suggestion
  <dt><dfn data-lt="">DID resolver|DID resolvers</dfn></dt>
```

peacekeeper: Sorry this doesn't make sense to me, what are you trying to do here?

rhiaro: I think this was probably supposed to be inside the `data-lt` attribute

TallTed: @rhiaro - Thank you, yes.
@peacekeeper - Suggestion revised.

peacekeeper: I see, but I believe this is not necessary.. ReSpec should take care of plural forms automatically.

TallTed: ReSpec docs do suggest that it can handle (most) plurals (but apparently not optionals [i.e., `(s)`]), which means there are several others to be deleted from the terms.html doc, the presence of which others were what suggested to me the need for this one.

I don't have a horse in this race, aside from consistency within the doc.

msporny: Fixed @TallTed's concern in https://github.com/w3c/did-spec/pull/82/commits/c1dbcfe4f41680dea3fc468d81eb783d5ffd665c

msporny:
Editorial, commenters concerns addressed, multiple positive reviews, merging.

peacekeeper:
@msporny I think that last force-push messed up the html tags between the "DID subject" and "DID resolver" definitions. Also, it seems there is now a link from the term definition to the corresponding section in the spec, when I think we just decided recently NOT to do that?

msporny:
> I think that last force-push messed up the html tags between the "DID subject" and "DID resolver" definitions. Also, it seems there is now a link from the term definition to the corresponding section in the spec, when I think we just decided recently NOT to do that?

Fixed in https://github.com/w3c/did-spec/pull/87/commits/eacba50d68775d15e662216164efec10b0fa3a60 and https://github.com/w3c/did-spec/pull/87/commits/e842e72cda3ea78ba6d4c1ae825df9fee424ce33.

PR #83, opened 2019-10-21
add `did controller(s)` to matching terms

TallTed:


terms.html
```diff
@@ -72,7 +72,7 @@
 
 
 
-  <dt><dfn data-lt="did controller|did controllers">DID Controller</dfn></dt>
+  <dt><dfn data-lt="did controller|did controllers|did controller(s)">DID Controller</dfn></dt>
```
peacekeeper:
```suggestion
  <dt><dfn data-lt="did controller|did controllers|did controller(s)">DID controller</dfn></dt>
```

Issue #5, opened 2019-09-18
Where will the DID contexts(s) live?


dmitrizagidulin:
Will the eventual `https://www.w3.org/2019/did/v1` and the current `https://w3id.org/did/v0.11` context for DIDs be version controlled in this repository now, or will it continue to reside at the CCG repo?

If the former, I would propose to pull over [PR #272 - Add a contexts/README.md to explain the DID context versioning scheme](https://github.com/w3c-ccg/did-spec/pull/272) over to this repo as well, since the topic of `@context` versioning may be confusing to the wider audience.

OR13:
I recommend that both the human documentation and the JSON-LD contexts be hosted under this repo, and that we minimize use of / or links to 3rd party services which may have the ability to alter documentation or contexts in ways that impact usability or security.

It will also be easer to ensure documentation is up to date if a single PR can fix both the context and the human readable content.

Understanding of context processing is a requirement for understanding the security and usability of DIDs, I'm very much in favor of centralizing that trust here, I think many people don't realize how awesome JSON-LD is, and since DID Documents are JSON-LD, we have a responsibility to be up front about the security implications of working with JSON-LD.

iherman:
My question/comment is on the URL used for the context file. At the moment, it is`https://www.w3.org/2019/did/v1`. I do not know whether, by now, this is cast in concrete; in many cases such context files (and possible namespace documents if separate terms are to be properly defined in RDF, too) are put under `/ns` at W3C. Like, for example, `http://www.w3.org/ns/anno.jsonld` for annotations or `https://www.w3.org/ns/pub-context` for publication manifests. (I know that the VC context file is under `/2018/`, though).

If this is not yet too late, I would propose to move the context file to `https://www.w3.org/ns/did/v1`.

Whichever way we go, setting up a redirect from that URL to a github repo file shouldn't be a problem. This is how, for example, `https://www.w3.org/ns/pub-context` is managed (although in that case, the context file is managed in a separate repo on github).

dmitrizagidulin:
As I've tried to argue in the [DID context versioning PR](https://github.com/w3c-ccg/did-spec/pull/272/files), versioning the context is highly desirable (and even necessary, in our situation).

OR13:
without versioning, how can we offline these contexts? won't that lead to lots of confusion over which offline context was used? 

I'm in favor of datespace over namespace, appears to add more useful information than `ns`.

iherman:
[...]

PR #96, opened 2019-11-05
Update @context URI

rhiaro:
On the 2019-11-05 call it was resolved to use `https://www.w3.org/ns/did/v1` as the @context URI.

See also #5

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/96.html" title="Last updated on Nov 5, 2019, 5:59 PM UTC (92c7b10)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/96/d612653...92c7b10.html" title="Last updated on Nov 5, 2019, 5:59 PM UTC (92c7b10)">Diff</a>

### DIFF
--- a/FPWD/2019-11-07/index.html
+++ b/FPWD/2019-11-07/index.html
@@ -1168,7 +1168,7 @@ used to interact with the entity.
           <div class="marker">
       <a class="self-link" href="#example-2-minimal-self-managed-did-document">Example<bdi> 2</bdi></a><span class="example-title">: Minimal self-managed DID document</span>
     </div> <pre class="nohighlight">{
-  "@context": "https://www.w3.org/2019/did/v1",
+  "@context": "https://www.w3.org/ns/did/v1",
   "id": "did:example:123456789abcdefghi",
   "authentication": [{
     <span class="comment">// used to authenticate as did:...fghi</span>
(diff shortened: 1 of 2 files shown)


peacekeeper:
Quick note that there's an objection to this (see https://github.com/w3c/did-core/issues/5#issuecomment-550297895), so merging may have been premature, and depending on the outcome we may have to revert this.

iherman:
The FPWD goes out tomorrow and the version on /TR is now set with `/ns/did`. This _MUST_ be changed before this evening my time if we want another value!

@burnburn @brentzundel

burnburn:
> The FPWD goes out tomorrow and the version on /TR is now set with /ns/did. This MUST be changed before this evening my time if we want another value!

It is true that the publication cannot go out with this change as long as we have a standing objection.  @iherman , if you do not see a clear resolution of Manu's objection before evening your time, and the change is not reverted or marked as still under discussion, then you must halt publication tomorrow.

iherman:
I am not sure whether this objection is against the URL or against the FPWD. In theory, there is no problem changing that particular technical content of the core spec in a later draft.

burnburn:
> I am not sure whether this objection is against the URL or against the FPWD. In theory, there is no problem changing that particular technical content of the core spec in a later draft.

I personally agree with you.  However, this PR merge has been objected to and MUST NOT go out in the FPWD if Manu has not released his objection.  In that case, if the change is not reverted or marked by the editors as still under discussion, by evening your time (can you give us a precise time?), then you must halt publication tomorrow.  If you disagree please contact the chairs directly immediately.

iherman:
It is not my place to agree or disagree on this, it is up to the chairs:-)

I would like to call it a day at 7pm my time at the latest, knowing that I also have a call between 5pm and 6pm.

burnburn:
> It is not my place to agree or disagree on this, it is up to the chairs:-)

I am glad you do not disagree :)

PR #99, opened 2019-11-08
Initial setup for echidna

iherman:
Setting up echina (=automatic publication).

Beware: merging this means that, after this step _all_ merge to the master branch will start the automatic process to publish on `https://www.w3.org/TR/did-core/`.

(Good if we are all around when it happens the first time; unexpected problems may occur with an echidna setup :-)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,17 @@
+language: node_js
+ 
+branches:
+  only:
+    - master
+ 
+env:
+  global:
+    - URL="https://w3c.github.io/did-core/ECHIDNA"
+    - DECISION="https://www.w3.org/2019/did-wg/Meetings/Minutes/2019-10-22-did#resolution1"
+    - secure: "XB1q1xJNjGnW77S9dQTXDqlZNLUDYLQeotLMnrdIxTx9uwRrVWTCDZ6WmUKjJ3dWKeoU9YF2aL8V9AK8DejDsoYp9NVBk4Vyym6R3o62v2k4lClUo9sDuNwSTBHw6mmIdzXL4ihjwyGaIhkyGWb0EFNTPu31m1PEAoE0kjeLM2aRXYOZD0p2Zo4oStz+uHNk+6Za8aLMLq+zP/XYksWuNr0FqdHFddsdFhSMzEEVpjnzWnP+ix5+mOPsoYJgPJ0S2NSa7q+wHCtI8Dr0pyK1CiMdnGR18z3W67MftaLiLiRSVj7JqxninFRViu/4ZIyqG2tr34Yo2SC3E4CCECNYQ/XktKeu02q6pS87XRuHCgmrBcMkmDM6rFwicBBceq49zXFp46GPbrJuVQCMM8tVgMRzu5llsDyyQRr2QHbOSfxN15lmacygEegjHBro9iWv+181FGeXxU42ffL8RTq6JH3ZW256sFZVNszdnUo/EXM5x4ZKXVLFd7WmK32bR+2/ZO1VhMyjTSCIGX7qi0FKj373TwfuJQQG/IFOueR+K56DaY7+v+qFQq48ueyuZZvLOp665L5Z0oTR1BAu3NONQfICHl9BoA5rEywNjobuQ2+ZpgE2hAiA9l58/8zlklG/JtoaEq3xIZq7/O2r7lmq8yuFVsJKmDVRDjojo7831zE="
+ 
+script:
+  - echo "ok"
+ 
+after_success:
+  - test $TRAVIS_PULL_REQUEST = false && curl "https://labs.w3.org/echidna/api/request" --data "url=$URL" --data "decision=$DECISION" --data "token=$TOKEN"
--- a/ECHIDNA
+++ b/ECHIDNA
@@ -1,0 +1,4 @@
+# ECHIDNA configuration
+index.html?specStatus=WD&shortName=did-core respec
+contexts/did-v0.11.jsonld
+contexts/did-v1.jsonld


.travis.yml
```diff
@@ -0,0 +1,17 @@
...
+ 
+branches:
+  only:
+    - master
+ 
+env:
+  global:
+    - URL="https://w3c.github.io/did-core/ECHIDNA"
+    - DECISION="https://www.w3.org/2019/did-wg/Meetings/Minutes/2019-10-22-did#resolution1"
+    - secure: "XB1q1xJNjGnW77S9dQTXDqlZNLUDYLQeotLMnrdIxTx9uwRrVWTCDZ6WmUKjJ3dWKeoU9YF2aL8V9AK8DejDsoYp9NVBk4Vyym6R3o62v2k4lClUo9sDuNwSTBHw6mmIdzXL4ihjwyGaIhkyGWb0EFNTPu31m1PEAoE0kjeLM2aRXYOZD0p2Zo4oStz+uHNk+6Za8aLMLq+zP/XYksWuNr0FqdHFddsdFhSMzEEVpjnzWnP+ix5+mOPsoYJgPJ0S2NSa7q+wHCtI8Dr0pyK1CiMdnGR18z3W67MftaLiLiRSVj7JqxninFRViu/4ZIyqG2tr34Yo2SC3E4CCECNYQ/XktKeu02q6pS87XRuHCgmrBcMkmDM6rFwicBBceq49zXFp46GPbrJuVQCMM8tVgMRzu5llsDyyQRr2QHbOSfxN15lmacygEegjHBro9iWv+181FGeXxU42ffL8RTq6JH3ZW256sFZVNszdnUo/EXM5x4ZKXVLFd7WmK32bR+2/ZO1VhMyjTSCIGX7qi0FKj373TwfuJQQG/IFOueR+K56DaY7+v+qFQq48ueyuZZvLOp665L5Z0oTR1BAu3NONQfICHl9BoA5rEywNjobuQ2+ZpgE2hAiA9l58/8zlklG/JtoaEq3xIZq7/O2r7lmq8yuFVsJKmDVRDjojo7831zE="
```
msporny: Err, I don't know how this file works, but it looks like we're going to commit some kind of secret to the repository? Can you use this secret to do damage to W3C servers?

msporny: I guess my question is... can I take this "secure" value and attack anyone's servers with it?

iherman: To be honest, I do not know the details; maybe @deniak knows. 

What I know: this 'secure' is, I believe, like a signature of a data which combines a secure token I have got from github as an administrator, plus a secure token that I have received for this specific repository and echidna process, all generated by travis while I log in _in that specific repository_. If and when we set up the same echidna process for, say, the use cases (which I presume we will) then I will have to ask for a separate token for that repository+echidna and regenerate a 'secure' by being logged into that repository, yielding a different 'secure'. 

Bottom line is: I do not think so...

iherman:
Just a side remark: this setup of echidna will trigger a publication process _every time_ a merge of a PR occurs. This means that if several PR-s are merged on the same day, the process will start several times that day.

This may be overkill, but it has a more disagreeable consequence: the "previous" version in the document's header may become meaningless, because it will refer on itself (TR long URI-s at W3C are based on date, not full time).

@deniak is this an issue that we have to be mindful about? Also, does echidna and respec take care of setting the previous publication date, or is it something the editors have to do in their respective PR-s?

All in all, this approach may need some extra workflow control for editors...

msporny:
> This means that if several PR-s are merged on the same day, the process will start several times that day.

I've been not saying anything thinking that this might not be an issue.

How about this instead: Can we auto-publish on a particular day of the week? Like Monday at 05:00 GMT? Our current and typical edit cycle would have most PRs from the previous weeks discussion merged by that time, allowing for a fairly stable view on the latest WD of the spec. This would also avoid the issue you mention above, @iherman.

iherman:
> How about this instead: Can we auto-publish on a particular day of the week? Like Monday at 05:00 GMT? Our current and typical edit cycle would have most PRs from the previous weeks discussion merged by that time, allowing for a fairly stable view on the latest WD of the spec. This would also avoid the issue you mention above, @iherman.

@msporny this is certainly much better and indeed solves the issue. And I guess it might be possible; it all depends on some clever hacking with travis.  However, I have no practice in this at all, ie, I do not know how to do that. If someone can come up with the right modification of our setup, that can work (@deniak, is there an example for this?)

I believe the extra action of adding the previous publication's date, after each publication, into the respec structure of `index.html` is unavoidable. That seems to be orthogonal to the question of how the publishing workflow works. But this means that some human intervention on the repo for each publication is necessary; the publication is never 100% invisible to the editors.

(I must admit that my preferred version would be different. Instead of automatically publishing at every commit to 'master', we set up a separate branch called, say, 'publish'. Any time we merge+commit into 'publish' echidna starts up on the content of that branch. We do not have to have anything special for this branch, it can publish from the respec source, etc; the only difference is that the editors (or the chairs, for that matter) have to manually trigger publication when they feel like doing it by doing a merge+commit. But, as I said, the publication is never 100% invisible anyway, so the difference for the editors seems to be pretty minor.)

.travis.yml
deniak: The `secure` variable is the [encrypted token](https://github.com/w3c/echidna/wiki/Token-creation) Echidna uses to make sure the request is safe. Only travis can decrypt it.
Also, the token is linked to a specific GitHub repository, so even if someone can get the token, he will only be able to publish what's in that repository.

deniak:
You don't need to worry about `previous versions` and same day publications. Since you are submitting a respec document (without specifying any date), the generator will take care of checking what the previous version really is.
If you are publishing multiple times a day, the previous version will remain the same as we are using dated URLs. Basically, the links in the header of the document will not change. Only its content will be updated.

The result is that if someone looks at the document in the morning, it may not be the same later, but we have always allowed editors to edit theirs documents on the day of the publication. It wasn't common as it's cumbersome to generate a new version of the document and go through the webmaster but with Echidna, it's easier if you rely on travis.

iherman:
> You don't need to worry about previous versions and same day publications. Since you are submitting a respec document (without specifying any date), the generator will take care of checking what the previous version really is.
If you are publishing multiple times a day, the previous version will remain the same as we are using dated URLs. Basically, the links in the header of the document will not change. Only its content will be updated.
>
> The result is that if someone looks at the document in the morning, it may not be the same later, but we have always allowed editors to edit theirs documents on the day of the publication. It wasn't common as it's cumbersome to generate a new version of the document and go through the webmaster but with Echidna, it's easier if you rely on travis.

@deniak I discover new things every day! This means that my worries were unfounded. I.e., the current setup will work without problems (modulo respec issues).  I am happy to stand corrected:-)

.travis.yml
msporny: Thank you @deniak, that answers my question. We're good to go here.

iherman:
@msporny just to avoid any situation when we wait on one another: I do not intend to be the one merging this PR. It should be done when all editors feel comfortable that the core is ready for echidna, so I would prefer one of you to take this decision when you feel like it...

talltree:
This is not an area where I have any expertise, so I defer to @msporny and @peacekeeper on this one.

msporny:
@iherman: 
> @msporny just to avoid any situation when we wait on one another: I do not intend to be the one merging this PR.

@talltree wrote:
> This is not an area where I have any expertise, so I defer to @msporny and @peacekeeper on this one.

I think we have enough review on this now, the configuration isn't set in stone, we can update it later. I'd like to see if we can configure Travis to do weekly releases instead of for every commit, but we can deal w/ that in a different PR (or we may just need to configure Travis directly).

Merging.

iherman:
@msporny I hope it will work!

If there are problems, @deniak is a super-helpful guy :-)

deniak:
You are now all set. From now on, each commit to the master branch will trigger the publication of a WD, e.g. https://lists.w3.org/Archives/Public/public-tr-notifications/2019Nov/0028.html
You'll just have to deactivate travis once you reach CR.

iherman:
@msporny @talltree @peacekeeper : actually, the merge itself triggered a publication; so https://www.w3.org/TR/did-core/ is now the latest version, dated November 13...

talltree:
@iherman Thanks, you are a master!

iherman:
@talltree I am certainly not... I was just copy pasting!

Issue #67, opened 2019-10-10
Supported public key formats?


msporny:
@selfissued wrote:
> We should be removing public key formats from the spec - not adding them. We should get down to one key format to increase interoperability of implementations - preferably JWK.

We should decide (as soon as possible) whether we will be supporting exactly one or more than one key expression format in the DID specification.

Options include: JWK, COSE Key, LD Keys, PEM, or expression of key-like data structures (for things like `ethereumAddress`).

We will need at least @awoie, @selfissued, @gjgd, @dlongley, @ChristopherA, and @csuwildcat to participate in this discussion. 

Related to PRs #56, #63 and issues #55, #24, #20, #15, and #7.

msporny:
We should start the issue off w/ examples of how various types of keys can be expressed using the options above.

selfissued:
For us to have a well-informed discussion on this topic, it would be useful to have an inventory of the different kinds of public keys that we will want to be able represent.  I'm aware of (1) RSA public keys, which need a mantissa and an exponent, (2) elliptic curve keys that need an X and Y pair, and (3) elliptic curve keys that only require an X value (all of which have standard JWK representations).

Are there different kinds of public keys we care about that aren't of any of the kinds 1-3 above?  If so, what values do their representations contain?  (For extra credit, please provide specification references so that we can all get up to speed them :-) .)

ChristopherA:
> (2) elliptic curve keys that need an X and Y pair, and (3) elliptic curve keys that only require an X value (all of which have standard JWK representations)

No, X and Y is not correct. What is needed is the finite field form (in the case of bitcoin's secp256k1 it is Galois y² = x³ + 7), the finite field's modulus (usually called p, which in secp25k1 is p = 2^256 – 2^32 – 977 which makes this a Kobliz Galois field), and a generator base point (usually called g, which in secp256k1 is TWO numbers, 0279BE667EF9DCBBAC55A06295CE870B07029BFCDB2DCE28D959F2815B16F81798, or in “compressed form” 040x79BE667EF9DCBBAC55A06295CE870B07029BFCDB2DCE28D959F2815B16F81798483ADA7726A3C4655DA4FBFC0E1108A8FD17B448A68554199C47D08FFB10D4B8). 

[...]

PR #100, opened 2019-11-10
Add public key value expression goals. Add key expression formats for RSA, secp256k1, secp256r1, ed25519, and curve25519.

msporny:
This PR attempts to partially resolve issue #67 by stating that public key formats will be limited. It also adds a suggestion for how RSA keys are expressed.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/100.html" title="Last updated on Dec 10, 2019, 7:53 PM UTC (5287f42)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/100/02b8613...5287f42.html" title="Last updated on Dec 10, 2019, 7:53 PM UTC (5287f42)">Diff</a>

### DIFF
--- a/common.js
+++ b/common.js
@@ -116,6 +116,17 @@ var ccg = {
       status: "Internet-Draft",
       publisher: "IETF"
     },
+    "BASE58": {
+      title: "The Base58 Encoding Scheme",
+      date: "December 2019",
+      href: "https://tools.ietf.org/html/draft-msporny-base58",
+      authors: [
+        "Satoshi Nakamoto",
+        "Manu Sporny"
+      ],
+      status: "Internet-Draft",
+      publisher: "IETF"
+    },
     "DNS-DID": {
       title: "The Decentralized Identifier (DID) in the DNS",
       date: "February 2019",
--- a/common.js
+++ b/common.js
@@ -121,7 +121,6 @@ var ccg = {
       date: "December 2019",
       href: "https://tools.ietf.org/html/draft-msporny-base58",
       authors: [
-        "Satoshi Nakamoto",
         "Manu Sporny"
       ],
       status: "Internet-Draft",
--- a/index.html
+++ b/index.html
@@ -1389,11 +1389,9 @@ the <code>publicKeyBase58</code> property.
 ed25519
             </td>
             <td>
-Ed25519 public key values MUST be encoded in either JSON Web Key (JWK)
-format using the <code>publicKeyJwk</code> property or as the raw 32-byte
-public key value encoded in Base58 Bitcoin format using the
-<code>publicKeyBase58</code> property. Other Ed25519 public key formats
-MUST NOT be used.
+Ed25519 public key values MUST be encoded as the raw 32-byte public key value
+encoded in Base58 Bitcoin format using the <code>publicKeyBase58</code> property
+if they are not encoded in JWK format.
             </td>
           </tr>
           <tr>
--- a/index.html
+++ b/index.html
@@ -1340,7 +1340,9 @@ detail how revocation is performed and tracked.
       </p>
 
       <p>
-The following table summarizes the key formats supported by this specification:
+All public key formats MUST be expressed in either JSON Web Key (JWK)
+format using the <code>publicKeyJwk</code> property or one of the native
+formats listed in the table below. All other key formats MUST NOT be used.
       </p>
 
       <table class="simple">
@@ -1361,10 +1363,9 @@ Support
 RSA
             </td>
             <td>
-RSA public key values MUST be encoded in either JSON Web Key (JWK)
-format using the <code>publicKeyJwk</code> property or Privacy Enhanced Mail
-(PEM) format using the <code>publicKeyPem</code> property. Other RSA public
-key formats MUST NOT be used.
+RSA public key values MUST be encoded in Privacy Enhanced Mail
+(PEM) format using the <code>publicKeyPem</code> property if they are not
+encoded in JWK format.
             </td>
           </tr>
         </tbody>
--- a/index.html
+++ b/index.html
@@ -1330,24 +1330,10 @@ All public key properties MUST be from the Linked Data Cryptographic Suite Regis
 A registry of key types and formats is available in Appendix <a href="#registries"></a>
       </p>
 
-      <p class="note">
-        The following is a non-exhaustive list of public key
-        properties used by the community:
-        <code>publicKeyPem</code>, <code>publicKeyJwk</code>,
-        <code>publicKeyHex</code>, <code>publicKeyBase64</code>,
-        <code>publicKeyBase58</code>, <code>publicKeyMultibase</code>,
-        <code>ethereumAddress</code>.
-      </p>
 
       <p>
-If a public key does not exist in the <a>DID document</a>, it MUST be
-assumed the key has been revoked or is invalid. The <a>DID document</a> MAY
-contain revoked keys. A <a>DID document</a> that contains a revoked key MUST
-also contain or refer to the revocation information for the key (e.g.,
-a revocation list). Each <a>DID method</a> specification is expected to
-detail how revocation is performed and tracked.
+The following table summarizes the key formats supported by this specification:
       </p>
-
       <p>
 Example:
       </p>
--- a/index.html
+++ b/index.html
@@ -1409,7 +1409,7 @@ the raw 32-byte public key value encoded in Base58 Bitcoin format using the
 Curve25519
             </td>
             <td>
-Curve25519 public key values MUST either be encoded as a JWK or be encoded as
+Curve25519 (also known as X25519) public key values MUST either be encoded as a JWK or be encoded as
 the raw 32-byte public key value in Base58 Bitcoin format using the
 <code>publicKeyBase58</code> property.
             </td>
--- a/index.html
+++ b/index.html
@@ -664,7 +664,7 @@ that depend on the type of key it is. For more information, see Section
 
       <p>
 This specification strives to limit the number of formats for expressing public
-key material in a DID Document to the fewest possible, to increase  the
+key material in a DID Document to the fewest possible, to increase the
 likelihood of interoperability. The fewer formats that implementers have to
 implement, the more likely it will be that they will support all of them. This
 approach attempts to strike a delicate balance between ease of implementation
--- a/index.html
+++ b/index.html
@@ -1352,11 +1352,10 @@ Support
 RSA
             </td>
             <td>
-RSA public key values MAY be encoded in Privacy Enhanced Mail (PEM) format using
-the <code>publicKeyPem</code> property or MAY be encoded in JSON Web Key (JWK)
-format using the <code>publicKeyJwk</code> property. At least PEM format or
-JWK format MUST be supported by DID Document Processors. The use of other
-key formats for RSA public keys MUST NOT be used.
+RSA public key values MUST be encoded in either JSON Web Key (JWK)
+format using the <code>publicKeyJwk</code> property or Privacy Enhanced Mail
+(PEM) format using the <code>publicKeyPem</code> property. Other RSA public
+key formats MUST NOT be used.
             </td>
           </tr>
         </tbody>
--- a/index.html
+++ b/index.html
@@ -1345,6 +1345,13 @@ format using the <code>publicKeyJwk</code> property or one of the native
 formats listed in the table below. All other key formats MUST NOT be used.
       </p>
 
+      <p class="issue">
+The Working Group is still debating whether the default base encoding format
+will be Base58 (Bitcoin) or base64url. The entries in the table below
+currently assume Base58 (Bitcoin), but may change to base64url once the
+group achieves consensus on this particular issue.
+      </p>
+
       <table class="simple">
         <thead>
           <tr>
--- a/index.html
+++ b/index.html
@@ -1347,7 +1347,7 @@ formats listed in the table below. Public key expression MUST NOT use any other 
 
       <p class="issue">
 The Working Group is still debating whether the base encoding format used
-will be Base58 (Bitcoin) [[BASE58]], base64url [[RFC4648]] or base16 (hex) [[RFC4648]]. The entries in the
+will be Base58 (Bitcoin) [[BASE58]], base64url [[RFC7515]] or base16 (hex) [[RFC4648]]. The entries in the
 table below currently assume PEM and Base58 (Bitcoin), but may change to base64url and/or base16 (hex) once
 the group achieves consensus on this particular issue.
       </p>
--- a/index.html
+++ b/index.html
@@ -1379,12 +1379,11 @@ Privacy Enhanced Mail (PEM) format using the <code>publicKeyPem</code> property.
 secp256k1
             </td>
             <td>
-Secp256k1 key values MAY be expressed as the raw 32-byte value encoded in
-base58-btc format using the <code>publicKeyBase58</code> property or
-MAY be encoded in JSON Web Key (JWK) format using the <code>publicKeyJwk</code>
-property. Both the raw 32-byte base58-btc format and JWK format MUST be
-supported by DID Document Processors. The use of other key formats for RSA MUST
-cause the DID Document Processor to raise an error.
+Secp256k1 public key values MUST be encoded in either JSON Web Key (JWK)
+format using the <code>publicKeyJwk</code> property or as the raw 32-byte
+public key value encoded in Base58 Bitcoin format using the
+<code>publicKeyBase58</code> property. Other Secp256k1 public key formats
+MUST NOT be used.
             </td>
           </tr>
         </tbody>
--- a/index.html
+++ b/index.html
@@ -1399,9 +1399,34 @@ raw 32-byte public key value in Base58 Bitcoin format using the
 secp256r1
             </td>
             <td>
-Secp256r1 public key values MUST be encoded as the raw 32-byte public key value
-encoded in Base58 Bitcoin format using the <code>publicKeyBase58</code> property
-if they are not encoded in JWK format.
+Secp256r1 public key values MUST either be encoded as a JWK or be encoded as
+the raw 32-byte public key value encoded in Base58 Bitcoin format using the
+<code>publicKeyBase58</code> property.
+            </td>
+          </tr>
+          <tr>
+            <td>
+ed25519
+            </td>
+            <td>
+Ed25519 public key values MUST be encoded in either JSON Web Key (JWK)
+format using the <code>publicKeyJwk</code> property or as the raw 32-byte
+public key value encoded in Base58 Bitcoin format using the
+<code>publicKeyBase58</code> property. Other Ed25519 public key formats
+MUST NOT be used.
+            </td>
+          </tr>
+          <tr>
+            <td>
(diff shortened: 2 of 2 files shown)


OR13:
I'd rather not see us attempt to formalize key encoding on a per key type basis, as we are appearing to do for RSA. IMO, the biggest interop win we could get here would be to agree on a single MUST support key encoding that would apply to all key types, and I'd like to see language of the following format:

For all public key types, a DID Document processor must not error when handling a key of type. 

Alternate encodings for specific key types MAY be supported on a case by case basis, so long as they are described in the spec, for example PEM encoding of RSA keys.

selfissued:
I completely agree with @OR13 's suggestion.  It's simple and results in a straightforward spec that implementers are likely to get right, maximizing interoperability!

msporny:
> IMO, the biggest interop win we could get here would be to agree on a single MUST support key encoding

I think this suggestion flies in the face of what has been deployed over the past decade. The reality is that there is no *one* supported key encoding format today. There have been attempts to get alignment (JWK), but those attempts haven't resulted in what we'd like to see (which is one supported key encoding format across the entire public key expression ecosystem.

I think @kdenhartog's point here cuts to the heart of the matter:

https://github.com/w3c/did-core/pull/102/files#r344938209

> I'm not sure if anyone is currently using JWKs as a representation right now for ed25519 keys, (legacy implementations appear to use base58) but I read this as implementers of future DID registries are encouraged to use JWKs rather than base58.

In other words, by mandating one key format, we're not reflecting reality... and specs should reflect reality. My understanding is that we were headed towards supporting multiple key formats. The latest arguments seem to be backpedaling to one key format (yes, that would be ideal... but reality states that this has not been happening over the past decade). Some data points to consider:

1. RSA libraries almost universally understand PEM encoding for public keys.
2. Ed25519 libraries almost universally understand raw bytes for public keys (with base58 encoding).
3. secp256k1 libraries almost universally understand raw bytes for public keys (with hex encoding).

If we are going to mandate the use of JWK, we're going to have to see if implementers are actually going to implement that solution. I suggest that, given the current trajectory of the conversation, that we get some real data from as many of the cryptographic library implementations as possible. Which libraries support which formats today, and use that to guide the discussion. Bonus points if we can get implementers of those libraries to solidly commit to supporting a single key format (like JWK).

OR13:
I imagine a survey will turn up JWK and PEM formats are the most popular standards for representing multiple key formats, because of OAuth/OIDC and OpenSSL... I'm not sure we should lean on what DID Method implementers have decided, over what existing libraries support in the wild, bigger boats normally have right of way.

I'm not sure what format the survey should take, but I'm always in favor of data driven decision making.

Are we specifically focusing on web based cryptographic libraries?

Just to be clear, I'd be ok picking PEM as the MUST support key format that all DID Document processors would not error on.

To reiterate, I'm not suggesting that there MUST be only one key format.

I'm in favor of there being a single key format that DID Document processors will not error on for all key types, either JWK or PEM.

msporny:
> I imagine a survey will turn up JWK and PEM formats are the most popular standards for representing multiple key formats, because of OAuth/OIDC and OpenSSL... I'm not sure we should lean on what DID Method implementers have decided, over what existing libraries support in the wild, bigger boats normally have right of way.

We're just gathering data now to determine if our hunches are accurate. Let's do some science... gather data, then analyze the data. There some questionable assertions flying around this discussion, so having a common data set to analyze might help us navigate the discussion more easily.

> I'm not sure what format the survey should take, but I'm always in favor of data driven decision making.

First attempt at data gathering is here:

https://github.com/w3c/did-core/issues/67#issuecomment-552947629

> Are we specifically focusing on web based cryptographic libraries?

No, all base cryptographic libraries... not just web-based ones. Fundamentally, the best approach might be to enable people to express keys such that they can be directly consumed using existing software libraries (instead of having to write a conversion layer... and messing that up).

> Just to be clear, I'd be ok picking PEM as the MUST support key format that all DID Document processors would not error on.

Let's see what the data shows... I expect that PEM/DER is the most broadly supported format in base crypto libraries, but I don't know (because it doesn't look like anyone has done a recent survey in the last 3 years or so).

> To reiterate, I'm not suggesting that there MUST be only one key format.

Good, glad to hear... because I think that would put the group into a "split into pro/anti camps"... and we'll deadlock quickly.

> I'm in favor of there being a single key format that DID Document processors will not error on for all key types, either JWK or PEM.

Got it... although that might also cause pro/anti camps to pop up due to what current libraries support. We're trying to decrease implementation burden, which is another way to increase interop.

I don't think that point has been clearly made yet. There are multiple ways to get interop:

1. Standardize what sub-ecosystems within a larger ecosystem are already doing (standardize on RSA, ed25519, secp256k1, secp256r1 separately).
2. Standardize at the ecosystem level (standardize on JWK).

Both can be valid approaches. The first reflects the current reality. The second attempts to create a single reality.

msporny:
Updated PR to reflect discussion from latest telecon, straw poll results summarized here: https://github.com/w3c/did-core/issues/67#issuecomment-554680898

index.html
```diff
@@ -622,6 +622,17 @@ <h2>
 <code>type</code>, and a <code>controller</code>, as well as other properties
 which depend on what type of key it is.
       </p>
+
+      <p>
+When expressing public key material in a DID Document, this specification
+strives to provide as few formats as possible to increase the likelihood of
+interoperability. The fewer formats that implementers have to implement, the
```
```
When expressing public key material in a DID Document, this specification
strives to provide as few formats as possible to increase the likelihood of
interoperability.
```
->
```
This specification strives to limit the number of formats for expressing 
public key material in a DID Document to the fewest possible, to increase 
the likelihood of interoperability.
```

index.html
```diff
@@ -622,6 +622,17 @@ <h2>
...
+more likely it will be that they will support all of them. This approach
```
```suggestion
more likely it will be that all implementations will support all formats. This approach
```

index.html
```diff
@@ -622,6 +622,17 @@ <h2>
...
+attempts to strike a delicate balance between ease of implementation and
+supporting formats that have historically had broad deployment. The specific
```
```suggestion
support for formats that have historically had broad deployment. The specific
```

index.html
msporny: Fixed in 8ae2a154f700647dfb3a9780e52d357269beb7b6.

index.html
```diff
@@ -1328,7 +1328,9 @@ <h2>
       </p>
 
       <p>
-The following table summarizes the key formats supported by this specification:
+All public key formats MUST be expressed in either JSON Web Key (JWK)
```
msporny: @selfissued this is the resolution you requested and that passed on the call earlier in the week. Just drawing your attention to it.

index.html
```diff
@@ -1325,6 +1327,45 @@ <h2>
...
+          </tr>
+        </thead>
+
+        <tbody>
+          <tr>
+            <td>
+RSA
+            </td>
+            <td>
+RSA public key values MUST be encoded in Privacy Enhanced Mail
```
```
RSA public key values MUST be encoded in Privacy Enhanced Mail
(PEM) format using the <code>publicKeyPem</code> property if they are not
encoded in JWK format.
```
->
```
RSA public key values MUST be encoded in Privacy Enhanced Mail
(PEM) format using the <code>publicKeyPem</code> property or in
JSON Web Key (JWK) format using the <code>publicKeyJwk</code>
property.
```

TallTed: I do not recall such a straw poll (which [you refer to as a resolution](https://github.com/w3c/did-core/pull/100#discussion_r351089798) above, where you cite the special Monday call, which I was not able to join). I do not think I would have agreed with it, given the result, and I do not recall it being mentioned in the regular Tuesday call. Can you provide a specific pointer to the relevant section of minutes?

msporny:
@selfissued wrote:
> I would prefer that the wording be changed to "RSA public key values MUST be either be encoded as a JWK or be encoded in Privacy Enhanced Mail (PEM) format using the publicKeyPem property." That would put the two representations on a more even editorial footing.

I have applied your exact wording, @selfissued.

index.html
```diff
@@ -1325,6 +1327,44 @@ <h2>
 detail how revocation is performed and tracked.
       </p>
 
+      <p>
+All public key formats MUST be expressed in either JSON Web Key (JWK)
+format using the <code>publicKeyJwk</code> property or one of the native
+formats listed in the table below. All other key formats MUST NOT be used.
```
```suggestion
formats listed in the table below. Public key expression MUST NOT use any other key format.
```

common.js
```diff
@@ -116,6 +116,17 @@ var ccg = {
       status: "Internet-Draft",
       publisher: "IETF"
     },
+    "BASE58": {
+      title: "The Base58 Encoding Scheme",
+      date: "December 2019",
+      href: "https://tools.ietf.org/html/draft-msporny-base58",
+      authors: [
+        "Satoshi Nakamoto",
```
selfissued: Delete the Satoshi Nakamoto author from this Internet Draft.  The IETF requires IPR declarations from all authors and unless you can convince Satoshi to make his IPR declaration, this draft will not be advanced to RFC status with him as an author.

index.html
```diff
@@ -1325,6 +1327,44 @@ <h2>
 detail how revocation is performed and tracked.
       </p>
 
+      <p>
+All public key formats MUST be expressed in either JSON Web Key (JWK)
+format using the <code>publicKeyJwk</code> property or one of the native
```
selfissued: Either replace the word "native" with "other" or "additional" or simply delete it.  (Native isn't correct because the native key formats used by cryptographic algorithms are all binary.)

index.html
```diff
@@ -1325,6 +1327,44 @@ <h2>
...
+formats listed in the table below. All other key formats MUST NOT be used.
+      </p>
+
+      <p class="issue">
+The Working Group is still debating whether the default base encoding format
```
selfissued: Change "default base encoding format" to "base encoding format used".  (There are no defaults - only specified representations.)

peacekeeper:
I propose that we still also discuss hex encoding (base 16) as an option for secp256k1 public keys (https://github.com/w3c/did-core/pull/101) and perhaps also for secp256r1 public keys (https://github.com/w3c/did-core/pull/109). While in the Bitcoin/Ethereum world Base58 encoding is clearly popular for addresses, I believe that for public keys hex encoding may be more popular.

I am fine with @selfissued 's editorial clarification, since I believe it doesn't say that Base58 (Bitcoin) and base64url are the ONLY options at this point.

Also, if I understood correctly, I think the plan was to discuss this in more detail elsewhere after the PRs are merged.

index.html
```diff
@@ -1325,6 +1327,84 @@ <h2>
 detail how revocation is performed and tracked.
       </p>
 
+      <p>
+All public key formats MUST be expressed in either JSON Web Key (JWK)
+format using the <code>publicKeyJwk</code> property or one of the native
```
```suggestion
format using the <code>publicKeyJwk</code> property or one of the 
```

common.js
msporny:
```suggestion
```

index.html
```diff
@@ -1325,6 +1327,84 @@ <h2>
...
+formats listed in the table below. All other key formats MUST NOT be used.
+      </p>
+
+      <p class="issue">
+The Working Group is still debating whether the default base encoding format
```
```suggestion
The Working Group is still debating whether the base encoding format used
```

index.html
msporny: Applied wording that you seemed to be ok with in another thread to this @TallTed.

msporny:
@ChristopherA wrote:
> Secp256k1 needs to be specific to ecdsa/asa Koblitz, as Schnorr is different.

Can you suggest some specification text? I've noted that we can specify whether the public key value is a Koblitz one or a Schnorr one via the `type` property. I can also update the text right now as Schnorr keys are 33 bytes (IIRC) and Koblitz ones are 32 bytes. 

I'm confused by what you would like given your comment, can you please clarify... concrete spec text will go a long way to alleviate the confusion.

ChristopherA:
Interestingly, [bip-schnorr](https://github.com/sipa/bips/blob/bip-schnorr/bip-schnorr.mediawiki) public keys are now 32 bytes through an [x-only trick](https://link.medium.com/p49tGlQ6f2).

My proposal is there are two entries: secp256k1-koblitz and the other is secp256k1-schnoor. Both are 32 bytes. Or if someone complains the Schnorr version is too new, just name it secp256k1-koblitz and leave out the other. But do not call it only secp256k1.

index.html
```diff
@@ -1336,6 +1339,84 @@ <h2>
...
 
+      <p>
+All public key formats MUST be expressed in either JSON Web Key (JWK)
+format using the <code>publicKeyJwk</code> property or one of the 
+formats listed in the table below. Public key expression MUST NOT use any other key format.
+      </p>
+
+      <p class="issue">
+The Working Group is still debating whether the base encoding format used
+will be Base58 (Bitcoin) [[BASE58]] or base64url [[RFC4648]]. The entries in the
```
peacekeeper:
```suggestion
will be Base58 (Bitcoin) [[BASE58]], base64url [[RFC4648]] or base16 (hex) [[RFC4648]]. The entries in the
```

index.html
```diff
@@ -1336,6 +1339,84 @@ <h2>
...
+table below currently assume Base58 (Bitcoin), but may change to base64url once
```
peacekeeper:
```suggestion
table below currently assume PEM and Base58 (Bitcoin), but may change to base64url and/or base16 (hex) once
```

awoie:
IMO, the mandatory key formats should be either JWK or PEM as they are widely accepted formats. Many tools have built-in support for reading and converting between them. Most libraries don't support b58 keys. If I use an existing library and want to provide a public key from a DID Document that contains b58 keys only, then I would have to implement the conversion from b58 into the format the library supports and that for every crypto algorithm. In some programming languages it is hard to find support for Ed25519 and if it gets encoded as b58, then it makes finding the right tool even harder. IMO, that is a big issue for adoption.

msporny:
@awoie wrote:
> IMO, the mandatory key formats should be either JWK or PEM as they are widely accepted formats.

We explored that position in the calls and it was not leading to consensus.

> Many tools have built-in support for reading and converting between them. Most libraries don't support b58 keys. 

Citation? There were a number of assertions that were being thrown around such as this during the discussion (from all sides), which is why we started this:

https://github.com/w3c/did-core/issues/67#issuecomment-552947629

If someone is going to claim more support for X over Y and doesn't have the data to back it up at this point, that's going to be problematic. base58 support doesn't exist in many PEM libraries, but neither does base16... but there is plenty of base58 support in Bitcoin (for example). So, it's not as cut and dry as people are making it seem.

At this point, one interpretation of what you're writing, @awoie is that you're objecting to this PR (which would be really unfortunate at this point because we have consensus among the people that were regularly participating in the discussion).

Are you objecting to merging this PR? If so, what changes (that we could get to consensus on) are you suggesting? Would you be okay with noting your concerns in an issue inline in the spec and merging that? If so, can you write the exact text you'd like to see in the issue?

awoie:
@msporny according to the [excel sheet (import/export overview)](https://docs.google.com/spreadsheets/d/12dwUaAruKKpq3a3IfEMEMhpRhI7oM1tQnkmDbW2VGoU/edit#gid=0), there is only one lib that natively supports base58 and that one has support for PEM/DER as well. Ok, one could argue all these JOSE libs are no fully-fledged crypto libs because they use bouncycastle, nodejs or other libs under the hood but they are still widely used and people who are using them, don't want to use other crypto libs just to convert a key from one format to another. I don't object to this PR as long as there is a way for implementers to specify their preferred public key format upon resolution/dereferencing and we recommend that resolvers should support certain key formats.

awoie:
@msporny I will create a separate issue for the issue stated above.

msporny:
> @msporny according to the excel sheet (import/export overview)

That excel spreadsheet isn't done at all... I started it, added some data, and then it wasn't finished... So we all need to roll up our sleeves and do the work to learn about what encoding formats are actually supported and at what layer.

> @msporny I will create a separate issue for the issue stated above.

Great, thank you... My read is that you are not objecting to this PR based on the following reasons:

1. You are raising an issue that the group must process before we go to CR.
2. There is a discussion currently about whether options can be sent to a matrix parameter or DID resolver that does the conversation automatically.

For those reasons, the merge after the call tomorrow for this PR isn't blocked.

awoie:
@msporny Thanks for the heads up. Yes, no objections for those reasons.

common.js
msporny: Done.

index.html
msporny: Done.

msporny:
@ChristopherA wrote:
> My proposal is there are two entries: secp256k1-koblitz and the other is secp256k1-schnoor. Both are 32 bytes. Or if someone complains the Schnorr version is too new, just name it secp256k1-koblitz and leave out the other. But do not call it only secp256k1.

I have specified secp256k1-koblitz in the table as I believe that's the value that we were discussing during the calls. I hesitated to add Schnorr, even though it shouldn't be a problem, because doing so in this PR might have raised a last minute objection to the PR. So instead, I highlighted in an inline issue in the specification that the group is still discussing if Schnorr will be discussed in the specification, and if so, how it will be expressed and encoded. This makes Koblitz values clear, and keeps the door open for Schnorr, which should happen in a separate PR after we have review and consensus. There are also questions around secp[256|384|521]r1 and what we're doing there that will most likely also happen in parallel.

These changes were made in: https://github.com/w3c/did-core/pull/100/commits/96f38b3a70d75f2ddd7af4b81455036b5d208ba1

index.html
```diff
@@ -1336,6 +1339,91 @@ <h2>
...
 
+      <p>
+All public key formats MUST be expressed in either JSON Web Key (JWK)
+format using the <code>publicKeyJwk</code> property or one of the
+formats listed in the table below. Public key expression MUST NOT use any other key format.
+      </p>
+
+      <p class="issue">
+The Working Group is still debating whether the base encoding format used
+will be Base58 (Bitcoin) [[BASE58]], base64url [[RFC4648]] or base16 (hex) [[RFC4648]]. The entries in the
```
selfissued: Change the base64url reference from RFC4648 to RFC7515, since it specifies no padding

TallTed: @selfissued - Maybe you could bring this up in a new PR/issue, so this one (which was OK with you three days ago) can be merged?

selfissued: I filed this review during the call because I'd raised this in a comment last week and @msporny said on the call that he'd addressed all the comments but apparently mine was missed.  It's a 4-character change - then it can be merged.

TallTed: I see [your previous comment](https://github.com/w3c/did-core/pull/100#pullrequestreview-327053977) now. Took some digging, as github doesn't reveal it when searching for `rfc7515` nor `7515`. I'm putting this in now *as an explicit change request*.

```suggestion
will be Base58 (Bitcoin) [[BASE58]], base64url [[RFC7515]] or base16 (hex) [[RFC4648]]. The entries in the
```

msporny: Applied in https://github.com/w3c/did-core/pull/100/commits/5287f429d14aefa26d691afd7d868057ffe0e68e.

msporny:
Normative and substantive changes, multiple reviews, lots of debate, lots of feedback, much of it applied, intent to merge sent out last week and announced on call today. We have consensus to merge. Merging.

PR #101, opened 2019-11-10
Add key expression formats for secp256k1.

msporny:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/101.html" title="Last updated on Dec 6, 2019, 3:50 PM UTC (3eadfb7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/101/5b3bfcb...3eadfb7.html" title="Last updated on Dec 6, 2019, 3:50 PM UTC (3eadfb7)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1392,6 +1392,19 @@ the raw 32-byte public key value encoded in Base58 Bitcoin format using the
 <code>publicKeyBase58</code> property.
             </td>
           </tr>
+          <tr>
+            <td>
+secp256k1
+            </td>
+            <td>
+Secp256k1 key values MAY be expressed as the raw 32-byte value encoded in
+base58-btc format using the <code>publicKeyBase58</code> property or
+MAY be encoded in JSON Web Key (JWK) format using the <code>publicKeyJwk</code>
+property. Both the raw 32-byte base58-btc format and JWK format MUST be
+supported by DID Document Processors. The use of other key formats for RSA MUST
+cause the DID Document Processor to raise an error.
+            </td>
+          </tr>
         </tbody>
       </table>
 
--- a/index.html
+++ b/index.html
@@ -1369,7 +1369,7 @@ ed25519
             <td>
 Ed25519 public key values MUST either be encoded as a JWK or be encoded as
 the raw 32-byte public key value in Base58 Bitcoin format using the
-<code>publicKeyBase58</code> property if they are not encoded in JWK format.
+<code>publicKeyBase58</code> property.
             </td>
           </tr>
           <tr>
@@ -1379,7 +1379,7 @@ secp256k1
             <td>
 Secp256k1 public key values MUST either be encoded as a JWK or be encoded as the
 raw 32-byte public key value in Base58 Bitcoin format using the
-<code>publicKeyBase58</code> property if they are not encoded in JWK format.
+<code>publicKeyBase58</code> property.
             </td>
           </tr>
           <tr>
--- a/index.html
+++ b/index.html
@@ -1367,9 +1367,9 @@ Privacy Enhanced Mail (PEM) format using the <code>publicKeyPem</code> property.
 secp256k1
             </td>
             <td>
-Secp256k1 public key values MUST be encoded as the raw 32-byte public key value
-encoded in Base58 Bitcoin format using the <code>publicKeyBase58</code>
-property if they are not encoded in JWK format.
+Secp256k1 public key values MUST either be encoded as a JWK or be encoded
+as the raw 32-byte public key value encoded in Base58 Bitcoin format using
+the <code>publicKeyBase58</code> property.
             </td>
           </tr>
         </tbody>
--- a/index.html
+++ b/index.html
@@ -1364,6 +1364,18 @@ Privacy Enhanced Mail (PEM) format using the <code>publicKeyPem</code> property.
           </tr>
           <tr>
             <td>
+secp256r1
+            </td>
+            <td>
+Secp256r1 public key values MUST be encoded in either JSON Web Key (JWK)
+format using the <code>publicKeyJwk</code> property or as the raw 32-byte
+public key value encoded in Base58 Bitcoin format using the
+<code>publicKeyBase58</code> property. Other secp256r1 public key formats
+MUST NOT be used.
+            </td>
+          </tr>
+          <tr>
+            <td>
 ed25519
             </td>
             <td>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1338,6 +1338,19 @@ <h2>
...
             </td>
           </tr>
+          <tr>
+            <td>
+secp256k1
+            </td>
+            <td>
+Secp256k1 key values MAY be expressed as the raw 32-byte value encoded in
+base58-btc format using the <code>publicKeyBase58</code> property or
+MAY be encoded in JSON Web Key (JWK) format using the <code>publicKeyJwk</code>
```
dlongley: Is it common for anyone to use JWK for this key type? If not, we could just specify the one format here.

ChristopherA:
One of my naming concerns here is that a kobliz pubkey (used by bitcoin today) is not the same key as a Schnorr pubkey (which bitcoin is moving to). Both use secp256k1.

— Christopher Allen

index.html
selfissued: Yes, use of JWKs for Ed25519 is common.  That's why members of the community developed RFC 8037 - to define how to use the 25519 and 448 curves with JSON.

index.html
```diff
@@ -1338,6 +1338,19 @@ <h2>
 key formats for RSA MUST cause the DID Document Processor to raise an error.
             </td>
           </tr>
+          <tr>
+            <td>
+secp256k1
+            </td>
+            <td>
+Secp256k1 key values MAY be expressed as the raw 32-byte value encoded in
+base58-btc format using the <code>publicKeyBase58</code> property or
```
yancyribbens: s/base58-btc/base58

index.html
```diff
@@ -1338,6 +1338,19 @@ <h2>
...
+MAY be encoded in JSON Web Key (JWK) format using the <code>publicKeyJwk</code>
+property. Both the raw 32-byte base58-btc format and JWK format MUST be
```
yancyribbens: s/base58-btc/base58/

index.html
msporny: There are two base58 encodings... Bitcoin's and Flickr's... We need to be specific about which one we're using (most use Bitcoin's base58 encoding scheme).

index.html
msporny: There are two base58 encodings... Bitcoin's and Flickr's... We need to be specific about which one we're using (most use Bitcoin's base58 encoding scheme).

msporny:
@ChristopherA wrote:
> One of my naming concerns here is that a kobliz pubkey (used by bitcoin today) is not the same key as a Schnorr pubkey (which bitcoin is moving to). Both use secp256k1.

We should be fine here... the Linked Data `type` property can be used to differentiate between the two. For example:

```
{
  "id": "did:example:1234456#key1",
  "type": "secp256k1KoblitzVerificationKey2019",
  "publicKeyBase58": "3luvf89r34905j3ouf934h5i8yf9s3"
}
```

```
{
  "id": "did:example:1234456#key2",
  "type": "secp256k1SchnorrVerificationKey2021",
  "publicKeyBase58": "3luvf89r34905j3ouf934h5i8yf9s3"
}
```

index.html
msporny: > Yes, there are people using the secp256k1 JWK representation

Could you get these people to fill out the survey: 

https://forms.gle/Hovf3irBJ5KwgXLQ6

If any of these people are writing base cryptography libraries (not wrapping libraries), please also get them to fill out:

https://docs.google.com/spreadsheets/d/12dwUaAruKKpq3a3IfEMEMhpRhI7oM1tQnkmDbW2VGoU/edit

index.html
yancyribbens: Thanks for clarifying.  There doesn't seem to be any `base58-btc` mentioned elsewhere and since Bitcoin is credited with the first base58 implementation I would have assumed that was what was meant.

msporny:
I've updated this PR based on comments in this thread and where the group seems to be currently as outlined here: https://github.com/w3c/did-core/issues/67#issuecomment-554680898

OR13:
I'm not sure if the `crv` is worth providing guidance, but I want to call out a common issue with JWK representations of secp256k1: https://github.com/decentralized-identity/papers/issues/23

Note that the crv "P-256K" is legacy / deprecated, https://github.com/panva/jose#secp256k1

https://tools.ietf.org/html/draft-ietf-cose-webauthn-algorithms-01

> Changed the JOSE curve identifier from "P-256K" to "secp256k1".

If we can help the community avoid this ambiguity I believe we should.

index.html
msporny: There is now a reference to BASE58 in the spec, which points here: https://tools.ietf.org/html/draft-msporny-base58

index.html
msporny: There is now a specific ref here: https://tools.ietf.org/html/draft-msporny-base58

msporny:
@selfissued wrote:
> The PR as written does not reflect the results of the straw poll on the last working group key representation call. 

There is a catch all statement in the spec now that says: 

> All public key formats MUST be expressed in either JSON Web Key (JWK) format using the <code>publicKeyJwk</code> property or one of the native formats listed in the table below.

I notified you of that change here: https://github.com/w3c/did-core/pull/100#discussion_r351089798

I think you realized this shortly after you wrote the comment above... I'm leaving this comment in here for others that may not want to jump around the comments to figure out the order in which all of this happened.

msporny:
@selfissued wrote:
> I would prefer that the wording be changed to "Secp256k1 public key values MUST either be encoded as a JWK or be encoded as the raw 32-byte public key value encoded in Base58 Bitcoin format using the publicKeyBase58 property." That would put the two representations on a more even editorial footing.

I have applied your exact wording to the PR, @selfissued.

ChristopherA:
-1 I still have the problem that secp256k1 public keys can be koblitz ecdsa or schnoor. You must say which.

selfissued:
@ChristopherA is of course right when he writes "_I still have the problem that secp256k1 public keys can be koblitz ecdsa or schnoor. You must say which._"

msporny:
@selfissued wrote:
> @ChristopherA is of course right when he writes "I still have the problem that secp256k1 public keys can be koblitz ecdsa or schnoor. You must say which."

It was already demonstrated how we can do this here:

https://github.com/w3c/did-core/pull/101#pullrequestreview-316278222

index.html
```diff
@@ -1338,6 +1338,19 @@ <h2>
...
+supported by DID Document Processors. The use of other key formats for RSA MUST
```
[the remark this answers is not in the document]
msporny: You were looking at an outdated PR... that text no longer exists in the most recent PR.

msporny:
Ok, this PR is good to go, merging down into msporny-issue-67, please make all future comments there.

PR #102, opened 2019-11-10
Add key expression formats for ed25519.

msporny:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/102.html" title="Last updated on Dec 6, 2019, 3:47 PM UTC (493b8ed)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/102/943db24...493b8ed.html" title="Last updated on Dec 6, 2019, 3:47 PM UTC (493b8ed)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1392,6 +1392,19 @@ the raw 32-byte public key value encoded in Base58 Bitcoin format using the
 <code>publicKeyBase58</code> property.
             </td>
           </tr>
+          <tr>
+            <td>
+secp256k1
+            </td>
+            <td>
+Secp256k1 key values MAY be expressed as the raw 32-byte value encoded in
+base58-btc format using the <code>publicKeyBase58</code> property or
+MAY be encoded in JSON Web Key (JWK) format using the <code>publicKeyJwk</code>
+property. Both the raw 32-byte base58-btc format and JWK format MUST be
+supported by DID Document Processors. The use of other key formats for RSA MUST
+cause the DID Document Processor to raise an error.
+            </td>
+          </tr>
         </tbody>
       </table>
 
--- a/index.html
+++ b/index.html
@@ -1369,7 +1369,7 @@ ed25519
             <td>
 Ed25519 public key values MUST either be encoded as a JWK or be encoded as
 the raw 32-byte public key value in Base58 Bitcoin format using the
-<code>publicKeyBase58</code> property if they are not encoded in JWK format.
+<code>publicKeyBase58</code> property.
             </td>
           </tr>
           <tr>
@@ -1379,7 +1379,7 @@ secp256k1
             <td>
 Secp256k1 public key values MUST either be encoded as a JWK or be encoded as the
 raw 32-byte public key value in Base58 Bitcoin format using the
-<code>publicKeyBase58</code> property if they are not encoded in JWK format.
+<code>publicKeyBase58</code> property.
             </td>
           </tr>
           <tr>
--- a/index.html
+++ b/index.html
@@ -1364,6 +1364,18 @@ Privacy Enhanced Mail (PEM) format using the <code>publicKeyPem</code> property.
           </tr>
           <tr>
             <td>
+secp256r1
+            </td>
+            <td>
+Secp256r1 public key values MUST be encoded in either JSON Web Key (JWK)
+format using the <code>publicKeyJwk</code> property or as the raw 32-byte
+public key value encoded in Base58 Bitcoin format using the
+<code>publicKeyBase58</code> property. Other secp256r1 public key formats
+MUST NOT be used.
+            </td>
+          </tr>
+          <tr>
+            <td>
 ed25519
             </td>
             <td>
--- a/index.html
+++ b/index.html
@@ -1405,6 +1405,19 @@ supported by DID Document Processors. The use of other key formats for RSA MUST
 cause the DID Document Processor to raise an error.
             </td>
           </tr>
+          <tr>
+            <td>
+secp256k1
+            </td>
+            <td>
+Secp256k1 key values MAY be expressed as the raw 32-byte value encoded in
+base58-btc format using the <code>publicKeyBase58</code> property or
+MAY be encoded in JSON Web Key (JWK) format using the <code>publicKeyJwk</code>
+property. Both the raw 32-byte base58-btc format and JWK format MUST be
+supported by DID Document Processors. The use of other key formats for RSA MUST
+cause the DID Document Processor to raise an error.
+            </td>
+          </tr>
         </tbody>
       </table>
 
--- a/index.html
+++ b/index.html
@@ -1395,6 +1395,19 @@ supported by DID Document Processors. The use of other key formats for RSA MUST
 cause the DID Document Processor to raise an error.
             </td>
           </tr>
+          <tr>
+            <td>
+secp256k1
+            </td>
+            <td>
+Secp256k1 key values MAY be expressed as the raw 32-byte value encoded in
+base58-btc format using the <code>publicKeyBase58</code> property or
+MAY be encoded in JSON Web Key (JWK) format using the <code>publicKeyJwk</code>
+property. Both the raw 32-byte base58-btc format and JWK format MUST be
+supported by DID Document Processors. The use of other key formats for RSA MUST
+cause the DID Document Processor to raise an error.
+            </td>
+          </tr>
         </tbody>
       </table>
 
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1348,6 +1348,19 @@ <h2>
...
+          <tr>
+            <td>
+ed25519
+            </td>
+            <td>
+ed25519 key values MAY be expressed as the raw 32-byte value encoded in
+base58-btc format using the <code>publicKeyBase58</code> property or
+MAY be encoded in JSON Web Key (JWK) format using the <code>publicKeyJwk</code>
+property. Both the raw 32-byte base58-btc format and JWK format MUST be
+supported by DID Document Processors. The use of other key formats for RSA MUST
```
troyronda: This line has RSA instead of ed25519?

index.html
```diff
@@ -1348,6 +1348,19 @@ <h2>
...
+            </td>
+          </tr>
+          <tr>
+            <td>
+ed25519
+            </td>
+            <td>
+ed25519 key values MAY be expressed as the raw 32-byte value encoded in
+base58-btc format using the <code>publicKeyBase58</code> property or
+MAY be encoded in JSON Web Key (JWK) format using the <code>publicKeyJwk</code>
```
dlongley: One question here is ... does anyone really use JWK for ed25519? If not, we could eliminate it and just have the common base58 representation.

selfissued: Yes, use of JWKs for Ed25519 is common. That's why members of the community developed RFC 8037 - to define how to use the 25519 and 448 curves with JSON.

dlongley:
Note: We also need one of these PRs for Curve25519 (X25519KeyAgreementKeys) -- I'd expect it to look pretty much the same as this one.

index.html
kdenhartog: I'm not sure if anyone is currently using JWKs as a representation right now for ed25519 keys, (legacy implementations appear to use base58) but I read this as implementers of future DID registries are encouraged to use JWKs rather than base58.

index.html
```diff
@@ -1348,6 +1348,19 @@ <h2>
 MAY be encoded in JSON Web Key (JWK) format using the <code>publicKeyJwk</code>
 property. Both the raw 32-byte base58-btc format and JWK format MUST be
 supported by DID Document Processors. The use of other key formats for RSA MUST
```
tplooker: RSA mentioned on the end of this line too?

index.html
msporny: This was a mistake, will fix.

index.html
msporny: This was a mistake, will fix.

index.html
```diff
@@ -1348,6 +1348,19 @@ <h2>
...
+cause the DID Document Processor to raise an error.
+            </td>
+          </tr>
+          <tr>
+            <td>
+ed25519
+            </td>
+            <td>
+ed25519 key values MAY be expressed as the raw 32-byte value encoded in
+base58-btc format using the <code>publicKeyBase58</code> property or
```
yancyribbens: While base58 was introduced by btc, I don't think there's any difference between `base58` and `base58-btc`.  Would be more clear to simply say `base58` here IMO.

index.html
```diff
@@ -1348,6 +1348,19 @@ <h2>
...
+MAY be encoded in JSON Web Key (JWK) format using the <code>publicKeyJwk</code>
+property. Both the raw 32-byte base58-btc format and JWK format MUST be
```
yancyribbens: same comment here. s/base58-btc/base58/

index.html
msporny: @selfissued wrote:
> Yes, use of JWKs for Ed25519 is common.

Do you have data or a citation for this assertion? I expect that's true in certain communities, but which ones, and which implementations?

index.html
msporny: There are two base58 encodings... Bitcoin's and Flickr's... We need to be specific about which one we're using (most use Bitcoin's base58 encoding scheme).

index.html
msporny: There are two base58 encodings... Bitcoin's and Flickr's... We need to be specific about which one we're using (most use Bitcoin's base58 encoding scheme).

index.html
selfissued: I'm aware of all these implementations:

- python: https://github.com/latchset/jwcrypto/
- node.js: https://github.com/panva/jose
- elixir: https://github.com/joken-elixir/joken
- go: https://github.com/pascaldekloe/jwt
- php: https://github.com/web-token/jwt-framework

I'm sure there are others...

index.html
msporny: Fixed in 41b0b4781bab8fdfae2cc42990bacc670342e04f.

index.html
msporny: Fixed in 41b0b4781bab8fdfae2cc42990bacc670342e04f.

index.html
msporny: Tried to clarify in Fixed in 41b0b4781bab8fdfae2cc42990bacc670342e04f.

index.html
msporny: Tried to clarify in Fixed in 41b0b4781bab8fdfae2cc42990bacc670342e04f.

msporny:
> Please revise the wording along the lines of that suggested for RSA keys in #100 (review) to indicate that allowing for PublicKeyBase58 is an exception, and explicitly limiting the algorithms and curves that it may be used with.

Updated language to reflect the most recent straw poll results in https://github.com/w3c/did-core/issues/67#issuecomment-554680898.

index.html
```diff
@@ -1346,6 +1346,18 @@ <h2>
...
+            </td>
+          </tr>
+          <tr>
+            <td>
+ed25519
+            </td>
+            <td>
+Ed25519 public key values MUST be encoded in either JSON Web Key (JWK)
+format using the <code>publicKeyJwk</code> property or as the raw 32-byte
+public key value encoded in Base58 Bitcoin format using the
```
TallTed: Each occurrence of `Base58 Bitcoin format` should be linked to (a reference to) the relevant spec/RFC/whatever.

OR13: https://tools.ietf.org/id/draft-multiformats-multibase-00.html#rfc.appendix.B.3 ?

TallTed: This appears to be a better choice, albeit not a standard in the usual sense, as it actually *says* something -- <https://en.bitcoin.it/wiki/Base58Check_encoding>

index.html
```diff
@@ -1365,6 +1365,16 @@ <h2>
           </tr>
           <tr>
             <td>
+ed25519
+            </td>
+            <td>
+Ed25519 public key values MUST be encoded as the raw 32-byte public key value
```
```
Ed25519 public key values MUST be encoded as the raw 32-byte public key value
encoded in Base58 Bitcoin format using the <code>publicKeyBase58</code> property
if they are not encoded in JWK format.
```
->
```
Ed25519 public key values MUST be encoded in JSON Web Key (JWK) format using
the <code>publicKeyJwk</code> property or with the raw 32-byte public key
value encoded in Base58 Bitcoin format using the <code>publicKeyBase58</code>
property.
```

TallTed: I do not recall such a straw poll (which [you refer to nearby as a resolution](https://github.com/w3c/did-core/pull/100#discussion_r351089798), where you cite the special Monday call, which I was not able to join). I do not think I would have agreed with it, given the result, and I do not recall it being mentioned in the regular Tuesday call. Can you provide a specific pointer to the relevant section of minutes?

msporny:
@selfissued wrote:
> I would prefer that the wording be changed to "Secp256k1 public key values MUST either be encoded as a JWK or be encoded as the raw 32-byte public key value encoded in Base58 Bitcoin format using the `publicKeyBase58` property." That would put the two representations on a more even editorial footing.

I have applied a modified version of what you suggested as I think you meant to say ed25519 since this is the Ed25519 PR: https://github.com/w3c/did-core/pull/102/commits/12f2ca3ba624da604479f761f148aca4c2b4d7c5

index.html
msporny: This was done in the msporny-issue-67 PR.

msporny: BTW, Base58Check *is not* Base58 BTC -- Base58Check is a bad name and requires a funky double sha256 + checksum algorithm above and beyond base encoding. We are suggesting base58btc. That said, the issue notes that we haven't made a decision here yet, and still need to do so.

index.html
msporny: The table now does the following thing you requested:
>  it should be fully explicit (and mention both/all encoding options for each public key in the row)

msporny:
Ok, this PR is good to be merged down into msporny-issue-67-secp256k1... please make all future comments there.

PR #109, opened 2019-11-17
Add key expression formats for secp256r1.

msporny:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/109.html" title="Last updated on Dec 6, 2019, 3:45 PM UTC (553639d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/109/12f2ca3...553639d.html" title="Last updated on Dec 6, 2019, 3:45 PM UTC (553639d)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1392,6 +1392,19 @@ the raw 32-byte public key value encoded in Base58 Bitcoin format using the
 <code>publicKeyBase58</code> property.
             </td>
           </tr>
+          <tr>
+            <td>
+secp256k1
+            </td>
+            <td>
+Secp256k1 key values MAY be expressed as the raw 32-byte value encoded in
+base58-btc format using the <code>publicKeyBase58</code> property or
+MAY be encoded in JSON Web Key (JWK) format using the <code>publicKeyJwk</code>
+property. Both the raw 32-byte base58-btc format and JWK format MUST be
+supported by DID Document Processors. The use of other key formats for RSA MUST
+cause the DID Document Processor to raise an error.
+            </td>
+          </tr>
         </tbody>
       </table>
 
--- a/index.html
+++ b/index.html
@@ -1369,7 +1369,7 @@ ed25519
             <td>
 Ed25519 public key values MUST either be encoded as a JWK or be encoded as
 the raw 32-byte public key value in Base58 Bitcoin format using the
-<code>publicKeyBase58</code> property if they are not encoded in JWK format.
+<code>publicKeyBase58</code> property.
             </td>
           </tr>
           <tr>
@@ -1379,7 +1379,7 @@ secp256k1
             <td>
 Secp256k1 public key values MUST either be encoded as a JWK or be encoded as the
 raw 32-byte public key value in Base58 Bitcoin format using the
-<code>publicKeyBase58</code> property if they are not encoded in JWK format.
+<code>publicKeyBase58</code> property.
             </td>
           </tr>
           <tr>
--- a/index.html
+++ b/index.html
@@ -1364,6 +1364,18 @@ Privacy Enhanced Mail (PEM) format using the <code>publicKeyPem</code> property.
           </tr>
           <tr>
             <td>
+secp256r1
+            </td>
+            <td>
+Secp256r1 public key values MUST be encoded in either JSON Web Key (JWK)
+format using the <code>publicKeyJwk</code> property or as the raw 32-byte
+public key value encoded in Base58 Bitcoin format using the
+<code>publicKeyBase58</code> property. Other secp256r1 public key formats
+MUST NOT be used.
+            </td>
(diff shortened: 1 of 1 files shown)


TallTed:
Phrasing definitely needs revision, to make the two alternatives more clear. This segment --
```
... public key values MUST be encoded in either JSON Web Key (JWK)
format using the <code>publicKeyJwk</code> property or as the raw 32-byte
public key value encoded in Base58 Bitcoin format ...
```
-- becoming perhaps --
```
... public key values MUST be encoded either with the entire key value JSON Web Key (JWK) format using the <code>publicKeyJwk</code> property or with the raw 32-byte public key value encoded in Base58 Bitcoin format ...
```

index.html
```diff
@@ -1383,6 +1383,16 @@ <h2>
 property if they are not encoded in JWK format.
             </td>
           </tr>
+          <tr>
+            <td>
+secp256r1
+            </td>
+            <td>
+Secp256r1 public key values MUST be encoded as the raw 32-byte public key value
```
```
Secp256r1 public key values MUST be encoded as the raw 32-byte public key value
encoded in Base58 Bitcoin format using the <code>publicKeyBase58</code> property
if they are not encoded in JWK format.
```
->
```
Secp256r1 public key values MUST be encoded in JSON Web Key (JWK) format using
the <code>publicKeyJwk</code> property or with the raw 32-byte public key
value encoded in Base58 Bitcoin format using the <code>publicKeyBase58</code>
property.
```

TallTed: I do not recall such a straw poll (which [you refer to nearby as a resolution](https://github.com/w3c/did-core/pull/100#discussion_r351089798), where you cite the special Monday call, which I was not able to join). I do not think I would have agreed with it, given the result, and I do not recall it being mentioned in the regular Tuesday call. Can you provide a specific pointer to the relevant section of minutes?

msporny:
@awoie wrote:
> I would be more in favour of having hex or base64 instead of publicKeyBase58 just because there is no real standard that we can refer to for base58.

Fixed in https://tools.ietf.org/html/draft-msporny-base58-01.

I'll update the refs in the spec if the group decides to support base58.

msporny:
@selfissued wrote:
> I would prefer that the wording be changed to "Secp256r1 public key values MUST either be encoded as a JWK or be encoded as the raw 32-byte public key value encoded in Base58 Bitcoin format using the publicKeyBase58 property." That would put the two representations on a more even editorial footing.

Used your language to fix in https://github.com/w3c/did-core/pull/109/commits/7bc30186c2a072cde4787c07331c39f0a88f0f15.

index.html
```diff
@@ -1367,9 +1367,9 @@ <h2>
 ed25519
             </td>
             <td>
-Ed25519 public key values MUST either be encoded as a JWK or be encoded as the
-raw 32-byte public key value in Base58 Bitcoin format using the
-<code>publicKeyBase58</code> property.
+Ed25519 public key values MUST either be encoded as a JWK or be encoded as
+the raw 32-byte public key value in Base58 Bitcoin format using the
+<code>publicKeyBase58</code> property if they are not encoded in JWK format.
```
```suggestion
<code>publicKeyBase58</code> property.
```
The trailing clause is now redundant.

index.html
```diff
@@ -1379,6 +1379,16 @@ <h2>
             <td>
 Secp256k1 public key values MUST either be encoded as a JWK or be encoded as the
 raw 32-byte public key value in Base58 Bitcoin format using the
+<code>publicKeyBase58</code> property if they are not encoded in JWK format.
```
```suggestion
<code>publicKeyBase58</code> property.
```
The trailing clause is now redundant.

kdenhartog:
@msporny can we modify this to support the different all the different NIST curves defined? Right now, I believe P-256 meet the requirements of government contracts which is 112 bits of security defined in [NIST-SP800-131Ar2](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-131Ar2.pdf). However, from what I can tell, NSA currently [defines the parameters ](https://apps.nsa.gov/iaarchive/programs/iad-initiatives/cnsa-suite.cfm) that should be used in commercial applications as needing to use P-384. From what I can tell many European and other western countries (NZ included) require it's use. Can we add support for secp384r1 (P-384) and secp521r1 (P-521) as well with the same parameters? Could we group them together in some way?

msporny:
> @msporny can we modify this to support some more of the different NIST curves defined? 

At some point, yes, let's please not do this now because we finally got consensus to try to merge down what we have to master. We will get those curves in there, @kdenhartog ... just not in this PR. Please raise an issue so we can track your request.

msporny:
Ok, this PR is good to go, merging down into PR msporny-issue-67-ed25519. Please make all future comments there.

PR #110, opened 2019-11-17
Add key expression formats for Curve25519.

msporny:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/110.html" title="Last updated on Dec 6, 2019, 3:41 PM UTC (008b918)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/110/d56304c...008b918.html" title="Last updated on Dec 6, 2019, 3:41 PM UTC (008b918)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1392,6 +1392,19 @@ the raw 32-byte public key value encoded in Base58 Bitcoin format using the
 <code>publicKeyBase58</code> property.
             </td>
           </tr>
+          <tr>
+            <td>
+secp256k1
+            </td>
+            <td>
+Secp256k1 key values MAY be expressed as the raw 32-byte value encoded in
+base58-btc format using the <code>publicKeyBase58</code> property or
+MAY be encoded in JSON Web Key (JWK) format using the <code>publicKeyJwk</code>
+property. Both the raw 32-byte base58-btc format and JWK format MUST be
+supported by DID Document Processors. The use of other key formats for RSA MUST
+cause the DID Document Processor to raise an error.
+            </td>
+          </tr>
         </tbody>
       </table>
 
--- a/index.html
+++ b/index.html
@@ -1405,6 +1405,19 @@ supported by DID Document Processors. The use of other key formats for RSA MUST
 cause the DID Document Processor to raise an error.
             </td>
           </tr>
+          <tr>
+            <td>
+secp256k1
+            </td>
+            <td>
+Secp256k1 key values MAY be expressed as the raw 32-byte value encoded in
+base58-btc format using the <code>publicKeyBase58</code> property or
+MAY be encoded in JSON Web Key (JWK) format using the <code>publicKeyJwk</code>
+property. Both the raw 32-byte base58-btc format and JWK format MUST be
+supported by DID Document Processors. The use of other key formats for RSA MUST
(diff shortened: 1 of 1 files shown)


OR13:
If this IETF document is relevant to the base58 encoding format we are specifically allowing as an alternative to JWK, I would love to see it (or something more authoritative) directly cited: https://tools.ietf.org/id/draft-multiformats-multibase-00.html#rfc.appendix.B.3

awoie:
That document does not specify the base58 encoding itself but rather points out that there different base58 interpretations or de-facto specs such as bitcoin and twitter. We should clearly define what base58 format we are going to refer to, or we should just use base64. Ideally, that should be an official standard published by IETF, Oasis or W3C. Base64 is defined in an official IETF document here: https://tools.ietf.org/html/rfc4648. I don't think that the benefits of base58 really matter in this context.

msporny:
@selfissued wrote: 
> I would prefer that the wording be changed to Curve25519 public key values MUST either be encoded as a JWK or be encoded as the raw 32-byte public key value encoded in Base58 Bitcoin format using the publicKeyBase58 property." That would put the two representations on a more even editorial footing.

Fixed in https://github.com/w3c/did-core/pull/110/commits/e15fbf00bb09f3281dd75fba053ccaa350054597 using your language.

msporny:
@awoie wrote:
> That document does not specify the base58 encoding itself but rather points out that there are different base58 interpretations or de-facto specs such as bitcoin and twitter. We should clearly define what base58 format we are going to refer to, or we should just use base64.

Fixed by referring to https://tools.ietf.org/html/draft-msporny-base58.

index.html
```diff
@@ -1392,6 +1392,16 @@ <h2>
...
             </td>
           </tr>
+          <tr>
+            <td>
+Curve25519
+            </td>
+            <td>
+Curve25519 public key values MUST either be encoded as a JWK or be encoded as
+the raw 32-byte public key value in Base58 Bitcoin format using the
+<code>publicKeyBase58</code> property if they are not encoded in JWK format.
```
```suggestion
<code>publicKeyBase58</code> property.
```
The trailing phrase is redundant.

index.html
```diff
@@ -1389,6 +1389,16 @@ <h2>
...
 the raw 32-byte public key value encoded in Base58 Bitcoin format using the
+<code>publicKeyBase58</code> property.
+            </td>
+          </tr>
+          <tr>
+            <td>
+Curve25519
+            </td>
+            <td>
+Curve25519 public key values MUST either be encoded as a JWK or be encoded as
```
dlongley:
```suggestion
Curve25519 public key (aka X25519 public key) values MUST either be encoded as a JWK or be encoded as
```

TallTed: `Curve25519 public key (aka X25519 public key) values `
->
`Curve25519 public key (also known as X25519 public key) values`

or

`Curve25519 public key (aka X25519 public key) values`
-> 
`Curve25519 (also known as X25519) public key values`

```suggestion
Curve25519 (also known as X25519) public key values MUST either be encoded as a JWK or be encoded as
```

msporny:
Ok, applied everything, starting the merge down to master... next stop msporny-issue-67-secp256r1.

PR #116, opened 2019-11-20
Decentralized Identifiers (DIDs) section review

grantnoble:
Grammar and punctuation changes, reformatting of info into lists, changing ordered lists to unordered lists, and line spacing changes

***
<a href="https://pr-preview.s3.amazonaws.com/grantnoble/did-core/pull/116.html" title="Last updated on Nov 25, 2019, 12:14 PM UTC (a9a07e8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/116/a8d0c28...grantnoble:a9a07e8.html" title="Last updated on Nov 25, 2019, 12:14 PM UTC (a9a07e8)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -658,25 +658,30 @@ Decentralized Identifiers (DIDs)
     </h1>
 
     <p>
-The concept of a globally unique <a>decentralized identifier</a> is not
-new; <a>Universally
-Unique Identifiers</a> (UUIDs) were first developed in the 1980s and
-later became a standard feature of the Open Software Foundation’s
-<a href="https://en.wikipedia.org/wiki/Distributed_Computing_Environment">Distributed
-Computing Environment</a>. <a>UUIDs</a> achieve global uniqueness without a
-centralized registry service by using an algorithm that generates
-128-bit values with sufficient entropy that the chance of collision are
-infinitesimally small. <a>UUIDs</a> are formally specified in [[RFC4122]] as a
-specific type of Unified Resource Name (URN).
+The concept of a globally unique <a>decentralized identifier</a> is not new.
+<a>Universally Unique Identifiers</a> (UUIDs) were first developed in the 1980s
+and later became a standard feature of the Open Software Foundation’s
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -658,54 +658,57 @@ <h1>
...
+A <a>DID</a> is similar to a <a>UUID</a> except that:
     </p>
+      <ul>
+        <li>
+Like a URL, it can be resolved or dereferenced to a standard resource describing
+the subject. That is, a <a>DID document</a>. For more information, see Section
+<a href="#did-documents"></a>.
+        </li>
+        <li>
+Unlike a URL, the <a>DID document</a> usually contains cryptographic material
```
TallTed: Probably fodder for a fresh issue, a URL is parallel to a DID, not to a DID document.

`Unlike a URL, the <a>DID document</a>` should be something like "Unlike most resources returned when dereferencing URLs, a <a>DID document</a>"

grantnoble: Updated.

index.html
```diff
@@ -983,25 +996,25 @@ <h2>
       </h2>
 
       <p>
-For the broadest interoperability, <a>DID</a> normalization should be as
-simple and universal as possible. Therefore:
+For the broadest interoperability, <a>DID</a> normalization should be as simple
```
rhiaro:
```suggestion
For the broadest interoperability, <a>DID</a> normalization is as simple
```

grantnoble: How about "...ensure DID normalization is as simple..."?

rhiaro: My thinking was, in trying to dodge that non-normative should, that this is the spec where we say what the normalization rules are so we can just authoritatively state that it is simple and universal. But I don't feel strongly, could also leave it as is.

grantnoble: Ah. OK, I see where you are coming from now. Reworded as
"For the broadest interoperability, make DID normalization as simple and universal as possible:"

Issue #39, opened 2019-09-23
[Absract] "proof purposes" is a new term but is not defined anywhere in the document


brentzundel:
@mwherman2000 moved from CCG (https://github.com/w3c-ccg/did-spec/issues/213)

rhiaro:
"proof purpose" is no longer in the abstract (or anywhere except some example comments..), I think we can close this issue.

dlongley:
We want to define the term, so we should keep the issue open. The way a DID controller expresses authorized verification methods is via proof purpose relations: i.e., "verification methods [X, Y, Z] can be used to verify proofs made for the purpose of Z (e.g., "authentication", "assertionMethod", "capabilityInvocation", ...)".

PR #123, opened 2019-11-25
Rewrite authentication section

### LINKED ISSUE #2: Rewrite Authentication section - controller is wrong

rhiaro:
Mostly by pulling @dlongley's words from various places around the web :)

This is a first pass at sorting out a lot of uses of 'controller' and should fix the problems with the authentication section. I also updated the definition of 'controller', and added definitions for 'proof purpose' and 'verification method'. It contains new or changed normative language. Needs careful review please! Anticipating changes and corrections.

The Authorization & Delegation section still contains old language still that I'm not sure about.

The Binding of Identity section in Security Considerations is I think also wrong, but I haven't tried to do anything with that yet.

When complete, should fix #2, #3, #22, #39 and help with #4.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/123.html" title="Last updated on Feb 5, 2020, 9:26 PM UTC (24244d1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/123/bee98cb...24244d1.html" title="Last updated on Feb 5, 2020, 9:26 PM UTC (24244d1)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1366,7 +1366,7 @@ The semantics of the <code>controller</code> property are the same when the
 subject of the relationship is the <a>DID document</a> as when the subject
 of the relationship is a key. Since a key can't control itself, and the
 key controller cannot be inferred from the <a>DID document</a>, it is
-necessary to expicitly express the controller of the key is a
+necessary to explicitly express the identity of the controller of the key
 (which may be the <a>DID controller</a> or <a>DID subject</a>). The
 difference is that the value of <code>controller</code> on a key is not
 necessarily a <a>DID controller</a>. <a>DID controller(s)</a> are
--- a/index.html
+++ b/index.html
@@ -43,6 +43,8 @@
     // previousPublishDate:  "1977-03-15",
     // previousMaturity:  "WD",
 
+    pluralize: true,
+
     // extend the bibliography entries
     localBiblio: ccg.localBiblio,
     github: {
--- a/index.html
+++ b/index.html
@@ -1304,6 +1304,7 @@ Public Keys
       </h2>
 
       <p>
+A public key is a <a>verification method</a>.
 Public keys are used for digital signatures, encryption and other cryptographic
 operations, which in turn are the basis for purposes such as authentication
 (see Section <a href="#authentication"></a>) or establishing secure
@@ -1312,6 +1313,19 @@ communication with <a>service endpoints</a> (see Section
 authorization mechanisms of <a>DID</a> CRUD operations (see Section
 <a href="#did-operations"></a>), which can be defined by <a>DID method</a>
 specifications.
+      </p>
+
+      <p class="note" title="Verification methods">
+A public key is just one type of <a>verification method</a>. Ideally,
+public keys are expressed in a <a>DID document</a> according to what
+they are authorized for: <code>authentication</code>, <code>
+capabilityInvocation</code>, <code>capabilityDelegation</code>,
+<code>keyAgreement</code>, <code>assertionMethod</code>, and <em>not</em>
+just listed using the <code>publicKey</code> property. The latter
+implies the <a>DID controller</a> has authorized any key therein for any
+purpose. It's an ambient authority model which is dangerous. Instead, the
+<a>DID controller</a> should be explicit about what the keys can be used
+for using these relations; see <a href="#authentication"></a>.
       </p>
 
       <p>
--- a/index.html
+++ b/index.html
@@ -1347,6 +1347,19 @@ The value of the <code>controller</code> property, which identifies the
 controller of the corresponding private key, MUST be a valid <a>DID</a>.
       </p>
 
+      <p class="note" title="This is not the DID Controller">
+The <code>controller</code> property denotes a <a>DID controller</a>
+only when it appears at the <em>top level</em> of the <a>DID document</a>.
+The <code> controller</code> property when it appears as part of a <a>public
+key description</a> denotes the controller of the <em>corresponding
+private key</em> only, which may or may not be described in another
+<a>DID document</a>. A default <code>controller</code> field cannot
+be inferred by using a key's position in a tree, and so is explicitly
+included for each key. Linked Data Proof libraries typically expect the
+<code>controller</code> field to always exist and may throw an exception
+if it is missing.
+      </p>
+
       <p>
 All public key properties MUST be from the Linked Data Cryptographic Suite
 Registry. For a registry of key types and formats, see Appendix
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -1410,14 +1428,23 @@ <h2>
...
-Authentication is separate from Authorization because the
-<a>DID controller(s)</a> may wish to enable others to update their DID Document
-(for example, to assist with key recovery as discussed in Section
-<a href="#key-revocation-and-recovery"></a>) without enabling them to prove
-control (and thus be able to impersonate the <a>DID controller(s)</a>).
+Authentication is a mechanism by which a <a>DID method</a> can establish
+the authority of the party attempting to authenticate. The
+<em>verifier</em> of an authentication attempt can check if the
+authenticating party is presenting a valid proof of authentication, that is,
+that they are who they say they are.
```
dlongley: Hmm, this doesn't seem right. Instead:

```
Authentication is a mechanism by which an entity can prove it is the DID subject. The <em>verifier</em> of an authentication attempt can check if the authenticating party is presenting a valid proof of authentication, that is, that they are who they say they are. Note that a successful authentication on its own may or may not confer authority; that is up to the verifying application. 
```

index.html
```diff
@@ -1410,14 +1428,23 @@ <h2>
...
+      </p>
+
+      <p class="note" title="Uses of authentication">
+If authority is established, it is up to the <a>DID method</a> to decide
```
dlongley:
```suggestion
If authentication is established, it is up to the <a>DID method</a> or other application to decide
```

index.html
```diff
@@ -1410,14 +1428,23 @@ <h2>
...
+what to do with that information. A particular <a>DID method</a> could
+decide that authenticating as a <a>DID controller</a> is sufficient to,
+for example, update or delete the <a>DID document</a>. Another <a>DID
+method</a> could require different keys, or a different verification method
+entirely, to be presented in order to update or delete the <a>DID document</a>
+than that used to authenticate. In other words, what is done <em>after</em>
+the authentication check is out of scope for the DID data model, but <a>DID
+methods</a> are expected to define this themselves.
```
dlongley:
```suggestion
methods</a> and applications are expected to define this themselves.
```

index.html
```diff
@@ -2078,6 +2150,22 @@ <h2>
...
+      <p>
+Determining the authority of a party to carry out the operations is method-specific.
+For example, a <a>DID method</a> MAY:
+      </p>
+      <ul>
+        <li>
+use the <a>verification methods</a> listed under <a>authentication</a> to decide whether an update/deactivate operation is allowed; or
+        </li>
+        <li>
+use other constructs in the <a>DID Document</a> to decide this; or
```
dlongley: Hmm, we may want to mention: For example, a verification method specified under `capabilityInvocation` could be used to verify the invocation of a capability to update the DID Document.

index.html
```diff
@@ -1262,6 +1278,19 @@ <h2>
...
+The <code>controller</code> property denotes a <a>DID controller</a>
+only when it appears at the <em>top level</em> of the <a>DID document</a>.
+The <code> controller</code> property when it appears as part of a <a>public
+key description</a> denotes the controller of the <em>corresponding
+private key</em> only, which may or may not be described in another
+<a>DID document</a>. A default <code>controller</code> field cannot
+be inferred by using a key's position in a tree, and so is explicitly
+included for each key. Linked Data Proof libraries typically expect the
+<code>controller</code> field to always exist and may throw an exception
+if it is missing.
```
dlongley: The default controller for a key would be itself (the "subject" of the key) just like it is with a DID Document and the DID subject. Since a key can't control itself and the controller of the key should (typically) be the DID subject, the `controller` field must be specified.

I don't think we need to say that Linked Data Proof libraries expect it to exist ... in fact, if they do, that should be fixed so that they assume the controller is the key itself if it is not present. Regardless, keys in a DID document should always have a `controller` because they are controlled by the DID subject (or, transitively, the DID controller if it is not the same as the DID subject).

rhiaro: Yeah, some of that was old wording that I thought was a bit odd. I agree with taking the ldproof part out. `controller` is already a MUST for keys, so I don't think we need to say anything about a default controller, unless that MUST ever comes down to a SHOULD, because it's always explicit otherwise.

dlongley: I think it may also be useful to show that `controller` is being reused with the same semantics as when it appears in the DID Document. The `controller` of a thing/object is specified by the `controller` property -- and since a key can't control itself (it has no keys of its own for example), we want to express that keys are controlled by a DID subject (again, or transitively that DID subject's DID controller).

rhiaro: I misread what you said initially but yeah I agree

rhiaro: That reminds me, can a key have more than one controller? I think we need to explicitly state that it can be an array too right?

rhiaro: Is it true that there could be a case where a key controller is not the DID controller for that DID, but some other entity?

index.html
```diff
@@ -1262,6 +1278,19 @@ <h2>
 controller of the corresponding private key, MUST be a valid DID.
       </p>
 
+      <p class="note" title="Key controller(s) and DID controller(s)">
+The semantics of the <code>controller</code> property are the same when the
+subject of the relationship is the <a>DID document</a> as when the subject
+of the relationship is a key. Since a key can't control itself, and the
+key controller cannot be inferred from the <a>DID document</a>, it is
+necessary to expicitly express the controller of the key is a
```
```suggestion
necessary to explicitly express the identity of the controller of the key
```

index.html
rhiaro: Updated

index.html
rhiaro: Updated

index.html
msporny: > That reminds me, can a key have more than one controller? 

No, it can't.

> I think we need to explicitly state that it can be an array too right?

No, I don't think controller can ever be an array, IIRC.

index.html
```diff
@@ -1229,6 +1232,19 @@ <h2>
...
 defined by <a>DID method</a> specifications.
       </p>
+
+      <p class="note" title="Verification methods">
+A public key is just one type of <a>verification method</a>. Ideally,
+public keys are expressed in a <a>DID document</a> according to what
+they are authorized for: <code>authentication</code>, <code>
+capabilityInvocation</code>, <code>capabilityDelegation</code>,
+<code>keyAgreement</code>, <code>assertionMethod</code>, and <em>not</em>
+just listed using the <code>publicKey</code> property. The latter
```
tplooker: To clarify is the implication of the following sentence starting with "The latter....", that by default any keys listed under the public key section of the did document and not mentioned under any verification method, implying those keys should be interpreted as being authorised for all proof purposes? 

My understanding was that keys listed under the public key section of the did document were just expressions of association of keys to the did subject. Of course applications consuming the did document could choose to infer whatever authority they want from keys listed in the public key section however I don't think the specification should imply this ambient authority type behaviour as a default when no proof purposes are defined in the did document.

msporny: Yes, agreed... boo to ambient authority, will fix.

msporny: Fixed in: https://github.com/w3c/did-core/pull/123/commits/1f6cfc764b60af06396e0a88a3ac0ce084f1d154

PR #138, opened 2019-12-06
Add "transform-keys" matrix parameter.

peacekeeper:

Description: Identifies a variation of a DID document in which public keys are transformed
to a certain format. Supported values are "jwk" (JSON Web Key) and "jwks"
(JSON Web Key Set).

Example: `did:example:1234;transform-keys=jwk`

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/138.html" title="Last updated on Dec 6, 2019, 3:16 PM UTC (65bf8e1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/138/a8d0c28...65bf8e1.html" title="Last updated on Dec 6, 2019, 3:16 PM UTC (65bf8e1)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -817,6 +817,17 @@ Identifies a certain version timestamp of a <a>DID document</a> to be resolved
 Note: This parameter may not be supported by all <a>DID methods</a>.
             </td>
           </tr>
+
+          <tr>
+            <td>
+<code>transform-keys</code>
+            </td>
+            <td>
+Identifies a variation of a <a>DID document</a> in which public keys are transformed
+to a certain format. Supported values are "jwk" (JSON Web Key) and "jwks"
+(JSON Web Key Set). Note: This parameter is method-independent.
+            </td>
+          </tr>
         </tbody>
       </table>
 


peacekeeper:
Additional explanation:

During ongoing discussions on supported public key formats in DID documents, and in the context of work on OIDC-based DID Auth, the idea has come up a few times that a DID resolver could automatically convert public keys in DID documents from one format to another. For some additional background and early experimentation, see https://hackmd.io/XmL-Bjh5TdqV4fj6nwdPEQ

As with all matrix parameters, the question should be if this type of parameter needs to be part of a DID URL, or if it is preferable to pass this to a DID resolver as a separate option. This is similar to the question whether you want certain parameters to be part of an HTTP URL, or passed via HTTP headers.

In this particular case, I think I am actually leaning towards the latter (use a resolver option instead of matrix parameter), but I still wanted to raise the PR to enable a discussion about this.

msporny:
> In this particular case, I think I am actually leaning towards the latter (use a resolver option instead of matrix parameter), but I still wanted to raise the PR to enable a discussion about this.

+1, this is my preference as well. This should be an option passed to the DID Resolver, that's the appropriate layer for this feature.

awoie:
@peacekeeper can we RECOMMEND support for that parameter for the reasons outlined in https://github.com/w3c/did-core/pull/100#issuecomment-563187682 and 
https://github.com/w3c/did-core/pull/100#issuecomment-563343310?

dlongley:
> This should be an option passed to the DID Resolver, that's the appropriate layer for this feature.

+1

I don't think matrix parameters should be used to represent the potentially unbounded number of transformations that could be performed on a DID Document.

I think matrix parameters should only be used when you need to use a DID URL to refer to some external resource that is *identified relative to a DID subject* via some content in its DID Document or some "version" of that content based on, e.g., time or sequence.

For example, suppose there's a resource that is currently on server X but that may move to server Y in the future. You can use a DID URL with a matrix parameter to refer to that resource in a stable way -- because whether you need to go through server X or Y to resolve it depends on the content of the DID Document at resolution time.

kdenhartog:
> In this particular case, I think I am actually leaning towards the latter (use a resolver option instead of matrix parameter)

+1

My thinking on this has been that we'd actually prefer to take this route whenever we don't want to pass data, by reference rather than by value. So for example, if I wanted to pass a key in the initial state of my did-doc after I've already rotated the key AND I've opted to reuse the id. Then this is where the use of the version matrix parameter is of value to be included in the URI. Specifically because I'm trying to use the URI to pass the value by reference, rather than passing the raw key (aka passing by value). If I don't need to pass data by reference though then I don't think we need to add a matrix parameter (and modify the URI as such).

peacekeeper:
> My most recent thinking on this has been that we'd actually prefer to take this route whenever we don't need to pass data by reference.

+1. Also see the language recently added to the spec in https://github.com/w3c/did-core/pull/144.

Since we all agree, I'll close this. This was exactly the kind of discussion I wanted to have.

PR #139, opened 2019-12-06
Add "id" DID parameter instead of "service".

peacekeeper:

Description: Identifies a public key or service or other part of the <a>DID document</a> by its `id` property.

Examples: `did:example:1234;id=key-1`, `did:example:1234;id=openid`

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/139.html" title="Last updated on Apr 7, 2020, 8:39 PM UTC (0b74a40)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/139/b8c3918...0b74a40.html" title="Last updated on Apr 7, 2020, 8:39 PM UTC (0b74a40)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -826,10 +826,10 @@ specified in [[HASHLINK]].
           </tr>
           <tr>
             <td>
-<code>service</code>
+<code>id</code>
             </td>
             <td>
-Identifies a service from the <a>DID document</a> by service ID.
+Identifies a public key or service or other part of the <a>DID document</a> by its `id` property.
             </td>
           </tr>
           <tr>


peacekeeper:
Additional explanation:

The `key` matrix parameter (see https://github.com/w3c/did-core/pull/59) for identifying a public key in the DID document has not found much support, but several people in the community have suggested that a generic `id` matrix parameter could be introduced to generically identify any part of a DID document with a matching `id` property. This would replace and generalize the `service` matrix parameter which is currently used to identity a service in the DID document.

One good argument against this parameter is that very similar DID URLs can be constructed by using a fragment, e.g. `did:example:1234#key-1` instead of `did:example:1234;id=key-1`. This DID URL fragment syntax already has broad support in the community and is familiar from HTTP URLs.

One small difference is that `did:example:1234#key-1` would be dependent on the media type of the DID document, and dereferencing the fragment (a "secondary resource") may happen in a different architectural component than resolving the DID (similar to how an HTTP fragment is dereferenced by a browser, not by a web server), whereas `did:example:1234;id=key-1` would be independent of the media type of the DID document and could be dereferenced entirely by a DID resolver rather than a client application.

This was briefly discussed at Rebooting-the-Web-of-Trust 9 in Prague, see https://github.com/WebOfTrustInfo/rwot9-prague/blob/master/draft-documents/did-resolution-v2.md.

Personally I don't have a strong preference either way, but wanted to raise this since it has come up a few times.

dlongley:
I see matrix parameters as away of dereferencing resources that exist outside of the DID Document (or a way to reference a different "version" of the DID Document). I'm concerned that simply specifying an `id` as a matrix parameter doesn't really do that at all, but references something internal, for which `#` makes more sense, has precedence, and works with existing software.

Furthermore, I don't think we can simply replace `service` with `id` because there were other resolution rules baked into `service` (that were driven only by its presence). When using the `service` matrix parameter, the resolution process would retrieve the value of the `serviceEndpoint` property of the service and inline replace the authority of the DID URL with it to create a new URL (that referenced an external resource). Additional changes would be required to how this process would work in order to replace `service` with `id`. And, if `id` is providing some useful general feature, I imagine we would want it to work for services just as much as anything else, whereby we would need yet another matrix parameter to instruct the resolution process to behave how `service` currently does.

In short, unless `id` is literally a replacement for `service` and functions the same way it used to -- i.e., it's just a renaming of `service` to the shorter `id` -- I think there will be problems with this approach. Even if it were just a direct rename ... I suspect the name would confuse people in that case.

msporny:
This PR exemplifies why I'm so concerned about matrix parameters. There seems to be consistent misuse of the functionality within the Working Group that will only become far worse in the DID Methods themselves. That is, unchecked, this feature is demonstrating itself to be ripe for abuse, which will lead to "bad things" happening in the ecosystem as a result.

This is why I'm requesting that we very clearly articulate when matrix parameters should be used and when they shouldn't be used... and we lock that down as tightly as possible. Even if we were to do that, though, I'm afraid that people will still abuse the feature (as this PR, and many of the other ones that have been recently raised, do).

peacekeeper:
@SmithSamuelM asked me to post the following comment on his behalf:

> As per the criteria discussed today in the DID WG call. I think the service use case of needing an indirection via the DiD. Doc to a service endpoint that does not overload the path, query, or fragment is a cogent justification for service matrix parameter. But I also think that if we can unify this type of indirection into an ID matrix parameter then we avoid having to create additional matrix parameters down the road. The ID must then indicate via another field in its block what type of indirection it is but we are already allowing that for service endpoints so it would not be ambiguous.

kdenhartog:
Marked pending close after resolution 1 on the 4/7/2020 call.

peacekeeper:
While I agree with closing https://github.com/w3c/did-core/pull/60 and https://github.com/w3c/did-core/pull/62, I disagree with closing this one; I think there has been insufficient understanding and discussion. I believe @SmithSamuelM originally proposed this.

The resolution on the [2020-04-07 call](https://www.w3.org/2019/did-wg/Meetings/Minutes/2020-04-07-did) was about the syntax of DID parameters, which is independent of the question which DID parameters we want.

kdenhartog:
@peacekeeper thanks for the clarification. Now that I'm seeing the difference, I agree that this is separate and the PR will require rework to use the new agreed upon way. I've removed the label.

OR13:
@peacekeeper can you get this mergeable, advocate for it, or close it :)

peacekeeper:
I think @SmithSamuelM originally proposed this. While this is independent of the matrix/query parameter syntax discussion (see my comment above), I have a feeling that the WG will not support a generic "id" DID parameter. Let's see if Sam has any comments, otherwise we can close in a few days.

SmithSamuelM:
If there is no interest You can close it.

peacekeeper:
Okay I have a feeling there won't be much support for this, so I'll close for now.

PR #144, opened 2019-12-17
Clarify when DID parameters should (not) be used.

peacekeeper:
In order to further discuss DID parameters (aka matrix parameters), it has been requested that we add text to clarify when to use them and when not to use them.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/144.html" title="Last updated on Dec 17, 2019, 2:49 PM UTC (f50e05b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/144/26f4891...f50e05b.html" title="Last updated on Dec 17, 2019, 2:49 PM UTC (f50e05b)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -807,8 +807,9 @@ Generic DID Parameter Names
 
       <p>
 The <a>DID URL</a> syntax supports a simple, generalized format for parameters
-based on the matrix parameter syntax ([[MATRIX-URIS]]). The ABNF above does not
-specify any parameter names (the <code>param-name</code> rule).
+based on the matrix parameter syntax ([[MATRIX-URIS]]). The ABNF above
+specifies the basic syntax (the <code>param-name</code> rule) but does not
+specify any concrete parameter names.
       </p>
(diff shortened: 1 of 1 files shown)


peacekeeper:
> this PR should be updated to clarify that matrix parameters MUST NOT be used.

This PR is simply clarifying when matrix parameters should be used and when they should not be used. It improves the current spec text which includes matrix parameters.

It's completely valid for the WG to propose to remove matrix parameters from the spec, but that should be in a different PR, and any such PR should also explain how the current use cases involving matrix parameter can still be fulfilled without them (or it should be argued why those use cases should not be supported anymore).

peacekeeper:
Merging, since this has been requested several times, I believe it's an improvement to the current text, and there have been no further comments in this thread.

This PR only adds clarification text about DID parameters. Any discussion or proposal to remove them from the spec should happen elsewhere.

PR #142, opened 2019-12-12
Enable pure JSON processing of DID Documents

msporny:
The purpose of this PR is to enable pure JSON processing of DID Documents while not creating large problems for JSON-LD processors. It is an attempt at a compromise for the various positions asserted in issue #128. Specifically, this PR:

* Makes it clear that you do not need to do any JSON-LD processing.
* Clarifies the processing rules for pure JSON processors.
* Ensures that the processing rules for pure JSON processors can be implemented via a two line `if` statement to keep things simple.
* Ensures that the JSON processing rules don't lead to different semantics from JSON-LD processing rules.
* Clarifies that the JSON-LD context is static and can be hard coded / cached forever (after the spec is an official global standard).

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/142.html" title="Last updated on Dec 12, 2019, 4:11 PM UTC (8b3361e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/142/26f4891...8b3361e.html" title="Last updated on Dec 12, 2019, 4:11 PM UTC (8b3361e)">Diff</a>

PR #145, opened 2019-12-17
removed "context" from requirements

jricher:
In order to facilitate the description of DID documents as pure JSON structures, this pull request removes requirements to use `@context` in the spec for all processors. Note that `@context` is still allowed to be used by JSON-LD providers, it's now just classified as an extension. The extensive examples using JSON-LD are left in as they are still valid, but additional plain-JSON extensions examples still need to be added.

Additionally, a formal requirement for how to process fields you don't know (MUST ignore, MAY ignore, etc) still needs to be agreed upon and added, I just wasn't sure quite where to put it so it's in the extensibility section.

***
<a href="https://pr-preview.s3.amazonaws.com/jricher/did-core/pull/145.html" title="Last updated on Dec 17, 2019, 4:37 PM UTC (1deeefb)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/145/26f4891...jricher:1deeefb.html" title="Last updated on Dec 17, 2019, 4:37 PM UTC (1deeefb)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -370,7 +370,6 @@ as well as services that can be used to interact with the entity.
 
     <pre class="example nohighlight" title="Minimal self-managed DID document">
 {
-  "@context": "https://www.w3.org/ns/did/v1",
   "id": "did:example:123456789abcdefghi",
   "authentication": [{
     <span class="comment">// used to authenticate as did:...fghi</span>
@@ -1122,55 +1121,6 @@ including whether these properties are required or optional.
     </p>
 
     <section>
-      <h2>Contexts</h2>
-
-        <p>
-When two software systems need to exchange data, they need to use terminology
-and a protocol that both systems understand. The <code>@context</code>
-property ensures that two systems operating on the same DID document are using
-mutually agreed terminology.
-        </p>
-
-        <p>
-<a>DID documents</a> MUST include the <code>@context</code> property.
-        </p>
-
-        <p class="note" title="The JSON-LD Context">
-More information about the
-<a href="https://www.w3.org/TR/json-ld/#the-context">JSON-LD Context</a>
-in general can be found in the [[!JSON-LD]] specification.
-        </p>
-
-        <dl>
-          <dt><dfn>@context</dfn></dt>
-          <dd>
-The value of the <code>@context</code> property MUST be one or more
-<a>URIs</a>, where the value of the first <a>URI</a> is
-<code>https://www.w3.org/ns/did/v1</code>. If more than one
-<a>URI</a> is provided, the <a>URIs</a> MUST be interpreted as an ordered set.
-It is RECOMMENDED that dereferencing the <a>URIs</a> results in a document
-containing machine-readable information about the context.
-          </dd>
-        </dl>
-
-      <p>
-Example:
-      </p>
-
-      <pre class="example nohighlight">
-{
-  "@context": "https://www.w3.org/ns/did/v1"
-}
-</pre>
-      <p>
-<a>DID method</a> specifications MAY define their own JSON-LD contexts.
-However it is NOT RECOMMENDED to define a new context unless
-necessary to properly implement the method. Method-specific contexts
-MUST NOT override the terms defined in the generic <a>DID</a> context.
-      </p>
-    </section>
-
-    <section>
       <h2>
 DID Subject
       </h2>
@@ -1254,7 +1204,6 @@ Example:
         </p>
 <pre class="example nohighlight" title="Basic DID document">
 {
-"@context": "https://www.w3.org/ns/did/v1",
 "id": "did:example:123456789abcdefghi",
 "authorizationCapability": [{
   // this entity is a delegate and may update any field in this
@@ -1430,7 +1379,6 @@ Example:
 
       <pre class="example nohighlight" title="Various public keys">
 {
-  "@context": ["https://www.w3.org/ns/did/v1", "https://w3id.org/security/v1"],
   "id": "did:example:123456789abcdefghi",
   <span class="comment">...</span>
   "publicKey": [{
@@ -1579,7 +1527,6 @@ Example:
 
       <pre class="example nohighlight" title="Authentication field containing three verification methods">
 {
-  "@context": "https://www.w3.org/ns/did/v1",
   "id": "did:example:123456789abcdefghi",
   <span class="comment">...</span>
   "authentication": [
@@ -1651,7 +1598,6 @@ Example:
 
       <pre class="example nohighlight" title="DID document with a controller property">
 {
-  "@context": "https://www.w3.org/ns/did/v1",
   "id": "did:example:123456789abcdefghi",
   "controller": "did:example:bcehfew7h32f32h7af3",
   "service": [{
@@ -1897,6 +1843,15 @@ is extensible in a number of different ways:
 
       <ul>
         <li>
+The requirement to enable extensible data description is provided 
+through the use of [[!JSON]], which allows description of objects,
+lists, and literals in a human and machine readable fashion. JSON
+objects allow extensions through the addition of properties within
+the object. A data processor MAY ignore any fields that it doesn't
+understand.
+        </li>
+
+        <li>
 The requirement to model complex multi-entity relationships is
 provided through the use of a graph-based data model.
         </li>
@@ -1919,12 +1874,12 @@ variety of signature suites.
 The requirement to provide all of the extensibility mechanisms
 outlined above in a data format that is popular among software
 developers and web page authors is enabled via the use of
-[[!JSON-LD]].
+[[!JSON]].
         </li>
       </ul>
 
       <p>
-This approach to data modeling is often called an "open world
+One approach to data modeling is often called an "open world
 assumption", meaning that anyone can say anything about any other
 thing. This approach often feels in conflict with building simple
 and predictable software systems. Balancing extensibility with
@@ -1934,7 +1889,9 @@ assumption than it is with closed software systems.
 
       <p>
 The rest of this section describes how both extensibility and program
-correctness are achieved through a series of examples.
+correctness are achieved through a series of examples using [[!JSON-LD]].
+Note that key JSON-LD fields such as @context are considered extensions
+themselves and MUST be ignored by any processor not using JSON-LD.
       </p>
 
       <p>
@@ -1943,7 +1900,6 @@ Let us assume that we start with the following <a>DID document</a>:
 
       <pre class="example nohighlight" title="A simple DID document">
 {
-  "@context": "https://example.org/example-method/v1",
   "id": "did:example:123456789abcdefghi",
   "publicKey": [{ <span class="comment">...</span> }],
   "authentication": [{ <span class="comment">...</span> }],
@@ -1955,7 +1911,7 @@ The contents of the <code>publicKey</code>,
 <code>authentication</code>, and <code>service</code> properties are
 not important for the purposes of this section. What is important is
 that the object above is a valid <a>DID document</a>. Let's assume
-that a developer wanted to extend the <a>DID document</a> to express
+that a developer wanted to extend the <a>DID document</a> using JSON-LD to express
 an additional piece of information: the subject's public photo
 stream.
       </p>


selfissued:
One key thing that's missing from this PR is a statement that JSON members that are not understood by an implementation MUST be ignored.  That's the foundation to enabling future extensions to not break existing deployments.  Can you please add it?  Thanks.

SmithSamuelM:
Yes!

SmithSamuelM:
@selfissued 

To clarify, Must be Ignored element does not mean reject the entire document but accept the document while eliding those not understood elements.  Must be ignored prevents a class of potential security exploits for those implementations that do not understand a given extension while still allowing extensibility by those implementations that do understand it.

dlongley:
I would prefer to see the plain JSON descriptions and examples first, prior to considering removal of `@context`. The whole argument from the LD-side is that once the plain JSON descriptions are agreed upon it should be obvious that requiring `@context` has no serious disadvantage for the plain JSON community -- rather, it only increases interoperability between it and the LD community.

I could potentially be swayed to support an issue marker stating our intent to add a requirement for `@context` back in should the above hold true once completed. In that case, I would prefer that we not remove all the content that is being deleted here and instead move it to a separate section on how to make your DID Document compatible with the LD community, i.e., just add `@context` with the specified value -- and that's it. That section should note that *not* adding `@context` simply means you won't have interop with the LD community (and we must explain why that's a good idea if we're going to accept it instead of just requiring interop).

cboscolo:
@dlongley, based on answers to my questions in #142, my problem with always requiring `@context` is that my DID method cannot have a method-specific section in the DID Document unless it also adds proper JSON-LD profiles to the `@context`. We do not want to take a dependency on JSON-LD.

I think the only path to ensure everyone can process the DID Documents is to require everyone to process them without an `@context`.

dlongley:
@cboscolo,

> @dlongley, based on answers to my questions in #142, my problem with always requiring `@context` is that my DID method cannot have a method-specific section in the DID Document unless it also adds proper JSON-LD profiles to the `@context`. We do not want to take a dependency on JSON-LD.

While it is ill advised to put method-specific data into the DID Document like that, you could still do it whether or not `@context` was required.

jricher:
@selfissued I commented above that exactly such a statement needs to be added but I didn't know where within the document it should go. It's currently in the extensibility section but there's likely a better place. I would love to see any suggestions you have for that.

SmithSamuelM:
@dlongley 

> just add @context with the specified value -- and that's it. That section should note that not adding `@context` simply means you won't have interop with the LD community (and we must explain why that's a good idea instead of just requiring interop).

I think this is the smoothest path forward. Start with limited assumptions about interop and extensibility mechanisms and then work backwards to what degree of interop is desired at what trade-offs for both encodings. Likewise for the various types of extensibility mechanisms. As apposed to a presumptive approach.  I think there is strong desire to find some middle ground.

cboscolo:
@dlongley wrote:
> While it is ill advised to put method-specific data into the DID Document
☝️ Not _ill-advised_, at all. Method-specific data in the DID Document is going to be very common. Yes, there are elements in the DID that we all want to standardize on.  But there are also elements that will be DID-method specific. In fact I suspect the best way to experiment with new ideas will be to do them first in the DID-method section, and if the global community find it to be useful, they will move to standardize them. (No need to respond to this comment as it is not the main point of this thread, and I don't want to digress any more.)

> you could still do it whether or not `@context` was required.
I really want to understand this statement better.  Could you elaborate on how it can stay in the scenario I described? (Even if just a PM to me.)

dlongley:
@cboscolo,

 > you could still do it whether or not @context was required.
I really want to understand this statement better. Could you elaborate on how it can stay in the scenario I described? (Even if just a PM to me.)

Unless you're using Linked Data Proofs on your DID Documents, there's nothing wrong with adding whatever arbitrary JSON properties and values you want to a DID Document that has an `@context` on it. A JSON-LD processor would completely ignore any properties that are not defined in the `@context`. See: https://www.w3.org/TR/json-ld11/#example-11-term-expansion-from-context-definition

The only time errors are thrown when doing "Linked Data things" is when you're trying to use Linked Data Proofs to digitally sign your data. Under this scenario, the software throws because it wants you to know you're trying to sign something that isn't defined! If you're not doing that, no worries.

jricher:
The fact that this isn't just a change to syntax/encoding tells me that the document, as it currently is written, is a JSON-LD spec. Full stop. This handwaving about it "not requiring JSON-LD" is dishonest and misleading, since if you take out one tiny piece of JSON-LD then suddenly it's a completely different spec to people.

SmithSamuelM:
In discussions about JSON and JSON-LD I often hear JSON-LD proponents state that JSON-LD is "just JSON". This is true in one narrow sense but not true in the broader sense. Which sense is applicable and accurate depends on the conversation.  JSON-LD is parsable JSON syntax. A JSON parser will not throw an error. In this very narrow sense one could say that JSON-LD is "just JSON".  But JSON-LD imbues semantics to JSON parsable symbols that are not "just JSON" in the semantic sense. Semantically JSON-LD is not the same as JSON! And therein lies all the difference.  The narrow sense of "just JSON" syntax is a useful insight when one is trying to convince someone else who shares the semantic RDF world model to adopt JSON-LD over some other REF syntax. One benefits from all the JSON tooling.  But when one is discussing JSON-LD to someone else that does not share the semantic RDF world model then JSON-LD is not "just JSON" anymore because the semantics are different. Full interoperability between the semantic world of JSON and the semantic world of JSON-LD is tantamount to JSON-LD. Including JSON-LD specific symbols but without imbuing them with JSON-LD semantics is at best confusing and at worst misleading. It's code that lies about its intended meaning (semantics).

dlongley:
> The fact that this isn't just a change to syntax/encoding tells me that the document, as it currently is written, is a JSON-LD spec. Full stop. This handwaving about it "not requiring JSON-LD" is dishonest and misleading, since if you take out one tiny piece of JSON-LD then suddenly it's a completely different spec to people.

Again, let's please stop making accusations of dishonesty and bad faith.

I think a number of people have always supported the notion that the spec was leveraging a lot from JSON-LD (and have even argued that in other threads). They have also argued that someone who wants to interpret the spec in terms of plain JSON should also be able to do so. The problem has been that the model hasn't been spelled out explicitly enough in the DID-core spec yet, requiring a reliance on too much knowledge that is external to the DID-core spec. So, what remains to be done is spelling everything out in plain JSON terms in a way that would make plain JSON folks comfortable.

I'm confident that we can also do that in a way that would be compatible with those who want to use JSON-LD so we can have full interop and not split the community. That way we can get buy in from everyone. I think this task of expressing things using plain JSON must be completed *regardless* of any other decision around this topic so we can get to a point where everyone is comfortable and we can reach consensus. This is essentially what I was asking for in my comment as a better way forward vs. just removing what was implicitly there before.

jricher:
I'm making no claims of bad faith. I am saying that it does not line up to claim something has no dependencies on a technology and then also claim it's irreparably broken when that technology is removed. 

What I'm not seeing is how removing the context field from the core spec breaks the JSON-LD version of it.

SmithSamuelM:
@jricher @dlongley 

Might I suggest that the presumption that full interoperability with JSON-LD/RDF semantics is the source of conflict. It is not true that full interoperability with JSON-LD/RDF is desired by all. It is true that if the '@context` is interpreted as an option extension then full interoperability between  JSON-LD/RDF documents and JSON only documents will indeed be broken. Interoperability mechanisms will have to be agreed upon.  For many this is either a desirable feature or a preferred trade-off.  The presumption that RDF is preferred by all is the problematic presumption.  Arguments for JSON-LD syntax based on that presumption (of a shared RDF world view and hence JSON-LD as the preferred syntax) will appear either irrational or illusory to those that do not share that preference.

dlongley:
@jricher,

> I am saying that it does not line up to claim something has no dependencies on a technology and then also claim it's irreparably broken when that technology is removed.
> 
> What I'm not seeing is how removing the context field from the core spec breaks the JSON-LD version of it.

So, to be clear, I believe the claim is that there is a way to add the DID-core spec using plain JSON language that would not alter how people are using it and understanding it today in any significant way -- but that work hasn't been done yet.

*Only* removing `@context` doesn't get us there, it does something else: it changes what the spec means to many people who are already using it today and it doesn't even create a workable spec for plain JSON readers. In other words -- I think there's a way to *add* to the DID-core spec to support plain JSON readers that I believe would work for them *and* that would continue to work for the JSON-LD readers because it would not change what they already believe to be there either explicitly or implicitly (again, we want to move away from the implicit regardless).

I suspect this is also what @peacekeeper is getting at. I would prefer to see a PR that adds what the plain JSON readers think is missing so they can read the spec... plainly. Then we can reconcile any differences and hopefully we'll all be happy with it!

SmithSamuelM:
@dlongley 

I agree that merely removing `@context` doesn't get us all the way there.  But I do believe that making it optional is a necessary first step.  I don't believe as you suggest that starting where we are now with `@context` mandatory and and then merely _adding_  to achieve plain JSON  is a viable path forward.

selfissued:
One of the useful things that https://github.com/w3c/did-core/pull/146 adds is an example of extending JSON in the normal JSON way - by simply adding a field.  The example adds a `photoStreamService` JSON member, which can be used by the implementations that understand it and can be ignored by those that don't.

If we can agree that this is how DIDs will be extended, then I believe we'll be really getting somewhere.  (Yes, we will eventually want to use a registry to register extensions that others want to use; that's something we can easily add once we have consensus for the basic extension mechanism.)

jricher:
To be clear: if this working group wants to declare that this is an RDF-based spec and JSON-LD is the preferred serialization language for that data model, then that's a decision that I could disagree with but live with. (I think it would be a mistake that would limit the draft's utility, but at least it would be honest.) 

What I don't want to see is something that claims to be one thing but really is something else, only to be discovered when someone comes in with the bare set of tools and finds that they need something else. 

Asking a regular JSON developer to even think of namespaces is already a bridge so far away from plain JSON that it's unrecognizable.

TallTed:
> [@selfissued] The example adds a `photoStreamService` JSON member, which can be used by the implementations that understand it and can be ignored by those that don't.

I am sure you meant to say --

_The example adds a `photoStreamService` JSON member, which can be **less safely** used by the implementations that **believe they** understand it and can be **more safely** ignored by those implementations that **recognize that they cannot be sure whether they understand it or not, until/unless there's a registry which defines such extension terms**._

-- because I'm sure you recognize that there's no way to prevent collisions between plain JSON extension terms, since that point has been made several times in this and nearby threads.

peacekeeper:
@SmithSamuelM 

> Might I suggest that the presumption that full interoperability with JSON-LD/RDF semantics is the source of conflict. [...] It is true that if the `@context` is interpreted as an optional extension then full interoperability between JSON-LD/RDF documents and JSON only documents will indeed be broken. Interoperability mechanisms will have to be agreed upon.

I think this is an accurate and helpful analysis of the source of the conflict. The problem I have is that the ongoing discussion has a lot of euphemistic lanaguage around "different encodings", "simplifying things", and "removing unnecessary content", which hide or downplay the fact that what is being proposed is not just to add an additional encoding, but to fork the information model or data model into two incompatible ones with incompatible extensibility models (I would call one a centralized, permissioned extensibility model, and the other one a decentralized, permissionless one).

I don't see how either of the two PRs call out this change that goes beyond just "encodings", or how we would deal with the interoperability gaps that arise from having multiple parallel data models.

iherman:
While we are at trying to spell out the core differences:

> ...but to fork the information model or data model into two incompatible ones with incompatible extensibility models (I would call one a centralized, permissioned extensibility model, and the other one a decentralized, permissionless one).

(From https://github.com/w3c/did-core/pull/145#issuecomment-568205241).

My impression is that the _data model_ is not really different either. There is an abstract model which I believe, is common and could be formulated in an abstract manner: very very it is a simple directed graph, dominated by a tree-like structure with very occasional cross-references among nodes that make it a graph (and the jury may still be out whether those cross-references are indeed necessary).

I believe the core difference is the extensibility model, as formulated by @peacekeeper in that quote. That is what we have to concentrate on, the rest is probably is easier in "just" syntax.

cboscolo:
I have a question for the proponents of keeping `@context` and JSON-LD support in DIDs.

Would it be acceptable to require any JSON-LD extending of the DID Document to be inlined in the `@context`? (That is, not allow it to be extended via a profile that relies on "centralized" infrastructure to process the DID Document using JSON-LD tools?)

msporny:
The recent DID WG F2F meeting in Amsterdam resulted in consensus in how to address the issue that this PR is attempting to address. The result of that discussion is here:

https://www.w3.org/2019/did-wg/Meetings/Minutes/2020-01-30-did#section2-5

The editors are currently working on a set of PRs that will implement the consensus position of the group from that meeting. That consensus-based PR will supersede this PR. 

Closing this PR as it goes in a direction that doesn't implement the current consensus position of the group. Feel free to re-open the issue or object and we can bring it up for discussion during the next meeting.

PR #156, opened 2020-01-25
Update did-v0.11.jsonld to match the current spec

OR13:
Add missing properties to context. 

Absence of these properties severely impacts interoperability when relying on https://www.w3.org/ns/did/v1

And per the current spec: https://w3c.github.io/did-core/#public-keys

`Public keys of all types MUST be expressed in either JSON Web Key (JWK) format using the publicKeyJwk property or one of the formats listed in the table below.`

### DIFF
--- a/contexts/did-v0.11.jsonld
+++ b/contexts/did-v0.11.jsonld
@@ -50,6 +50,8 @@
     "publicKey": {"@id": "sec:publicKey", "@type": "@id", "@container": "@set"},
     "publicKeyBase58": "sec:publicKeyBase58",
     "publicKeyPem": "sec:publicKeyPem",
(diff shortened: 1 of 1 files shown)


OR13:
IPR integration is bugged, my accounts are linked, it will not re-authorize.

OR13:
Absence of these property definitions currently prevents adoption of `https://www.w3.org/ns/did/v1` ... which cannot be relied on until they are added.

OR13:
In the meantime, if you are blocked by this, feel free to add the WIP sidetree extension context, which defines these properties with proper documentation: https://docs.element-did.com/contexts/sidetree/sidetree-v0.1.jsonld

OR13:
I'll wait for a way to fix this in did core.

PR #184, opened 2020-02-05
Rebased version of rhiaro's other PR #123

msporny:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/184.html" title="Last updated on Feb 5, 2020, 9:17 PM UTC (1f6cfc7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/184/bee98cb...1f6cfc7.html" title="Last updated on Feb 5, 2020, 9:17 PM UTC (1f6cfc7)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1366,7 +1366,7 @@ The semantics of the <code>controller</code> property are the same when the
 subject of the relationship is the <a>DID document</a> as when the subject
 of the relationship is a key. Since a key can't control itself, and the
 key controller cannot be inferred from the <a>DID document</a>, it is
-necessary to expicitly express the controller of the key is a
+necessary to explicitly express the identity of the controller of the key
 (which may be the <a>DID controller</a> or <a>DID subject</a>). The
 difference is that the value of <code>controller</code> on a key is not
 necessarily a <a>DID controller</a>. <a>DID controller(s)</a> are
--- a/index.html
+++ b/index.html
@@ -43,6 +43,8 @@
     // previousPublishDate:  "1977-03-15",
     // previousMaturity:  "WD",
 
+    pluralize: true,
+
     // extend the bibliography entries
     localBiblio: ccg.localBiblio,
     github: {
--- a/index.html
+++ b/index.html
@@ -1304,6 +1304,7 @@ Public Keys
       </h2>
 
       <p>
+A public key is a <a>verification method</a>.
 Public keys are used for digital signatures, encryption and other cryptographic
 operations, which in turn are the basis for purposes such as authentication
 (see Section <a href="#authentication"></a>) or establishing secure
@@ -1312,6 +1313,19 @@ communication with <a>service endpoints</a> (see Section
 authorization mechanisms of <a>DID</a> CRUD operations (see Section
 <a href="#did-operations"></a>), which can be defined by <a>DID method</a>
 specifications.
+      </p>
+
+      <p class="note" title="Verification methods">
+A public key is just one type of <a>verification method</a>. Ideally,
+public keys are expressed in a <a>DID document</a> according to what
+they are authorized for: <code>authentication</code>, <code>
+capabilityInvocation</code>, <code>capabilityDelegation</code>,
+<code>keyAgreement</code>, <code>assertionMethod</code>, and <em>not</em>
+just listed using the <code>publicKey</code> property. The latter
+implies the <a>DID controller</a> has authorized any key therein for any
+purpose. It's an ambient authority model which is dangerous. Instead, the
+<a>DID controller</a> should be explicit about what the keys can be used
+for using these relations; see <a href="#authentication"></a>.
       </p>
 
       <p>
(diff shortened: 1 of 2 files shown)


iherman:
The issue was discussed in a [meeting](https://www.w3.org/2019/did-wg/Meetings/Minutes/2021-03-11-did-topic#section4) on 2021-03-11

- no resolutions were taken

### 4. what goes into v1 DID Context? 

_See github issue [#184](https://github.com/w3c/did-core/issues/184)._



**Manu Sporny:** what goes into the v1 DID context. just stuff in the core spec, the things to make 80% of people's lives using it easier (crypto suites, linked domain stuff)? where do we draw the line?  
… We do need to define the context and what goes in it and what doesn't.  

**Orie Steele:** big one: i think we've learned a lot since the VC spec happened. i think the one thing i feel the strongest about is: you have to plan for v2 when you do v1.  

> *Manu Sporny:* +1 to plan for v2 when you do v1

**Orie Steele:** We have to think about when will v2 exist; what will people do in the years between v1 and v2  
… How are we going to make that not super painful so that they feel the spec is usable  
… VC Data Model terrible in this regard. didn't have spec registries to float an unstable version number  

> *Drummond Reed:* I am actually serious that I have to drop a few mins early. Besides the topic Manu is raising, the other one we should focus on is determining what additional policies are needed for managing DID Spec Registries on an ongoing basis.

**Orie Steele:** Proposal that we have a context that only represents DID Core terms, and a second one floated with version process, regularly cut releases as part of registry maintenance  

**Markus Sabadello:** should crypto suites be in security vocab rather than did core?  
… like VCs to make VC context just about VC data model, have crypto suites, key types in the separate security context  
… wonder if also do that with DID core context. more a question than opinion  

**Brent Zundel:** +1, simplest line to draw: this is what is in DID Core  
… extensibility to ease pain of maintaining VC spec  

**Manu Sporny:** the vc v1 context: we did not do a good job there. need to learn from mistakes made there.  
… glad to hear Orie say let's make this easier. trying to move away from kitchen sink contexts.  
… tried to do that in VC and it blew up  
… need to understand how people will use the DID Context.  
… some DID methods will use the base context and will layer their own context on it. leave it to the DID methods to specify what to include.  

> *Orie Steele:* examples: [https://w3id.org/security/bbs/v1](https://w3id.org/security/bbs/v1) - [https://w3id.org/security/jws/v1](https://w3id.org/security/jws/v1) - [https://w3id.org/security/pgp/v1](https://w3id.org/security/pgp/v1)

**Manu Sporny:** presume they will have their own DID method context to layer on top. feels like a good middle ground.  
… Another extreme: we could make it so that every tiny little crypto suite is split into two things: the crypto suite stuff you would put in the DID document, and what you would put in the credential  
… context entry becomes 4,6,8+ entries deep. don't like that  
… would like a base DID context, and one up to the DID method that they can iterate at their own pace.  

**Orie Steele:** +1 to what manu said, i think, the best way is to have the base context only thing in did core, method-specific context things the method has added, and any third ones specific to controller if allowed, other extensions, etc. probably should only have 2 or 3 urls in that list if you are doing it right. we should explain and recommend that as a best practice  

> *Manu Sporny:* +1 to what Orie just said.

> *Orie Steele:* This would be best : "https://www.w3.org/ns/did/v1" - "https://method.example/ns/did/v1" - "https://w3id.org/security/pgp/v1"

**Ivan Herman:** want to understand exactly what we say. base context: the v1 currently in our examples. what you mean is that that includes the properties and things only defined in the core document. is this what you mean?  
… or it includes those and the registered properties?  
… for the time being, there are a bunch of things neither in core or registry - the security things. makes me more nervous  

**Markus Sabadello:** another question for those who know this better: would we consider including other contexts in the did core context, e.g. the security context?  

> *Orie Steele:* NOOOOOO do not nest contexts

**Orie Steele:** markus_sabadello: definitely not. one of my least favorite thing is waiting for deeply nested contexts to resolve  

> *Charles Lehner:* method extension contexts, currently seen in security vocab. after that, things not currently registered anywhere

**Orie Steele:** context to include things in did core, things not defined in did core might be defined in method or seucirt vocab or somewhere else, and that is it  

> *Markus Sabadello:* Which one of the three contexts above defines Ed25519VerificationKey2018?

> *Manu Sporny:* I believe that this is what is being proposed:

```
"@context": [
   "https://w3id.org/did-method-example/v2.1",
   "https://w3id.org/some-extension-example/v3.4"
]
```

> *Manu Sporny:* markus_sabadello, the second one would define `Ed25519VerificatioNKey2018`

**Ivan Herman:** one more issue. some of us will hopefully work on chartering a working group for linked data signature. issue of security-related terms will come up there. we may end up with a context there. don't want them to end up in the core context of this WG. one more consideration to remember  

**Brent Zundel:** closing the meeting.  
… thank you for your contributions  

---


</details>

PR #197, opened 2020-02-14
Add producer/consumer language

jricher:
This adds language to describe the nature of representations as well as subsections for production and consumption for each representation. I have proposed text for the JSON section, and I want to point out the fact that the production and consumption aspects are in fact subtly different from each other. I also attempted to add text for the JSON-LD section, and would be glad to have input from JSON-LD developers about what is missing here. I have left the CBOR empty for the moment (but with stubs).

I removed much of the previous JSON-LD section as the text in question was either an explanation of JSON-LD (which is not needed in this document), and extolling of the benefits of JSON-LD (which does not belong in this document), or an explanation of how to do extensions with JSON-LD (which model has largely been OBE and will be addressed by a separate Extensibility section). 

I also removed all examples as they are likely out of date and need to be regenerated, as per most of the group's recent conversations. We can add them back in when we know what we want to give an example of. I would recommend that when we do this, we take the SAME EXACT did document and replicate it in all available formats. A kind of Rosetta Stone example for DID Documents.

Apologies if the document format is off, I am not deeply familiar with editing ReSpec to this level.

***
<a href="https://pr-preview.s3.amazonaws.com/jricher/did-core/pull/197.html" title="Last updated on Feb 25, 2020, 4:05 PM UTC (0e6da96)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/197/094e1d9...jricher:0e6da96.html" title="Last updated on Feb 25, 2020, 4:05 PM UTC (0e6da96)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1886,55 +1886,125 @@ the property MUST be a valid JSON-LD proof, as defined by
     <h1>Core Representations</h1>
 
       <p>
-A <a>DID document</a> MUST be a single JSON object conforming to [[!RFC8259]].
-Many of the concepts in this document were introduced by example using the
-JSON-LD syntax, a format for mapping JSON data into the RDF semantic graph
-model, as defined by [[!JSON-LD]]. This section formalizes how the data model
-(described in Sections <a href="#data-model"></a> and
-<a href="#did-documents"></a>) is realized in JSON-LD.
+All concrete representations of a <a>DID document</a> MUST be serialized using a 
+deterministic mapping that is unambiguously able to be parsed into the data model
+defined in this specification. All serialization methods MUST define rules for 
+the bidirectional translation of a <a>DID document</a> both into and out of the
+representation in question. As a consequence, translation between any two formats
+MUST be done by first parsing the source format into a <a>DID document</a> model
+(described in Sections <a href="#data-model"></a> and <a href="#did-documents"></a>)
+and then serializing the <a>DID document</a> back out into the target format.
       </p>
+
       <p>
-Although syntactic mappings are provided for JSON and JSON-LD only, applications
-and services can use any other data representation syntax, such as JXD (JSON XDI
-Data, a serialization format for the
-<a href="http://docs.oasis-open.org/xdi/xdi-core/v1.0/csd01/xdi-core-v1.0-csd01.xml">XDI graph model</a>),
+Although syntactic mappings are provided for JSON and JSON-LD here, applications
+and services MAY use any other data representation syntax, such as
 XML, YAML, or CBOR, that is capable of expressing the data model.
+      </p>
+      
+      <p>
+The production and consumption rules in this section apply to all implementations
+seeking to be fully compatible with independent implementations of the specification.
+Deployments of this specification MAY use a custom agreed-upon representation.
+See section ?? on Extensibility for more information.
       </p>
 
     <section>
       <h2>
-Pure JSON
+JSON
       </h2>
 
+<section>
+      <h3>Production</h3>
+
         <p>
-The data model, as described in Section <a href="#data-model"></a>, can be
-encoded in Javascript Object Notation (JSON) [[RFC8259]] by mapping property
+A <a>DID document</a> MUST be a single JSON object conforming to [[!RFC8259]]. All
+top-level properties of the <a>DID document</a> MUST be represented by using the
+property name as the name of the member of the JSON object. The values of
+properties of the data model described in Section <a href="#data-model"></a>, 
+including all extensions, MUST be encoded in JSON[[RFC8259]] by mapping property
 values to JSON types as follows:
         </p>
 
         <ul>
           <li>
-Numeric values representable as IEEE754 SHOULD be represented as a Number type.
+Numeric values representable as IEEE754 MUST be represented as a Number type.
           </li>
           <li>
-Boolean values SHOULD be represented as a Boolean type.
+Boolean values MUST be represented as a Boolean type.
           </li>
           <li>
-Sequence value SHOULD be represented as an Array type.
+Sequence value MUST be represented as an Array type.
           </li>
           <li>
-Unordered sets of values SHOULD be represented as an Array type.
+Unordered sets of values MUST be represented as an Array type.
           </li>
           <li>
-Sets of properties SHOULD be represented as an Object type.
+Sets of properties MUST be represented as an Object type.
           </li>
           <li>
-Empty values SHOULD be represented as a null value.
+Empty values MUST be represented as a null value.
           </li>
           <li>
 Other values MUST be represented as a String type.
           </li>
         </ul>
+
+        <p>
+All properties of the <a>DID document</a> MUST be included in the object.
+        </p>
+
+        <p>
+The property name <code>@context</code> MUST NOT be used as this is reserved
+for JSON-LD processors. 
+        </p>
+        
+</section>
+
+<section>
+  <h3>Consumption</h3>
+  
+    <p>
+The top-level element MUST be a JSON object. Any other data type at the top 
+level is an error and MUST be rejected. The top-level JSON object is taken 
+as the <a>DID document</a>. All members of this object are properties of the
+<a>DID document</a>. The object member name is the property name, and the
+member value is interpreted as follows:
+    </p>
+
+        <ul>
+          <li>
+Number types MUST interpreted as numeric values representable as IEEE754.
+          </li>
+          <li>
+Boolean types MUST be interpreted as a Boolean value.
+          </li>
+          <li>
+An Array type MUST be interpreted as a Sequence or Unordered set, depending on the
+definition of the property for this value.
+          </li>
+          <li>
+An Object type MUST be interpreted as a sets of properties.
+          </li>
+          <li>
+A null type MUST be interpreted as an Empty value.
+          </li>
+          <li>
+String types MUST be interpreted as Strings, which may be further parsed
+depending on the definition of the property for this value.
+          </li>
+        </ul>
+
+  <p>
+The value of the <code>@context</code> field MUST be ignored.
+  </p>
+  
+  <p>
+Unknown object member names MUST be ignored as unknown properties.
+  </p>
+  
+</section>
+      
     </section>
 
     <section>
@@ -1944,34 +2014,9 @@ JSON-LD
 
       <p>
 [[!JSON-LD]] is a JSON-based format used to serialize
-<a href="http://www.w3.org/TR/ld-glossary/#linked-data">Linked Data</a>. The
-syntax is designed to easily integrate into deployed systems already using
-JSON, and provides a smooth upgrade path from JSON to JSON-LD. It is primarily
-intended to be a way to use Linked Data in Web-based programming environments,
-to build interoperable Web services, and to store Linked Data in JSON-based
-storage engines.
-      </p>
-
-      <p>
-JSON-LD is useful when extending the data model described in this specification.
-Instances of the data model are encoded in JSON-LD in the same way they are
-encoded in JSON (see Section <a href="#pure-json"></a>), with the addition of
-the <code>@context</code> property. The
-<a href="https://www.w3.org/TR/json-ld/#the-context">JSON-LD Context</a>
-is described in detail in the [[!JSON-LD]] specification and its use is
-elaborated on in Section <a href="#extensibility"></a>.
-      </p>
-
-      <p>
-In general, the data model and syntaxes described in this document are designed
-such that developers can copy and paste examples into their software systems.
-The design goal of this approach is to provide a low barrier to entry while
-still ensuring global interoperability between a heterogeneous set of software
-systems. This section describes some of these approaches, which will likely go
-unnoticed by most developers, but whose details will be of interest to
-implementers. Noteworthy features provided by JSON-LD are:
-      </p>
-
+<a href="http://www.w3.org/TR/ld-glossary/#linked-data">Linked Data</a> in
+a JSON-based format.
+  </p>
       <ul>
         <li>
 The <code>@id</code> and <code>@type</code> keywords are aliased to
@@ -1983,222 +2028,68 @@ Data types, such as integers, dates, units of measure, and URLs, are
 automatically typed to provide stronger type guarantees for use cases that
 require them.
         </li>
-        <li>
-The <code>@protected</code> properties feature of JSON-LD 1.1 is used to ensure
-that terms defined by this specification cannot be overridden. This means that
-as long as the same <code>@context</code> declaration is made at the top of a
-<a>DID document</a>, interoperability is guaranteed between implementations
-that use a JSON-LD processor and implementations that do not.
-        </li>
       </ul>
-
       <section>
-        <h2>
-  Contexts
-        </h2>
+        <h2>Production</h2>
 
         <p>
-  When two software systems need to exchange data, they need to use terminology
-  and a protocol that both systems understand. The <code>@context</code>
-  property ensures that two systems operating on the same DID document are using
-  mutually agreed terminology.
-        </p>
-
-        <p>
-  <a>DID documents</a> MUST include the <code>@context</code> property.
-        </p>
-
-        <p class="note" title="The JSON-LD Context">
-  More information about the
-  <a href="https://www.w3.org/TR/json-ld/#the-context">JSON-LD Context</a> in
-  general can be found in the [[JSON-LD]] specification.
+The <a>DID document</a> is serialized following the rules in the JSON processor, with
+one addition: <a>DID documents</a> MUST include the <code>@context</code> property.
         </p>
 
         <dl>
           <dt><dfn>@context</dfn></dt>
           <dd>
   The value of the <code>@context</code> property MUST be one or more <a>URIs</a>,
-  where the value of the first <a>URI</a> is
+  where the value of the one <a>URI</a> is
+  <code>https://www.w3.org/ns/did/v1</code>. All members of the <code>@context</code>
+  property MUST exist be in the DID properties extension registry. 
+          </dd>
+        </dl>
+      </section>
+
+      <section>
+        <h2>
+  Consumption
+        </h2>
+
+    <p>
+The top-level element MUST be a JSON object. Any other data type at the top 
+level is an error and MUST be rejected. This top-level JSON object is interpreted
+using JSON-LD processing under the rules of the defined <code>@context</code> fields.
+    </p>
+
+        <p>
+        <dl>
+          <dt><dfn>@context</dfn></dt>
+          <dd>
+  The value of the <code>@context</code> property MUST be one or more <a>URIs</a>,
+  where the value of the one <a>URI</a> is
   <code>https://www.w3.org/ns/did/v1</code>. If more than one <a>URI</a> is
   provided, the <a>URIs</a> MUST be interpreted as an ordered set. It is
   RECOMMENDED that dereferencing each <a>URI</a> results in a document containing
   machine-readable information about the context.
           </dd>
         </dl>
-
-        <pre class="example nohighlight">
-  {
-    "@context": "https://www.w3.org/ns/did/v1"
-  }
-        </pre>
+        </p>
 
         <p>
-  <a>DID method</a> specifications MAY define their own JSON-LD contexts. However,
-  it is NOT RECOMMENDED to define a new context unless the context is necessary to
-  properly implement the method. Method-specific contexts MUST NOT override the
-  terms defined in the generic <a>DID</a> context.
+      Unknown object member names MUST be ignored as unknown properties.
         </p>
       </section>
-
-      <section>
-        <h2>JSON-LD Extensibility</h2>
-
-        <p>
-One of the goals of the <a>decentralized identifiers</a> data model is to
-enable permissionless innovation. The data model needs to be extensible in a
-number of different ways, including:
-        </p>
-
-        <ul>
-          <li>
-Modelling of complex multi-entity relationships is provided through the use of a
-graph-based data model.
-          </li>
-
-          <li>
-Extending the machine-readable vocabularies used to describe information in the
-data model, without relying on a centralized system, is accomplished through
-the use of [[LINKED-DATA]].
-          </li>
-
-          <li>
-Support for multiple types of cryptographic proof formats is accomplished
-through the use of Linked Data Proofs [[LD-PROOFS]], Linked Data Signatures
-[[LD-SIGNATURES]], and a variety of signature suites.
-          </li>
-
-          <li>
-All of the extensibility mechanisms outlined above are provided in a data format
-that is popular among software developers and web page authors and is enabled
-through the use of [[!JSON-LD]].
-          </li>
-        </ul>
-
-        <p>
-This approach to data modeling is often called an <em>open worldassumption</em>,
-meaning that anyone can say anything about any other thing. This approach often
-seems in conflict with building simple and predictable software systems.
-Balancing extensibility with program correctness is always more challenging with
-an open world assumption than it is with closed software systems.
-        </p>
-
-        <p>
-The rest of this section describes how both extensibility and program
-correctness are achieved through a series of examples.
-        </p>
-
-        <p>
-Let us assume that we start with the following <a>DID document</a>.
-        </p>
-
-        <pre class="example nohighlight" title="A simple DID document">
-{
-  "@context": "https://example.org/example-method/v1",
-  "id": "did:example:123456789abcdefghi",
-  "publicKey": [{ <span class="comment">...</span> }],
-  "authentication": [{ <span class="comment">...</span> }],
-  "service": [{ <span class="comment">...</span> }]
-}
-        </pre>
-
-        <p>
-The contents of the <code>publicKey</code>, <code>authentication</code>, and
-<code>service</code> properties are not important for the purpose of this
-section. What is important is that the object above is a valid
-<a>DID document</a>. Now assume that a developer wants to extend the
-<a>DID document</a> to express an additional piece of information, the subject's
-public photo stream.
-        </p>
-
-        <p>
-The first thing that a developer would do is create a JSON-LD Context containing
-the new term.
-        </p>
-
-        <pre class="example nohighlight" title="An example JSON-LD Context">
-{
-  "@context": {
-    "PhotoStreamService": "https://example.com/vocab#PhotoStreamService"
-  }
-}
-        </pre>
-
-        <p>
-Now that the JSON-LD Context is created, the developer MUST publish it somewhere
-that is accessible to any <a>DID document</a> processor. For this example,
-assume that the JSON-LD Context above is published at
-<code>did:example:contexts:987654321</code>. At this point, extending the first
-example in this section is just a matter of including the context above and
-adding the new property to the <a>DID document</a>.
-        </p>
-
-        <pre class="example nohighlight" title="A DID document with a custom extension">
-{
-  "@context": "https://example.org/example-method/v1",
-  "id": "did:example:123456789abcdefghi",
-  "authentication": [ <span class="comment">...</span> ],
-  "service": [<span class="highlight">{
-    "@context": "did:example:contexts:987654321",
-    "id": "did:example:123456789abcdefghi#photos",
-    "type": "PhotoStreamService",
-    "serviceEndpoint": "https://example.org/photos/379283"
-  }</span>]
-}
-        </pre>
-
-        <p>
-The examples so far show that it is easy to extend the
-<a>decentralized identifiers</a> data model in a permissionless and
-decentralized way. The mechanism also ensures that
-<a>decentralized identifiers</a> created in this way prevent namespace conflicts
-and semantic ambiguity.
-        </p>
-
-        <p>
-An extensibility model this dynamic does increase the implementation burden.
-Software written for such a system needs to determine if accepting
-<a>DID documents</a> with extensions is acceptable based on the risk profile of
-the application. Some applications might choose to accept but ignore extensions,
-others might choose to only accept certain extensions, while highly secure
-environments might disallow extensions. These decisions are up to the
-application developers and are specifically not the domain of this
-specification.
-        </p>
-
-        <p>
-Implementations MUST produce an error when an extension JSON-LD Context
-overrides the expanded URL for a term specified in this specification. To avoid
-the possibility of accidentally overriding terms, developers SHOULD scope their
-extensions. For example, the following extension scopes the new
-<code>PhotoStreamService</code> term so that it can only be used within the
-<code>service</code> property:
-        </p>
-
-        <pre class="example nohighlight" title="Scoping terms in a JSON-LD Context">
-{
-  "@context": {
-    <span class="highlight">"service": {
-      "@id": "https://w3id.org/did#service",
-      "@context": {
-        "PhotoStreamService": "https://example.com/vocab#PhotoStreamService"
-      }
-    }</span>
-  }
-}
-        </pre>
-
-        <p>
-Developers are urged to ensure that extension JSON-LD Contexts are highly
-available. Implementations that cannot fetch a context will produce an error.
-Strategies for ensuring that extension JSON-LD Contexts are always available
-include using content-addressed URLs for contexts, bundling context documents
-with implementations, or enabling aggressive caching of contexts.
-        </p>
-      </section>
+      
     </section>
 
     <section>
       <h2>CBOR</h2>
+    </section>
+
+    <section>
+        <h3>Production</h3>
+    </section>
+
+    <section>
+        <h3>Consumption</h3>
     </section>
 
   </section>
--- a/index.html
+++ b/index.html
@@ -3165,272 +3165,272 @@ document</a>.
       </p>
     </section>
 </section>
-    <section>
-      <h2>
+<section>
+    <h2>
 JSON LD Description and Extension
-      </h2>
-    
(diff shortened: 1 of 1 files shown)


iherman:
@jricher you should go to https://www.w3.org/users/112833 and, if you edit the profile, you should 'connect' your account with your github name. (The 'connected account' entry is on the left column of the page.) The IPR system will then consider you as a bona fide member of the WG for PR-s, too...

(Sorry it is so complicated. It has to be done only once.)

index.html
```diff
@@ -1832,55 +1832,125 @@ <h2>Proof</h2>
...
+      <p>
+The production and consumption rules in this section apply to all implementations
+seeking to be fully compatible with independent implementations of the specification.
+Deployments of this specification MAY use a custom agreed-upon representation.
+See section ?? on Extensibility for more information.
+      </p>
 
     <section>
       <h2>
-Pure JSON
```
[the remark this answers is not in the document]
jricher: I believe the initial change was a mistake and did not reflect group consensus. One group member asked to change it against the stated position of many other members of the group. I believe the editor's desire to not hold up the previous pull request on this one term is reasonable, but I hope we can undo that here.

index.html
```diff
@@ -1832,55 +1832,125 @@ <h2>Proof</h2>
...
+
         <p>
-The data model, as described in Section <a href="#data-model"></a>, can be
-encoded in Javascript Object Notation (JSON) [[RFC8259]] by mapping property
+A <a>DID document</a> MUST be a single JSON object conforming to [[!RFC8259]]. All
+top-level properties of the <a>DID document</a> MUST be represented by using the
+property name as the name of the member of the JSON object. The values of
+properties of the data model described in Section <a href="#data-model"></a>, 
+including all extensions, MUST be encoded in JSON[[RFC8259]] by mapping property
 values to JSON types as follows:
```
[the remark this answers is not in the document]
jricher: There are a lot of details for the JSONLD that I am likely missing, as I am not as familiar with that specification. I repeated the text as a starting point for that greater specificity. However, it is also my understanding that all the types of fields would be accounted for by the JSONLD context, which is now required to be processed here. Remember, these are rules for processing JSONLD documents, and so we can require anyone doing so to follow JSONLD rules and expectations. I will make that more clear in the introduction text.

index.html
```diff
@@ -1832,55 +1832,125 @@ <h2>Proof</h2>
...
-Unordered sets of values SHOULD be represented as an Array type.
+Unordered sets of values MUST be represented as an Array type.
           </li>
           <li>
-Sets of properties SHOULD be represented as an Object type.
+Sets of properties MUST be represented as an Object type.
           </li>
           <li>
-Empty values SHOULD be represented as a null value.
+Empty values MUST be represented as a null value.
```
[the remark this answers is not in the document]
jricher: This was in the previous text. I agree that "empty value" is not well defined in our document model, but I think the attempt here was to be able to represent null values for properties, as opposed to omitted properties. We need to define, within our document, what the datatypes we care about are. If there is one named "empty" or "null" or "undefined" or whatever, it needs to be mapped in each format.

index.html
```diff
@@ -1832,55 +1832,125 @@ <h2>Proof</h2>
...
           </li>
           <li>
 Other values MUST be represented as a String type.
           </li>
         </ul>
+
+        <p>
+All properties of the <a>DID document</a> MUST be included in the object.
```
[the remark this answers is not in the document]
jricher: The intended requirement here is actually to say that to make a DID document, you put all of the properties of the DID document into the object, and not into sub-objects or a top-level array with key fields or whatever other weird format people might try if we don't explicitly specify this mapping. I'll reword to make this clearer.

index.html
```diff
@@ -1832,55 +1832,125 @@ <h2>Proof</h2>
...
+A null type MUST be interpreted as an Empty value.
+          </li>
+          <li>
+String types MUST be interpreted as Strings, which may be further parsed
+depending on the definition of the property for this value.
+          </li>
+        </ul>
+
+  <p>
+The value of the <code>@context</code> field MUST be ignored.
```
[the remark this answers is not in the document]
jricher: If they try to "upconvert to JSONLD" then they are doing it wrong. 

What we're saying is that a JSON-formatted DID Document cannot rely on the context field. If you want to rely on the context field, then you need to create a JSONLD-formatted DID document and follow those rules.

index.html
```diff
@@ -1832,55 +1832,125 @@ <h2>Proof</h2>
...
+  </p>
+  
+  <p>
+Unknown object member names MUST be ignored as unknown properties.
```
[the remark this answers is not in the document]
jricher: I don't agree with your change. What does it meant to "Remove" a property on consumption? How is that different from ignoring the property and its value?

As for the approach to unknown properties overall: I agree that there are several approaches and the group needs to pick one, and exactly one, to be used across all formats. What was discussed in Amsterdam is ignoring unknown properties, and I think that's a reasonable starting point for the discussion. 

While I don't believe that the statement as written is what the group has agreed to, it would be a very bad idea to not have a requirement in this section. I would argue that we need to start with something and move from there. Right now we have nothing at all, and these sections don't make sense without having :some: rule.

jricher:
I disagree that this is a big pull request -- in fact, I believe it's the smallest possible set of changes to add the structure as discussed.

index.html
```diff
@@ -1832,55 +1832,125 @@ <h2>Proof</h2>
...
+Although syntactic mappings are provided for JSON and JSON-LD here, applications
+and services MAY use any other data representation syntax, such as
 XML, YAML, or CBOR, that is capable of expressing the data model.
       </p>
+      
+      <p>
+The production and consumption rules in this section apply to all implementations
+seeking to be fully compatible with independent implementations of the specification.
+Deployments of this specification MAY use a custom agreed-upon representation.
+See section ?? on Extensibility for more information.
```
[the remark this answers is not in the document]
jricher: I don't think the section I am referring to actually exists yet. I left a dangling pointer here to make sure this pull request wasn't doing too much at once.

index.html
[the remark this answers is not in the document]
jricher: Regardless, if @selfissued objects to changing this back, he can chime in here again.

index.html
```diff
@@ -1832,55 +1832,125 @@ <h2>Proof</h2>
...
 
+<section>
+      <h3>Production</h3>
+
         <p>
-The data model, as described in Section <a href="#data-model"></a>, can be
-encoded in Javascript Object Notation (JSON) [[RFC8259]] by mapping property
+A <a>DID document</a> MUST be a single JSON object conforming to [[!RFC8259]]. All
+top-level properties of the <a>DID document</a> MUST be represented by using the
+property name as the name of the member of the JSON object. The values of
```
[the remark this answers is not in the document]
jricher: We need a way to refer to the key and the value of each thing in the DID document. The best term I could find from the document model section was "property", where the "name" of the property is the string value listed in the document, and the "value" of the property is whatever it's designed to convey.

index.html
```diff
@@ -1890,34 +1960,9 @@ <h2>
 
       <p>
 [[!JSON-LD]] is a JSON-based format used to serialize
-<a href="http://www.w3.org/TR/ld-glossary/#linked-data">Linked Data</a>. The
```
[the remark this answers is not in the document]
jricher: I disagree with this approach, since all the text is available in document history and it makes no sense outside of its original section. Still, I have added it back in to a new appendix as requested.

index.html
```diff
@@ -1929,224 +1974,71 @@ <h2>
...
           <dd>
   The value of the <code>@context</code> property MUST be one or more <a>URIs</a>,
-  where the value of the first <a>URI</a> is
-  <code>https://www.w3.org/ns/did/v1</code>. If more than one <a>URI</a> is
-  provided, the <a>URIs</a> MUST be interpreted as an ordered set. It is
-  RECOMMENDED that dereferencing each <a>URI</a> results in a document containing
-  machine-readable information about the context.
+  where the value of the one <a>URI</a> is
+  <code>https://www.w3.org/ns/did/v1</code>. All members of the <code>@context</code>
+  property MUST exist be in the DID properties extension registry.
```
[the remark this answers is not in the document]
jricher: This is about processing globally interoperable documents. This is a MUST.

index.html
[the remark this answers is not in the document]
jricher: What you're reading from this statement is exactly correct: this is about JSON consumers, not JSONLD consumers.

To be more specific, the following requirement earlier in the section is intended to prevent exactly this kind of cross-format upconversion:

> translation between any two representations
MUST be done by parsing the source format into a <a>DID document</a> model
(described in Sections <a href="#data-model"></a> and <a href="#did-documents"></a>)
and then serializing the <a>DID document</a> model into the target representation.

So in other words, if I have a signal that the document is JSON but it has a context field, then I am not allowed to process it as JSONLD and I need to ignore the context field, which should not have been included in the first place because it's forbidden by the JSON production rules, above.

If I've got a document that's JSONLD, then I need to read the context field, as in the next section. And if I'm producing JSONLD, then I need to include the context field. 

This is why we need to have document metadata to convey representation information between producers and consumers. I did not want to add that to the document with this pull request as that is a separate issue and I strove to make this pull request as small as actionably possible.

index.html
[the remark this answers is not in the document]
jricher: Specifically, the pull request already addresses this with the requirement:

> Deployments of this specification MAY use a custom agreed-upon representation.

I would argue that a representation that uses unknown properties is using its own representation and is outside the scope of this standard.

index.html
```diff
@@ -1832,55 +1832,125 @@ <h2>Proof</h2>
...
+  
+    <p>
+The top-level element MUST be a JSON object. Any other data type at the top 
+level is an error and MUST be rejected. The top-level JSON object is taken 
+as the <a>DID document</a>. All members of this object are properties of the
+<a>DID document</a>. The object member name is the property name, and the
+member value is interpreted as follows:
+    </p>
+
+        <ul>
```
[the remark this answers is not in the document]
jricher: It's not a representation of the above. As noted in the pull request header, this is from a different perspective. If you compare the text of the two lists, you will see they are different in subtle but important ways. Namely, the directionality of the mapping is opposite, and that has consequences for what's on either side of the mapping. 

The previous version of the specification attempted to collapse this into a single list, which is the mistake I am trying to correct here.

index.html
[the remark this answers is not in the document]
jricher: So while my previous comment holds true in general for the JSON-LD section (which is incomplete), I realize now that your comment was in fact in response to the JSON section, not the JSON-LD section. In which case, this paragraph and the following list are absolutely vital and important to this specification's usefulness.

What's going on here is we are mapping between data types that we use within the DID document model to specific serialization structures within JSON. The second list tells us how to get back from JSON structures into the data types used within the DID document model. 

It is not enough to say "Just put it into JSON!" and hope for the best. Otherwise, what's to stop me from taking something that most people would do as this:

```json
{
   "foo": "bar",
   "baz": "qux"
}
```

And instead using this:

```json
[
   ["foo", "bar"],
   ["baz", "qux"]
]
```

Both are perfectly valid JSON, and could be parsed into the same underlying structure, but only one would follow the rules of this specification as I've written them out. Since this is a standard, we have the duty to be completely pedantic about how every transformation needs to work.

index.html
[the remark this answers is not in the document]
msporny: Hmm, if the section doesn't exist, then add something like this:

```
<p class="issue">
A link to a yet-to-be-written section on representation extensibility 
should be added here once that section has been written.
</p>
```

Bonus points for creating an issue and using `data-number` to point to the issue number:

```
<p class="issue"  data-number="123">
```

index.html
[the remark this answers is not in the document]
msporny: Final ping to @selfissued before this is changed back from "Pure JSON" to "JSON".

index.html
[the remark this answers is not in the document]
msporny: Ah, ok, now I understand. JSON uses "name" and "value". Javascript uses "key" and "value". RDF uses "subject" "predicate/property" "value". Some of us with feet in all of those camps have been using "property" and "value" with some success because it maps semi-cleanly between various communities (JSON/Javascript/RDF). "property name" and "name of the member" crosses the streams. 

Let's just use "property" and "value" in the spec... the VC spec survived multiple reviews by doing things like "X property" and "Y property" and doing things like "The value of the X property...".

index.html
```diff
@@ -1832,55 +1832,141 @@ <h2>Proof</h2>
...
+
+<section>
+      <h3>Production</h3>
 
         <p>
-The data model, as described in Section <a href="#data-model"></a>, can be
-encoded in Javascript Object Notation (JSON) [[RFC8259]] by mapping property
+A <a>DID document</a> MUST be a single JSON object conforming to [[!RFC8259]]. All
+top-level properties of the <a>DID document</a> MUST be represented by using the
+property name as the name of the member of the JSON object. The values of
```
```suggestion
property as the name of the member of the JSON object. The values of
```

index.html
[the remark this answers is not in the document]
msporny: > what's to stop me from taking something that most people would do as this

Well, the rule you wrote that says the top level thing MUST be a JSON object :), but I get your point. 

> Since this is a standard, we have the duty to be completely pedantic about how every transformation needs to work.

I'm fine w/ being completely pedantic for now and simplifying later if we can.

index.html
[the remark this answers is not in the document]
msporny: Can you mark this with an issue stating that we are considering not supporting null values in the abstract data model? No need for native support there, and if there is a need, we have other ways to represent null.

index.html
```diff
@@ -1832,55 +1832,125 @@ <h2>Proof</h2>
...
+</section>
+
+<section>
+  <h3>Consumption</h3>
+  
+    <p>
+The top-level element MUST be a JSON object. Any other data type at the top 
+level is an error and MUST be rejected. The top-level JSON object is taken 
+as the <a>DID document</a>. All members of this object are properties of the
+<a>DID document</a>. The object member name is the property name, and the
```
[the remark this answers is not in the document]
msporny: We should link to parts of the JSON spec for all the special language we use from there, and this spec for where we use "property". There's a lot of specialized language here that's being used that you need to read the JSON spec to understand. The mechanism to do that is:

```
<a data-cite="RFC8259#section-4">JSON object</a>
```

... and so on. We should also define "property" in the DID Core spec to refer to a property in the abstract data model (or whatever similar concept makes sense).

index.html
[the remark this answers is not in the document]
msporny: Yes, I get that the directionality of the mapping is opposite. I mentioned above that we'll keep this and try to simplify later. We should try to strive to collapse the amount of normative text in the spec if possible.

index.html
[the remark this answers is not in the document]
msporny: > What does it meant to "Remove" a property on consumption? How is that different from ignoring the property and its value?

Consider a piece of software that is a consumer that then produces what it consumes in a different representation. A consumer that ignores a property can pass it through to the different representation. A consumer that removes the property on consumption does not pass it through to the different representation. I'm arguing that we should do the latter... but the more we discuss this, the more I think you need to raise an issue and point to that issue here (with whatever text you want to suggest). Otherwise it looks we're adding text to the spec that the group has agreed to, which it hasn't. 

We typically state that implementers should be aware that the decision about whether to ignore or remove is currently under debate in the WG.

index.html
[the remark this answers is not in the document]
msporny: Ok, good, thanks. 

Also, please don't resolve conversations that you didn't start... the original commenter (me) doesn't see them, so it just looks like you closed the comment w/ no further consideration... and I have to click "unhide conversation" to see what you said.

index.html
```diff
@@ -1929,224 +1974,71 @@ <h2>
 automatically typed to provide stronger type guarantees for use cases that
 require them.
         </li>
-        <li>
```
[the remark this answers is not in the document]
msporny: Please respond to all comments folks raise on your PRs, even if you're copying/pasting... otherwise, I have no idea what action you took (because I may have missed the resolved conversation where you said you took an action). In general... don't resolve conversations you didn't start, and always respond to every comment on PRs so the other party knows you took action (or decided to not take action).

index.html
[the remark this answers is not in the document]
msporny: It's exactly because you put that text in there that I suggested "SHOULD" instead of "MUST"... because the spec allows the use of custom agreed-upon representations, and `@context` makes it easy to do that w/o having to use a registry.

Please raise and issue and mark the "MUST vs. SHOULD" as a point of contention that the WG is currently debating (and point to the issue). Let's avoid a perma-thread on the whole MUST vs. SHOULD debate wrt. entries in the registry by just noting this as an open issue and moving on.

index.html
```diff
@@ -3197,7 +3110,274 @@ <h2>
...
+    
+      <p class="note" title="JSON-LD Description">
+The following text was previously in the section on serialized representations. 
+It has been moved to this appendix to allow the text to be re-incorporated into
+the document in more appropriate sections or removed entirely, as required by
+the working group.
+      </p>
+      
+            <p>
+      [[!JSON-LD]] is a JSON-based format used to serialize
```
msporny: Indentation for this entire section is off, please fix. I know we're just moving it, but sometimes that doesn't end up happening and it gets splattered around the specification. We need to keep the formatting consistent to make the lives of people editing the spec and putting in PRs easier.

index.html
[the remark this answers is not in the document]
jricher: If we're going to raise it as an issue, I argue that we need to start with `MUST` and raise an issue to lower it to `SHOULD` if the WG agrees, for the reasons above. I strongly believe this section needs to have strong normative requirements and as few SHOULD/MAY portions as possible.

index.html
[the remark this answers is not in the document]
jricher: OK, I'll that.

index.html
jricher: I'm ok with this change, but we'll need to clarify that by "property" here we mean the string for the property itself and not any other piece. I'll add an issue (and link) for that here.

index.html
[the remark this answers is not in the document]
jricher: Yes, but VC is much more JSON-LD specific. I think this needs an additional issue (noted below) that I'll raise separately. Thanks!

index.html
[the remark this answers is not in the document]
jricher: If we can make the pedantry simpler without being less pedantic, I'm all for it. :)

index.html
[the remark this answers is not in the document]
jricher: I will add links as above, I agree with that.

index.html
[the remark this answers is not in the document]
jricher: I will raise an issue and point to this text, as it is definitely something the group needs to decide on.

index.html
jricher: OK, I'll fix this -- it was copied from the previous version as-is.

index.html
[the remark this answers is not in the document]
jricher: Yes, will do.

index.html
```diff
@@ -1831,322 +1831,291 @@ <h2>Proof</h2>
...
-          </li>
-        </ul>
-    </section>
+    <p>
+All concrete representations of a <a>DID document</a> MUST be serialized using a 
+deterministic mapping that is able to be unambiguously parsed into the data model
+defined in this specification. All serialization methods MUST define rules for 
+the bidirectional translation of a <a>DID document</a> both into and out of the
+representation in question. As a consequence, translation between any two representations
+MUST be done by parsing the source format into a <a>DID document</a> model
```
msporny: Untestable normative requirement? Struggling to understand how we write a test for this normative statement.

index.html
```diff
@@ -1831,322 +1831,291 @@ <h2>Proof</h2>
...
+(described in Sections <a href="#data-model"></a> and <a href="#did-documents"></a>)
+and then serializing the <a>DID document</a> model into the target representation. An
+implementation MUST NOT convert between representations without first parsing to a
```
msporny: Untestable normative requirement? How would we know if an implementation didn't first parse to a DID Document model?

index.html
```diff
@@ -1831,322 +1831,291 @@ <h2>Proof</h2>
...
-          <li>
-Empty values SHOULD be represented as a null value.
-          </li>
-          <li>
-Other values MUST be represented as a String type.
-          </li>
-        </ul>
-    </section>
+    <p>
+All concrete representations of a <a>DID document</a> MUST be serialized using a
```
msporny: How do we test this normative requirement in the test suite?

index.html
```diff
@@ -1831,322 +1831,291 @@ <h2>Proof</h2>
...
-[[!JSON-LD]] is a JSON-based format used to serialize
-<a href="http://www.w3.org/TR/ld-glossary/#linked-data">Linked Data</a>. The
-syntax is designed to easily integrate into deployed systems already using
-JSON, and provides a smooth upgrade path from JSON to JSON-LD. It is primarily
-intended to be a way to use Linked Data in Web-based programming environments,
-to build interoperable Web services, and to store Linked Data in JSON-based
-storage engines.
-      </p>
+    <p>
+Producers MUST indicate which representation of a document has been used, and consumers
```
msporny: This normative statement isn't clear how producers indicate which representation of a document has been used. My assumption is that there are two ways -- media type and including `@context`. We might want to state that this should only be done via media type?

index.html
```diff
@@ -1831,322 +1831,291 @@ <h2>Proof</h2>
...
+JSON
         </h2>
 
         <p>
-  When two software systems need to exchange data, they need to use terminology
-  and a protocol that both systems understand. The <code>@context</code>
-  property ensures that two systems operating on the same DID document are using
-  mutually agreed terminology.
+When producing and consuming DID documents that are in plain JSON (as noted by
+the document metadata), the following rules MUST be followed.
```
msporny: Hmm, again... do we trust the document metadata or the media type. The more variations there are, the more opportunity for misimplementation. Also, what happens when the media type disagrees with the document metadata? "application/json" media type for document containing "@context". The latter requires content sniffing, which is generally frowned upon. I suggest we specify that the media type is how you detect a plain JSON DID document from a JSON-LD document. We can also say that plain JSON DID documents MUST NOT contain a `@context` property (which I disagree with... suggest downgrading that to SHOULD NOT)... but again, the core issue here is how do you know something is a JSON document or a JSON-LD document? Content sniffing, media type, or both? If both, what are the rules when content sniffing disagrees with media type (typically, content sniffing wins... but again, let's try to avoid that anti-pattern).

index.html
```diff
@@ -1831,322 +1831,291 @@ <h2>Proof</h2>
...
+
+            <p class="issue" data-number="204">
+An "empty" value is not specified by this document. It seems to imply a null
+value, but this is unclear.
+            </p>
+
+            <p>
+All applicable properties of the <a>DID document</a> MUST be included in
+the root object. Properties MAY define additional data sub structures subject
+to the value representation rules, above.
```
```suggestion
to the value representation rules described in the list above.
```

index.html
```diff
@@ -1831,322 +1831,291 @@ <h2>Proof</h2>
...
+                </li>
+            </ul>
+
+            <p class="issue" data-number="204">
+An "empty" value is not specified by this document. It seems to imply a null
+value, but this is unclear.
+            </p>
+
+            <p>
+All applicable properties of the <a>DID document</a> MUST be included in
```
```suggestion
All mandatory properties of the <a>DID document</a> MUST be included in
```

Not sure what "applicable" is attempting to imply? Feels like a no-op -- "The things that apply to the DID Document must be included in the DID Document" -- is that's what's being said?

index.html
```diff
@@ -1831,322 +1831,291 @@ <h2>Proof</h2>
...
+        </section>
+
+        <section>
+            <h3>
+Consumption
+            </h3>
+
+            <p>
+The top-level element MUST be a JSON object. Any other data type at the top 
+level is an error and MUST be rejected. The top-level JSON object is taken
```
```suggestion
level is an error and MUST be rejected. The top-level JSON object is 
```

index.html
```diff
@@ -1831,322 +1831,291 @@ <h2>Proof</h2>
...
+
+        <section>
+            <h3>
+Consumption
+            </h3>
+
+            <p>
+The top-level element MUST be a JSON object. Any other data type at the top 
+level is an error and MUST be rejected. The top-level JSON object is taken 
+as the <a>DID document</a>. All members of this object are properties of the
```
```suggestion
the <a>DID document</a>. All members of this object are properties of the
```

index.html
```diff
@@ -1831,322 +1831,291 @@ <h2>Proof</h2>
...
+            <p>
+The <a>DID document</a> is serialized following the rules in the JSON processor, with
+one addition: <a>DID documents</a> MUST include the <code>@context</code> property.
+            </p>
+
+            <dl>
+                <dt><dfn>@context</dfn></dt>
+                <dd>
+The value of the <code>@context</code> property MUST be one or more <a>URIs</a>,
+where the value of the one <a>URI</a> is
```
```suggestion
where the value of the first <a>URI</a> is
```

index.html
```diff
@@ -1831,322 +1831,291 @@ <h2>Proof</h2>
...
+The top-level element MUST be a JSON object. Any other data type at the top 
+level is an error and MUST be rejected. This top-level JSON object is interpreted
+using JSON-LD processing under the rules of the defined <code>@context</code> fields.
+            </p>
+
+            <dl>
+                <dt><dfn>@context</dfn></dt>
+                <dd>
+The value of the <code>@context</code> property MUST be one or more <a>URIs</a>,
+where the value of the one <a>URI</a> is
```
```suggestion
where the value of the first <a>URI</a> is
```

msporny:
If you could rebase this PR and use full sentences and/or proper punctuation/capitalization in the commit messages, I'd appreciate it.  So this:

```
fix indentation of representation section
```

becomes this:

```
Fix indentation of representation section.
```
I'd like to preserve all of your commits when I rebase and don't want to have to go and rewrite master history to do so (and make the changes above)... since I don't have access to rewrite your branch.

index.html
jricher: This is exactly what the test suite is for. We give a producer a data model and tell it to produce a document which we need to be able to parse into an unambiguously equal data model.

index.html
jricher: Not every requirement can be tested automatically. That doesn't make the requirement meaningless.

index.html
jricher: We can use an implementation to translate between two formats by extracting the data model in the middle of the process and examining the model. This is not necessarily something that can be automated. Ultimately, the test suite should include corner cases that would mess up people who are taking shortcuts, like seeing a JSON document and throwing their own "context" field in there.

index.html
jricher: The media type is part of the metadata, and I would argue the most important part right now. That's why this section needs to be written up. See #203 (which is linked below), the "representation type" is the only one specifically named.

Detecting the presence or absence of a context field, or any other internal field, is an anti pattern and needs to be not just avoided but forbidden by the spec's definition of a consumer. Not only is it fragile with the JSON/JSON-LD case, it completely falls apart as soon as you talk about any other representation type, at all.

index.html
jricher: See above -- this is why we need document metadata, which includes the document media type. They will never be in conflict with each other because they aren't different things. 

The requirement to include the document type in this metadata information is exactly to avoid the kind of content sniffing you're talking about.

index.html
jricher: I'll remove the qualifier. If the group wants to define a qualifier due to discussion on issue #204 then it can be added back in.

index.html
jricher: I'll change this, but it was my understanding that order was not important, so this was meant to be "the value of one URI" (the "the" is extraneous in my edit).

index.html
jricher: This sentence will be reworded better.

index.html
msporny: Order is vitally important... we're using the JSON-LD 1.1 "@protected" feature, which disallows a term that's defined from being redefined (so people don't accidentally stomp on the base DID context). That means that the base DID context goes first. It also allows JSON-only processors to just check a known order for the `@context` property... but that's no longer a thing, so is less important (maybe).

In any case, order matters, the DID Context must be the first value in the ordered set.

index.html
msporny: > Not every requirement can be tested automatically. That doesn't make the requirement meaningless.

Over the last decade or so, W3C WGs have, in general, started to move away from including normative language that isn't testable in a test suite. W3C WGs are going this direction because of historical fighting over what an untestable normative statement actually means and what should be done if an implementation doesn't do the untestable normative thing. By and large, these normative statements are useless. There are exceptions, like normative language about how you should write a specification -- the DID Method section in this spec is a good example of this sort of language... but requiring a human being to step in and interpret whether or not an implementation is conformant is generally regarded as an anti-pattern.

Normative statements that require human intervention don't get put in test suites and thus those statements are effectively useless over time because developers run their implementations against a test suite, and if the test suite passes, the devs move on.

I was seeing if you had a plan about how this statement would be tested, and I'm hearing "we're not going to test this statement in the test suite", which automatically puts it into the "then let's not make it a normative requirement". You could restate it as this:

> translation between any two representations **is expected to** be done by parsing the source format into a <a>DID document</a> model

I suggest "is expected to" because there may be other valid ways to do this, like through some XSLT-like process that can do a direct translation w/o "going into" the DID document model (whatever that means).

I can merge this as is, but I expect that we'll remove it in time because it's not saying anything useful from a testability perspective.

index.html
msporny: "extracting the data model" -- but the data model is abstract? If you are saying that the concrete test that we can write to test this statement is injecting `@context` in a application/did+json document and feeding it to a consumer and seeing if it exposes @context in the data model, then I can buy that as a concrete test that we can write to check this normative requirement.

However, that's a really roundabout way of testing that and you have other normative requirements in the spec that would test that more directly... raising the question about whether or not this should be normative. You are also assuming the internals of an implementation and that's typically not a good thing to do when writing a spec. We're defining the data model and interfaces to an implementation, we shouldn't be defining how an implementation actually works internally. We can describe how we expect it to work, but as long as the outcome complies with the spec, that's what should matter.

You could reword as "When converting between representations, implementations are advised to parse to the data model in this specification and then generate the output representation by using the data model."

We can test your ignore `@context` if you're processing application/did+json in a separate test that is more aligned w/ normative language that you added later in this PR.

index.html
msporny: > We give a producer a data model

How exactly are we doing that? What does the input file look like?

index.html
msporny: > Detecting the presence or absence of a context field, or any other internal field, is an anti pattern and needs to be not just avoided but forbidden by the spec's definition of a consumer. Not only is it fragile with the JSON/JSON-LD case, it completely falls apart as soon as you talk about any other representation type, at all.

Right, I agree with you completely, which is why I think we might want to reword the above from:

> Producers MUST indicate which representation of a document has been used, and consumers

to this:

> Producers MUST indicate representations of this data model via a media type (see Section E...). Consumers MUST determine representations of this data model via the media type (see Section E...).

index.html
msporny: > document metadata

I suggest that we stop using this term... it's confusing, because there is more than one type of document metadata. To me, media type is different from document metadata (the first being specific, the second being this amorphous term that means different things to different people).

index.html
jricher: To be completely clear: this has no effect on JSON consumers (who ignore the property) or JSON producers (who don't output the property), so it's entirely up to the JSONLD definition, in which order is semantically important.

index.html
jricher: I will not add language to enable content sniffing nor will I sit by to see it added here.

index.html
jricher: Media type is one field of document metadata. Until we have a more concrete term for the metadata mechanism, I suggest we use this one.

index.html
jricher: I get what you're saying here and above. What I'm after is a requirement to prevent people from doing a shortcut process to translate between forms. This is a very big issue with writing abstract data models as specifications -- fundamentally, you can't test anything in an abstract data model, only in the concrete representations of those models.

msporny: Yes, and my point is that the text in the PR creates a normative requirement that involves the abstract data model in a way that's not testable (or testable via a more concrete test later in the document). In any case, I think you get the point, I'm fine w/ merging as is... but I expect future reviews (when working on the test suite) to remove this normative language from the spec (but keep the intent intact).

index.html
msporny: We could just avoid using the "document metadata" term entirely because it looks like all of your language just requires the decision of what representation you're using to be triggered off of the media type (nothing else is taken into consideration), which I think is good. We don't need to talk about document metadata (because we don't really care how it's signaled to the consumer as long as it's signaled clearly).

I don't feel very strongly about this, just noting that words are being used w/o a clear definition... and we'll have to fix that later.

msporny:
> I've replied to your comments and will apply several of your suggested changes.

@jricher are we good to go? I'm ready to merge.

jricher:
I think we're good. There are definitely still things for the working group to decide but I think this is a big step in a positive direction.

@iherman  I did the IPR link but it's not showing up, I'm not sure why it's not picking up on it but we can look into it.

index.html
jricher: I agree that it needs to be fixed later, but I think that it needs to be fixed by defining the metadata.

msporny:
Normative, substantive, multiple review cycles, comments addressed by PR author, merging.

msporny:
> merging

Oh FFS... Github APIs are down.

msporny:
> merging

Oh FFS... Github APIs are down.

PR #219, opened 2020-03-12
Add support for relative DID URLs. Fix #201.

### LINKED ISSUE #201: Add resolution procedure for relative URL references
A more general form of this was raised (in https://github.com/w3c-ccg/did-spec/issues/80 and https://github.com/w3c/did-core/issues/11) requesting that the group considers how relative URL references are resolved against a BASE URL (presumably, the DID subject identified by the DID Document).

The specification is currently silent wrt. what the BASE URL is and how relative URLs are resolved against that BASE URL.

We should specify that the BASE URL is the value of the top-level `id` property in the DID Document and relative URLs are resolved by normalizing them according to standard path/query transforms and then truncating the result to the BASE URL (which is more or less what happens in HTTP, Javascript libraries, and HTML links... so developers won't be surprised by the transformations).

msporny:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/219.html" title="Last updated on Mar 19, 2020, 12:39 AM UTC (c8a2508)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/219/414798f...c8a2508.html" title="Last updated on Mar 19, 2020, 12:39 AM UTC (c8a2508)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1128,10 +1128,11 @@ contain relative path components, query parameters, and fragment identifiers.
         <p>
 When resolving a relative DID URL reference, the algorithm specified in <a
 data-cite="RFC3986#section-5">RFC3986 Section 5: Reference Resolution</a>  MUST
-be used. The <strong>base URI</strong> value is the <a>DID subject</a>.  The
-<strong>scheme</strong> is <code>did</code>. The <strong>authority</strong> is a
-combination of <code>&lt;method-name>:&lt;method-specific-id></code>, and the
-<strong>path</strong>, <strong>query</strong>, and <strong>fragment</strong>
+be used. The <strong>base URI</strong> value is the <a>DID</a> that is
+associated with the <a>DID subject</a>, see Section <a href="#did-subject"></a>.
+The <strong>scheme</strong> is <code>did</code>. The <strong>authority</strong>
+is a combination of <code>&lt;method-name>:&lt;method-specific-id></code>, and
+the <strong>path</strong>, <strong>query</strong>, and <strong>fragment</strong>
 values are those defined in Section <a href="#path"></a>, Section <a
 href="#query"></a>, and Section <a href="#fragment"></a>, respectively.
         </p>
--- a/index.html
+++ b/index.html
@@ -1113,6 +1113,59 @@ usually contains cryptographic material enabling authentication of the
 
       </section>
 
+      <section>
+        <h2>Relative DID URLs</h2>
+
+        <p>
+A relative DID URL is any URL value in a <a>DID document</a> that does not start
+with <code>did:&lt;method-name>:&lt;method-specific-id></code>. More
+specifically, it is any URL value that does not start with the ABNF defined  in
+Section <a href="#generic-did-syntax"></a>. The contents of the URL typically
+refers to a resource in the same <a>DID document</a>. Relative DID URLs MAY
+contain relative path components, query parameters, and fragment identifiers.
+        </p>
+
+        <p>
+When resolving a relative DID URL reference, the algorithm specified in <a
+data-cite="RFC3986#section-5">RFC3986 Section 5: Reference Resolution</a>  MUST
+be used. The <strong>base URI</strong> value is the <a>DID subject</a>.  The
+<strong>scheme</strong> is <code>did</code>. The <strong>authority</strong> is a
+combination of <code>&lt;method-name>:&lt;method-specific-id></code>, and the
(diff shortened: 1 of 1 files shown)


csuwildcat:
Looks good to me.

index.html
```diff
@@ -1113,6 +1113,59 @@ <h2>Fragment</h2>
...
+specifically, it is any URL value that does not start with the ABNF defined  in
+Section <a href="#generic-did-syntax"></a>. The contents of the URL typically
+refers to a resource in the same <a>DID document</a>. Relative DID URLs MAY
+contain relative path components, query parameters, and fragment identifiers.
+        </p>
+
+        <p>
+When resolving a relative DID URL reference, the algorithm specified in <a
+data-cite="RFC3986#section-5">RFC3986 Section 5: Reference Resolution</a>  MUST
+be used. The <strong>base URI</strong> value is the <a>DID subject</a>.  The
```
dhh1128: I think this is another example where we are conflating the DID value and the DID subject. We drove that out of the spec once. The DID subject is the thing identified by the DID, not the identifier value. So I think this should say that the base URI is the DID value. Is that accurate?

index.html
```diff
@@ -1113,6 +1113,59 @@ <h2>Fragment</h2>
...
+<span class="comment">    // a relative DID URL used to reference a verification method above</span>
+    "<span class="highlight">#key-1</span>"
+  ]
+}
+        </pre>
+
+        <p>
+In the example above, the relative DID URL value will be transformed to an
+absolute DID URL value of <code>did:example:123456789abcdefghi#key-1</code>.
+
```
davidlehn: Remove blank line.
```suggestion
```

index.html
```diff
@@ -1113,6 +1113,58 @@ <h2>Fragment</h2>
...
+specifically, it is any URL value that does not start with the ABNF defined  in
+Section <a href="#generic-did-syntax"></a>. The contents of the URL typically
+refers to a resource in the same <a>DID document</a>. Relative DID URLs MAY
+contain relative path components, query parameters, and fragment identifiers.
+        </p>
+
+        <p>
+When resolving a relative DID URL reference, the algorithm specified in <a
+data-cite="RFC3986#section-5">RFC3986 Section 5: Reference Resolution</a>  MUST
+be used. The <strong>base URI</strong> value is the <a>DID subject</a>.  The
```
```suggestion
be used. The <strong>base URI</strong> value is the <a>DID value</a>.  The
```

OR13: I agree with @dhh1128

index.html
msporny: Yes, will fix.

index.html
msporny: Hmm, DID value isn't defined in our terminology. We had this discussion a while ago to not define "XYZ value" because where do we stop if we start there? Just about every property in the spec has a corresponding value. I've reworded this to address @dhh1128 core concern c8a2508970c4c1a435f72d3b0b9dfb5a136c33ed.

msporny:
Normative, multiple reviews, changes requested and made, merging.

csuwildcat:
Can someone tell me how I can use the relative base URI stuff to get rid of the need to have a verbose controller value in the case that it is the same as the base URI value of the top-level `id` property?

OR13:
you pretty much can't... this exact issue just caused serious broken interop issues between companies working with DHS, and `did:elem`... TL;DR; use URIs for `id`...

csuwildcat:
@OR13 I am not sure why language could not be added to describe the absence of a `controller` field as assuming the top-level id/controller is the one who controls the key. We certainly could do it, so I would like to know why. Having to paste larger IDs all over the place during resolution is something I would like to avoid.

OR13:
@csuwildcat controller presence is a different issue... I can tell you that from what I have seen, its required state, appears to be related to JSON-LD processing of the DID Document... its absence can cause Linked Data Signature libraries to fail...  for interop, and as the spec currently states:

``` 
Each public key object MUST have the type, controller, and specific public key properties, and MUST have an id property. The public key objects MAY include additional properties.
```

^ these also apply to other `verificationMethod`s as far as I can tell.

Speaking generally, as DID Method implementer trying to implement "Pure JSON" and  "JSON-LD" representations, I would object pretty strongly to creating more optionality in the did core spec... the more optionality, the weaker the spec, the harder to implement, the more likely to fail... 

We should probably be having this discussion on an issue, not a closed pull request.

Issue #174, opened 2020-01-31
Underspecified semantics of "updated" property


selfissued:
The specification does not give any sense of what kinds of actions constitute an update to the DID that should result in a change to the "updated" time.  For instance, does resolving the DID constitute an update?  This property will be virtually useless unless there's a common understanding of what actions constitute an update.

rhiaro:
I believe this would depend on the particular DID method, and should be specified in the DID method specification, but this should be clarified.

jandrieu:
This is a meta-data issue.

OR13:
@csuwildcat you may want to add some details about sidetree here... 

Sidetree is very restrictive in terms of what can be updated.

selfissued:
Even if this is method specific, the core spec needs to add the requirement on method specifications that they clearly define what `updated` means and does.

OR13:
I consider DID Methods that don't allow a controller to make full content updates to be a separate class of methods... as a did controller, I like to be able to control :)

peacekeeper:
"Update" is one of the 4 operations that DID methods must define. But I think the semantics of "updated" are NOT method-specific. I agree with @jandrieu that this is related to metadata, see https://github.com/w3c/did-spec/issues/65

csuwildcat:
> @csuwildcat you may want to add some details about sidetree here...
> 
> Sidetree is very restrictive in terms of what can be updated.

This is not true. The base protocol for Sidetree only strictly handles the DID Document modification keys, because it has to know which are allowed to modify the Doc. The base Sidetree protocol allows a Sidetree-based method implementer to merge literally anything they want into a DID Doc, or limit what can be put in on a per-Method basis, for example: one Sidetree-based DID Method could disallow string values in the doc larger than 1k, because they want to keep people from dropping a gig of base64'd cat pics into a DID Doc.

Sidetree does not limit you as an implementer, you decide this aspect of your Method. Some will allow a free-for-all and get dunked on if they ever achieve any significant scale, while others will be more specific/deliberate in how each of the DID Document's fields/values is modified. This doesn't mean you are not the controller of the Doc because the protocol you use attempts to codify rules that hinder dropping a GB of cat pics on the network.

csuwildcat:
For this question, of what Update means, I believe it should be defined as follows:

"An Update to a DID shall be considered any change made to a DID, in the scope of its Method implementation, that results in the modification of values that control operations on the DID or what is produced from resolution of the DID, such as the output DID Document."

[...]

PR #220, opened 2020-03-15
Add language regarding DID Method Update Operation

OR13:
- Add language about what an "Update" is.

Based on feedback here: https://github.com/w3c/did-core/issues/174

***
<a href="https://pr-preview.s3.amazonaws.com/OR13/did-core/pull/220.html" title="Last updated on Mar 21, 2020, 7:09 PM UTC (47382b3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/220/414798f...OR13:47382b3.html" title="Last updated on Mar 21, 2020, 7:09 PM UTC (47382b3)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2335,7 +2335,7 @@ updates are not possible.
         </p>
 
         <p>
-          An Update to a DID is any change in the data used to produce a DID Document. 
+          An update to a DID is any change in the data used to produce a DID Document. 
           DID Method Implementers are responsible for defining what constitutes an update, 
           and what properties of the DID Document are supported by a given DID Method. 
           There might be cases where an update occurred but the <a>DID document</a> representation does not change.
--- a/index.html
+++ b/index.html
@@ -2338,7 +2338,7 @@ updates are not possible.
 An update to a DID is any change, after creation, in the data used to produce a DID Document. 
 <a>DID Method</a> implementers are responsible for defining what constitutes an update, 
 and what properties of the DID Document are supported by a given <a>DID Method</a>. 
-For example, an update operation which replaces key material without changing it could be a valid update that doesn't result in changes to the <a>DID Document</a>.
+For example, an update operation which replaces key material without changing it could be a valid update that does not result in changes to the <a>DID document</a>.
         </p>
       </section>
 
--- a/index.html
+++ b/index.html
@@ -2335,7 +2335,7 @@ updates are not possible.
         </p>
 
         <p>
-An update to a DID is any change, after creation, in the data used to produce a DID Document. 
+An update to a <a>DID</a> is any change, after creation, in the data used to produce a <a>DID document</a>. 
 <a>DID Method</a> implementers are responsible for defining what constitutes an update, 
 and what properties of the <a>DID document</a> are supported by a given <a>DID method</a>. 
 For example, an update operation which replaces key material without changing it could be a valid update that does not result in changes to the <a>DID document</a>.
--- a/index.html
+++ b/index.html
@@ -2335,10 +2335,10 @@ updates are not possible.
         </p>
 
         <p>
-          An update to a DID is any change in the data used to produce a DID Document. 
-          <a>DID method</a> implementers are responsible for defining what constitutes an update, 
-          and what properties of the DID Document are supported by a given DID Method. 
-          There might be cases where an update occurred but the <a>DID document</a> representation does not change.
+An update to a DID is any change in the data used to produce a DID Document. 
+<a>DID method</a> implementers are responsible for defining what constitutes an update, 
+and what properties of the DID Document are supported by a given DID Method. 
+There might be cases where an update occurred but the <a>DID document</a> representation does not change.
         </p>
       </section>
 
--- a/index.html
+++ b/index.html
@@ -1796,7 +1796,7 @@ and <a href="#authentication"></a>.
       <h2>Created</h2>
 
       <p>
-A <a>DID document</a> MAY include a <code>created</code> property.
+A <a>DID document</a> SHOULD include a <code>created</code> property.
       </p>
 
       <dl>
@@ -1826,7 +1826,7 @@ change.
       </p>
 
       <p>
-A <a>DID document</a> MAY include an <code>updated</code> property.
+A <a>DID document</a> SHOULD include an <code>updated</code> property.
       </p>
 
       <dl>
--- a/index.html
+++ b/index.html
@@ -2336,7 +2336,7 @@ updates are not possible.
 
         <p>
           An update to a DID is any change in the data used to produce a DID Document. 
-          DID Method Implementers are responsible for defining what constitutes an update, 
+          <a>DID method</a> implementers are responsible for defining what constitutes an update, 
           and what properties of the DID Document are supported by a given DID Method. 
           There might be cases where an update occurred but the <a>DID document</a> representation does not change.
         </p>
--- a/index.html
+++ b/index.html
@@ -2337,7 +2337,7 @@ updates are not possible.
         <p>
 An update to a DID is any change, after creation, in the data used to produce a DID Document. 
 <a>DID Method</a> implementers are responsible for defining what constitutes an update, 
-and what properties of the DID Document are supported by a given <a>DID Method</a>. 
+and what properties of the <a>DID document</a> are supported by a given <a>DID method</a>. 
 For example, an update operation which replaces key material without changing it could be a valid update that does not result in changes to the <a>DID document</a>.
         </p>
       </section>
--- a/index.html
+++ b/index.html
@@ -2338,7 +2338,7 @@ updates are not possible.
           An Update to a DID is any change in the data used to produce a DID Document. 
           DID Method Implementers are responsible for defining what constitutes an update, 
           and what properties of the DID Document are supported by a given DID Method. 
-          There may be cases where an "Update" occurred but the DID Document representation does not change.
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -2331,7 +2331,14 @@ <h3>
...
 operations necessary to establish proof of control, <em>or</em> state that
-updates are not possible.
+updates are not possible. 
+        </p>
+
+        <p>
+          An Update to a DID is any change in the data used to produce a DID Document. 
+          DID Method Implementers are responsible for defining what constitutes an update, 
+          and what properties of the DID Document are supported by a given DID Method. 
+          There may be cases where an "Update" occurred but the did document representation does not change.
```
```suggestion
          There may be cases where an "Update" occurred but the DID Document representation does not change.
```

OR13:
@msporny I don't understand your objection. Both "created" and "updated" are currently marked as "SHOULD". 

I agree with your objection, and the PR is an attempt to move us towards, the perspective you are sharing.

Would you prefer I modify the PR to remove "created" and "updated" entirely?

msporny:
> Would you prefer I modify the PR to remove "created" and "updated" entirely?

Yes, please... you may be unaware that there were PRs already to do this (which failed, blocked by the metadata discussion):

https://github.com/w3c/did-core/pull/28
https://github.com/w3c/did-core/pull/27
https://github.com/w3c/did-core/pull/26

My expectation is that any PR you submit to remove created/updated/proof will be blocked for the same reason. The group is waiting on a resolution to the metadata discussion in order to resubmit those PRs.

Does that make sense?

OR13:
Based on the content of this PR, I would not expect it to be blocked by those... This PR is a step towards removing metadata from the doc... 

If any method decides to include such fields in the did document (did:web), but most others don't, then the spec should definitely not say "SHOULD", and IMO it should say MAY, and specify the timestamp format as it does currently... 

otherwise we are inviting methods to put whatever "metadata" they want in the did document... we are currently encouraging them to do that, with the use of the word "SHOULD".

Removing the properties seems like a change more likely to be objected to than marking them as "MAY" when we know that only a few (1) DID Method wants to use them.

msporny:
> If any method decides to include such fields in the did document (did:web), but most others don't, then the spec should definitely not say "SHOULD", and IMO it should say MAY, and specify the timestamp format as it does currently...

We're still talking past each other.

A subset of the group believe that the DID Document MUST NOT specify created/updated metadata at the top level.

The options on the table right now seem to be:

1. Put metadata at the top-level of the DID Document. This conflates metadata about the DID subject w/ metadata about the DID Document. This led to objections and the "metadata buckets" discussion.
2. Put metadata in a `didDocument` property at the top level. This seems to be an acceptable compromise to both sides of the discussion. Only did:web seems to want to do this at present (based on the last call on this topic), so there was a suggestion for the did:web method to define this property and use it.
3. Do not put metadata in the DID document, and instead make it part of a DID Resolution response, which is what some would prefer as the cleanest option.

Changing the language to MAY is choice # 1 above, which is the point of contention around metadata in the DID document... which led to objections last time, and I suspect will lead to objections this time (because the fundamentals haven't changed around the discussion).

OR13:
@msporny Ahh I see... I have reverted those changes, is the PR now does not comment on metadata properties, only on DID Method Operations.

index.html
```diff
@@ -2331,7 +2331,14 @@ <h3>
...
 operations necessary to establish proof of control, <em>or</em> state that
-updates are not possible.
+updates are not possible. 
+        </p>
+
+        <p>
+          An Update to a DID is any change in the data used to produce a DID Document. 
+          DID Method Implementers are responsible for defining what constitutes an update, 
+          and what properties of the DID Document are supported by a given DID Method. 
+          There may be cases where an "Update" occurred but the DID Document representation does not change.
```
```suggestion
          There might be cases where an update occurred but the <a>DID document</a> representation does not change.
```

We probably don't have `representation` defined in the terminology and should define it... can be done in a different PR.

index.html
```diff
@@ -2331,7 +2331,14 @@ <h3>
 The <a>DID method</a> specification MUST specify how a client can update a
 <a>DID document</a> on the <a>DID registry</a>, including all cryptographic
 operations necessary to establish proof of control, <em>or</em> state that
-updates are not possible.
+updates are not possible. 
+        </p>
+
+        <p>
+          An Update to a DID is any change in the data used to produce a DID Document.
```
msporny: Left justify the contents of the &lt;p> tag... gives Editors more area to work with wrt. spec text.

```suggestion
          An update to a DID is any change in the data used to produce a DID Document. 
```

No need to capitalize 'update'.

index.html
```diff
@@ -2331,7 +2331,14 @@ <h3>
 The <a>DID method</a> specification MUST specify how a client can update a
 <a>DID document</a> on the <a>DID registry</a>, including all cryptographic
 operations necessary to establish proof of control, <em>or</em> state that
-updates are not possible.
+updates are not possible. 
+        </p>
+
+        <p>
+          An Update to a DID is any change in the data used to produce a DID Document. 
+          DID Method Implementers are responsible for defining what constitutes an update,
```
```suggestion
          <a>DID method</a> implementers are responsible for defining what constitutes an update, 
```

index.html
OR13: Can I resolve this?

brentzundel:
The current proposed text is confusing to me. The section seems to be about updating a DID Document, but the text states that it is possible to update a DID Document without it changing.
Perhaps adding an example of an update that doesn't result in changes to the DID Doc would clarify things.

OR13:
@brentzundel 

I agree, but I'm not sure an example would be appropriate here, since it would certainly be DID Method Specific.... This seems to be a case where the implementation guideline would be helpful...

For the sake of clarifying, imagine creating an update operation that is signed which replaces key material without changing the material.. or creating an update operation which is not valid....

Both are updates according to sidetree protocol (did:ion, did:elem, etc...), neither results in a did document change...

I'd prefer not to introduce method specific examples in did core... it invites politics.

brentzundel:
Perhaps just the sentence, "For example, an update operation which replaces key material without changing it could be a valid update that doesn't result in changes to the DID Document."

index.html
```diff
@@ -2331,7 +2331,14 @@ <h3>
 The <a>DID method</a> specification MUST specify how a client can update a
 <a>DID document</a> on the <a>DID registry</a>, including all cryptographic
 operations necessary to establish proof of control, <em>or</em> state that
-updates are not possible.
+updates are not possible. 
+        </p>
+
+        <p>
+An update to a DID is any change in the data used to produce a DID Document.
```
kdenhartog: This sentence feels weird to me because it could be understood that the initial creation of a DID Document is an update.

Nothing => <basic did document>

Is that an intentional choice of language?

OR13: I think the language aligns well with KERI / did:peer / did:ethr  / sidetree, but I'm happy to change it if there is something stronger.

OR13: I added "after creation" to address this concern.

OR13:
@brentzundel thanks, I added you language, much clearer.

index.html
```diff
@@ -2331,7 +2331,14 @@ <h3>
 The <a>DID method</a> specification MUST specify how a client can update a
 <a>DID document</a> on the <a>DID registry</a>, including all cryptographic
 operations necessary to establish proof of control, <em>or</em> state that
-updates are not possible.
+updates are not possible. 
+        </p>
+
+        <p>
+An update to a DID is any change, after creation, in the data used to produce a DID Document.
```
```suggestion
An update to a <a>DID</a> is any change, after creation, in the data used to produce a <a>DID document</a>. 
```

index.html
```diff
@@ -2331,7 +2331,14 @@ <h3>
...
 <a>DID document</a> on the <a>DID registry</a>, including all cryptographic
 operations necessary to establish proof of control, <em>or</em> state that
-updates are not possible.
+updates are not possible. 
+        </p>
+
+        <p>
+An update to a DID is any change, after creation, in the data used to produce a DID Document. 
+<a>DID Method</a> implementers are responsible for defining what constitutes an update, 
+and what properties of the DID Document are supported by a given <a>DID Method</a>.
```
```suggestion
and what properties of the <a>DID document</a> are supported by a given <a>DID method</a>. 
```

index.html
```diff
@@ -2331,7 +2331,14 @@ <h3>
...
 operations necessary to establish proof of control, <em>or</em> state that
-updates are not possible.
+updates are not possible. 
+        </p>
+
+        <p>
+An update to a DID is any change, after creation, in the data used to produce a DID Document. 
+<a>DID Method</a> implementers are responsible for defining what constitutes an update, 
+and what properties of the DID Document are supported by a given <a>DID Method</a>. 
+For example, an update operation which replaces key material without changing it could be a valid update that doesn't result in changes to the <a>DID Document</a>.
```
```suggestion
For example, an update operation which replaces key material without changing it could be a valid update that does not result in changes to the <a>DID document</a>.
```

This line is also not wrapped to less than 80 characters. Please make sure you wrap to below 80 characters to help people that 1) have line wrapping turned off in their editors on purpose, so they can spot long code lines more easily, and 2) are doing reviews so diff-marking works across a variety of tools, where some of them don't make long lines easy to diff.

As a general rule, avoid contractions in spec text -- contractions tend to be easier for non-native readers to miss and are less formal than we should be in a spec.

OR13:
@msporny I have merged your changes, can you review again?

index.html
OR13: Can I resolve this thread?

Issue #131, opened 2019-11-28
inconsistent mental model connecting id, kid, and references inside a DID doc


dhh1128:
Elsewhere (can't find location at the moment but would be glad if others pasted a link...) I saw a discussion about how `id` should not be required on keys, if those keys are expressed in JWK format, since JWKs have a `kid` field. The issue generated lots of discussion, and I'm not sure how it resolved.

Implicit in that discussion is the assumption that a JWK's `kid` and a key's `id` in classic DID doc specs are intended to perform the same function, and that they have compatible semantics. However, it is clear to me that we are not sufficiently aligned as a community around these assertions, as evidenced by the following:

1. A `kid` is a hint used for caching, much like the `etag` header in HTTP. You are supposed to be able to look at the `kid` and say, "Oh, I've already dereferenced that key. No need to dereference again." (@selfissued , please correct me if I'm mischaracterizing the tribal knowledge about `kid` in JW* circles...) But currently, the `id` on a key in a DID doc has explicitly incompatible semantics with this usage, because the value referenced by `did:example:abc123#xyz` is allowed to be different every time it is dereferenced!

This disconnect is not causing the heartburn that it should. People are [proposing to use DID URLs that contain a `kid` as a pointer to a specific key value, even though the DID URL doesn't contain a timestamp or version stamp and is therefore ambiguous](https://github.com/hyperledger/aries-rfcs/issues/133#issuecomment-554496804). Validating signatures and doing other cryptographic operations with ambiguous inputs is the opposite of what DIDs are supposed to deliver, IMO.

2. Setting aside that first concern, I am seeing DID URLs in the form `did:example:abc123#xyz` as references to a key with `kid` = xyz -- as if `kid` could play the same role that `id` plays in fragments. Yet I know of no verbiage in any version of the DID spec that would support this. AFAIK, the only way to refer to a key by `kid` would be something using matrix parameters, like maybe `did:example:abc123;kid=xyz` -- and I'm not even sure that works. @peacekeeper, what does the guru say? So are we playing games with `@context` that map `kid` into the role that `id` plays in more traditional JSON-LD--and if so, where is this documented, and why did we think this was a good idea?

I think something has to give. `kid` and `id` can't exist in free variation with semantics that overlap in some cases but not others.

selfissued:
[...]

PR #221, opened 2020-03-15
Add publicKeyJwk with compound key identifier example and note

OR13:
Addresses https://github.com/w3c/did-core/issues/131

- Add note about public key identifiers and compound keys
- Add publicKeyJwk example

***

### :boom: Error: 504 Gateway Time-out :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Mar 15, 2020, 6:54 PM UTC)_.

<details>
<summary>More</summary>

PR Preview relies on a number of web services to run. There seems to be an issue with the following one:

:rotating_light: [Spec Generator](https://www.w3.org/2015/labs/) - Spec Generator is the web service used to build specs that rely on ReSpec.

:link: [Related URL](https://labs.w3.org/spec-generator/?type=respec&url=https%3A%2F%2Frawcdn.githack.com%2FOR13%2Fdid-core%2F20ae30caf2158026be71d6b45dfc4f362dd53a49%2Findex.html%3FisPreview%3Dtrue)

```
<html><body><h1>504 Gateway Time-out</h1>
The server didn't respond in time.
</body></html>

```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20w3c/did-core%23221.)._
</details>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1303,10 +1303,13 @@ the <a>DID document</a>; see Section <a href="#authorization-and-delegation"></a
       </p>
 
       <p class="note" title="Key Identifiers">
-The value of the <code>id</code> property may be structured as a <a href="https://en.wikipedia.org/wiki/Compound_key">compound key</a>. 
+        
+      
+The value of the <code>id</code> property MAY be structured as a <a href="https://en.wikipedia.org/wiki/Compound_key">compound key</a>. 
 This is especially useful for integrating with existing key management systems and key formats such as <a href="https://tools.ietf.org/html/rfc7517">JWK</a>.
-When possible a public key fingerprint SHOULD be used in the construction of compound keys. 
-However, as is the case with <code>kid</code> in JWK, one cannot rely on the identifier to map directly to the key value.
+It is RECOMMENDED that <a href="https://tools.ietf.org/html/rfc7517">JWK</a> <code>kid</code> values are set to the <a href="https://tools.ietf.org/html/rfc7638">public key fingerprint</a>. 
+It is RECOMMENDED that verification methods that use JWKs to represent their 
+public keys utilize the value of <code>kid</code> as their fragment identifier.
 See the first key in <a href="#example-11-various-public-keys">EXAMPLE 11</a> for an example of a public key with a compound key identifier.
       </p>
 
--- a/index.html
+++ b/index.html
@@ -1305,7 +1305,7 @@ the <a>DID document</a>; see Section <a href="#authorization-and-delegation"></a
       <p class="note" title="Key Identifiers">
 The value of the <code>id</code> property may be structured as a <a href="https://en.wikipedia.org/wiki/Compound_key">compound key</a>. 
 This is especially useful for integrating with existing key management systems and key formats such as <a href="https://tools.ietf.org/html/rfc7517">JWK</a>.
-When possible a thumbprint or fingerprint SHOULD be used in the construction of compound keys. 
+When possible a public key fingerprint SHOULD be used in the construction of compound keys. 
 However, as is the case with <code>kid</code> in JWK, one cannot rely on the identifier to map directly to the key value.
 See the first key in <a href="#example-11-various-public-keys">EXAMPLE 11</a> for an example of a public key with a compound key identifier.
       </p>
--- a/index.html
+++ b/index.html
@@ -1302,9 +1302,7 @@ expressed using the <code>controller</code> property on the top level of
 the <a>DID document</a>; see Section <a href="#authorization-and-delegation"></a>.
       </p>
 
-      <p class="note" title="Key Identifiers">
-        
-      
+      <p>
 The value of the <code>id</code> property MAY be structured as a <a href="https://en.wikipedia.org/wiki/Compound_key">compound key</a>. 
 This is especially useful for integrating with existing key management systems and key formats such as <a href="https://tools.ietf.org/html/rfc7517">JWK</a>.
 It is RECOMMENDED that <a href="https://tools.ietf.org/html/rfc7517">JWK</a> <code>kid</code> values are set to the <a href="https://tools.ietf.org/html/rfc7638">public key fingerprint</a>. 
--- a/index.html
+++ b/index.html
@@ -1304,7 +1304,7 @@ the <a>DID document</a>; see Section <a href="#authorization-and-delegation"></a
 
       <p class="note" title="Key Identifiers">
 The value of the <code>id</code> property may be structured as a <a href="https://en.wikipedia.org/wiki/Compound_key">compound key</a>. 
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1302,6 +1302,14 @@ <h2>Public Keys</h2>
 the <a>DID document</a>; see Section <a href="#authorization-and-delegation"></a>.
       </p>
 
+      <p class="note" title="Key Identifiers">
+The value of the <code>id</code> property may be structured as a <a href="https://en.wikipedia.org/wiki/Compound_key">compound key</a>. 
+This is especially useful for integrating with existing key mangement systems and key formats such as <a href="https://tools.ietf.org/html/rfc7517">JWK</a>.
```
```suggestion
This is especially useful for integrating with existing key management systems and key formats such as <a href="https://tools.ietf.org/html/rfc7517">JWK</a>.
```

index.html
```diff
@@ -1412,6 +1420,16 @@ <h2>Public Keys</h2>
   "id": "did:example:123456789abcdefghi",
   <span class="comment">...</span>
   "publicKey": [{
+    "id": "did:example:123#_Qq0UL2Fq651Q0Fjd6TvnYE-faHiOpRlPVQcY_-tA4A",
```
msporny: I really like this as a best practice for JWK `kid`s in DID documents.

index.html
```diff
@@ -1302,6 +1302,14 @@ <h2>Public Keys</h2>
 the <a>DID document</a>; see Section <a href="#authorization-and-delegation"></a>.
       </p>
 
+      <p class="note" title="Key Identifiers">
+The value of the <code>id</code> property may be structured as a <a href="https://en.wikipedia.org/wiki/Compound_key">compound key</a>. 
+This is especially useful for integrating with existing key management systems and key formats such as <a href="https://tools.ietf.org/html/rfc7517">JWK</a>.
+When possible a thumbprint or fingerprint SHOULD be used in the construction of compound keys.
```
```suggestion
When possible a public key fingerprint SHOULD be used in the construction of compound keys. 
```

index.html
```diff
@@ -1302,6 +1302,14 @@ <h2>Public Keys</h2>
 the <a>DID document</a>; see Section <a href="#authorization-and-delegation"></a>.
       </p>
 
+      <p class="note" title="Key Identifiers">
+The value of the <code>id</code> property may be structured as a <a href="https://en.wikipedia.org/wiki/Compound_key">compound key</a>.
```
msporny: On first read, this felt a bit strange to introduce the concept using database terminology. I'm wondering if there is an easier way to introduce this. Like, just tell people what the best practice is first and /then/ state that they can think of the example as analogous to compound database keys (although, I question the value of the explanation)... usually, unless the text adds something of great value, we can skip it.

msporny: Is this "may" normative? If not, replace it w/ a non-normative word... like "might", but that creates a problem w/ the SHOULD below... makes it seem less strongly suggestive.

OR13: I'm expecting an objection from @selfissued regarding introducing more undefined terminology.

OR13: can a note contain such language? I will make it normative, I'd prefer it to be.

msporny: Yeah, good point. Typically, no, a note shouldn't contain normative language. I've seen some specs sneak a couple of those in. 

If it were me, I'd remove `class="note"` -- this is normative and it should clearly state that.

OR13:
@msporny I implemented your feedback. I'm a fan of using existing terms whenever possible, with your changes hopefully its acceptable.

index.html
OR13: can I resolve this?

index.html
OR13: Can I resolve this?

index.html
```diff
@@ -1302,6 +1302,15 @@ <h2>Public Keys</h2>
 the <a>DID document</a>; see Section <a href="#authorization-and-delegation"></a>.
       </p>
 
+      <p>
+The value of the <code>id</code> property MAY be structured as a <a href="https://en.wikipedia.org/wiki/Compound_key">compound key</a>. 
+This is especially useful for integrating with existing key management systems and key formats such as <a href="https://tools.ietf.org/html/rfc7517">JWK</a>.
+It is RECOMMENDED that <a href="https://tools.ietf.org/html/rfc7517">JWK</a> <code>kid</code> values are set to the <a href="https://tools.ietf.org/html/rfc7638">public key fingerprint</a>. 
+It is RECOMMENDED that verification methods that use JWKs to represent their
```
dhh1128: This second recommendation surprises me. I like it, but it surprises me. The reason is that most examples I've seen use `id` instead -- and indeed, that seems to jive with what the relative URL language assumes.

Everything else in a DID document will use `id` as the fragment identifier. The `id` field is intentionally NOT pointing to immutable content; that is, by design, we want did:example:abc#key1 to be able to point to various, different values over its lifespan. I detest this design choice, considered it a mistake, and argued against it, but was overruled.

Recommending that we make the `kid` property the fragment identifier for a JWK would be great, except that it's so at odds with everything else we're doing in the doc.

dhh1128: If we keep this, can we *also* recommend that we NOT use an `id` property in the same JWK? It seems very messy to me to have a JWK have both, and yet to have its fragment identifier derive from `kid`. A recipe for errors.

dlongley: The `id` attribute is for the verification method, the JWK is the value of `publicKeyJwk`. IMO, it's best that we keep consistency with `id` for verification methods. There's a desire here for some linkage, if possible, between `kid` and `id`, but it may not always be possible.

OR13: indeed, per the RFC linked... `kid` can be whatever people want... just like `id`.... https://tools.ietf.org/html/rfc7638#section-3.1

`fingerprint` or `kid` are used in the existing systems... `verificationMethod.id` is used with resolvers... we need these to be different or we are going to lots of problems... here is a more complete example of publicKeyJwk integration with JWKS...

https://github.com/transmute-industries/lds-jws2020/blob/master/docs/example/didDocJwks.json

https://github.com/transmute-industries/lds-jws2020/blob/master/docs/example/didDoc.json

You can imagine that you might use the same JWKS with multiple DIDs... if you chose to implement a custom `kid`... now not only can you not use the caching properties associated with it... but you can't actually support 2 did documents with the same JWKS....

Thats why we are trying to recommend that they follow the convention of the following format (regardless of key representation)... 

`did:example:123#fingerprint`... in the case of did:key with publicKeyBase58, the fingerprint is calculated from multibase / multcodec... in the case of publicKeyJwk, its per the RFC...

@dlongley maybe it would help to add the recommendation regarding  publicKeyBase58?

index.html
OR13: Can I resolve this thread?

Issue #168, opened 2020-01-31
Underspecified semantics of revoked keys


selfissued:
Section 6.3 on Public Keys https://w3c.github.io/did-core/#public-keys contains these statements about revoked keys in DID documents:
> The DID document MAY contain revoked keys. A DID document containing a revoked key MUST also contain or refer to the revocation information for the key (for example, a revocation list). Each DID method specification is expected to detail how revocation is performed and tracked.

This is so underspecified as to be an open invitation for non-interoperability.  If every method defines different syntax and semantics for public keys in DID documents, there's not much point of even talking about revoked keys in the specification.

The one actionable statement on revoked keys is this one, and I'm fine with it remaining:
> If a public key does not exist in the DID document, it MUST be assumed the key was revoked or is invalid.

OR13:
Any did methods actually embedding revoked keys? I'm in favor of not allowing them in the document.

jandrieu:
They address the use case of retaining keys that should still be considered valid for actions in the past, but are not current. As discussed in #14, keys that don't show up should be considered revoked (or never valid), but there are situations where retained keys could be useful.

csuwildcat:
> They address the use case of retaining keys that should still be considered valid for actions in the past, but are not current. As discussed in #14, keys that don't show up should be considered revoked (or never valid), but there are situations where retained keys could be useful.

I sympathize with this desire, but why can't we serve this use case in a far more efficient, less spec encumbering manner via something like a cryptographically self-secured URI that listed/detailed any such keys and their past uses? (e.g. IPFS)

jandrieu:
By that argument, why not just export all of the keys and include them by reference?

Seems to me the DID Document should include the relevant cryptographic material directly without requiring a separate dereference to a separate storage system.

csuwildcat:
> By that argument, why not just export all of the keys and include them by reference?
> 
> Seems to me the DID Document should include the relevant cryptographic material directly without requiring a separate dereference to a separate storage system.

So DID Methods should be thought of a long-held data storage system? <-- this is specifically what just about any robustly decentralized DID Method implementation that proceeds beyond small batch cottage use will never be able to support (if they are trying to keep their systems as trustless as possible). My prediction is that the net effect of encouraging DID Doc bloat will push people toward things like `did:web`, because they will see it as a feature that they can drop megabytes of data into a DID Doc and do it 'for free'.

[...]

Issue #14, opened 2019-09-20
Standardize the key revocation list 


brentzundel:
@satazor moved from CCG (https://github.com/w3c-ccg/did-spec/issues/96)

msporny:
This is assigned to @csuwildcat ... waiting for him to be added to the group.

dhh1128:
I wonder if keys are the only data in a DID doc where it is desirable to test revocation. Do endpoints have a similar need?

csuwildcat:
I don't feel this should be something specified in the DID spec, as it may differ greatly between methods, with some methods choosing simpler routes, like implying revocation based on key removal.

OR13:
Wonder if we can adopt a revocation cert approach like what has been used historically with PGP for interoperability?

Or attempt to standardize the use of serviceEndpoints for facilitating revocation lists

glcohen:
I think there is value in including a `revokedKeys` section (or adding it as a flag to the existing key), at the least. For example, it may be useful to verify issuance for a revoked key, without the need to re-issue a credential. This section could certainly define a timestamp by which the revoked key may be removed from the did doc altogether.

Additionally, there could be value in recommending a replacement key, or adding a "succeededBy" property that references another key.

It was brought up that some DID methods are sensitive of large DID Documents. In this case, resolving a service that contains revoked keys should be a possibility, as long as the structure of the inactive key is consistent.

This is not to claim that all revoked keys need to be marked as revoked, certainly there could be a use case where removing a key all-together would make sense.

csuwildcat:
I feel that no matter what we should not force DID Methods to retain revoked keys, and allow them to imply revocation via removal of the key. For many of the more decentralized DID Methods, it simply isn't feasible to retain all that data on a constantly accumulating basis across tens of billions of DIDs.

One scalable way to do this would be to publish key revocation references in an off-chain Hub/datastore that has a well-known mechanism for location/retrieval. This would allow the DID controller to publish any number of key revocation entries with any amount of detail they wish, all without encumbering the scarce resources of the underlying decentralized systems the DIDs are anchored in.

csuwildcat:
To add a bit of detail: if we add a standard property/mechanism like `revokedKeys` that would potentially accrue a lot of data over time, or some key marking and retention convention that would do the same, I know some DID Methods could not feasibly support it at scale (and I imagine others will discover the scalability issues as well).

[...]

PR #224, opened 2020-03-15
A DID Document MUST NOT contain revoked keys.

OR13:
Revised based on feedback... DID Documents MUST NOT contain revoked keys.

Attempts to address:
- https://github.com/w3c/did-core/issues/14
- https://github.com/w3c/did-core/issues/168

***
<a href="https://pr-preview.s3.amazonaws.com/OR13/did-core/pull/224.html" title="Last updated on Mar 21, 2020, 7:08 PM UTC (e004e2b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/224/93bc34c...OR13:e004e2b.html" title="Last updated on Mar 21, 2020, 7:08 PM UTC (e004e2b)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1504,6 +1504,47 @@ responsibility of <a>DID resolvers</a> and other clients. For more information,
 see Section <a href="#resolution"></a>.
       </p>
 
+<p>
+Revoked keys MUST NOT be present in the <code>publicKey</code> property of a <a>DID document</a>.<br/>
+Revoked keys MAY be present in the <code>revokedKey</code> property of a <a>DID document</a>.
+</p>
+
+<pre class="example nohighlight" title="Revoked public keys">
+{
+  "@context": ["https://www.w3.org/ns/did/v1", "https://w3id.org/security/v1"],
+  "id": "did:example:123456789abcdefghi",
+  <span class="comment">...</span>
+  "publicKey": [],
+  "revokedKey": [{
+    "id": "did:example:123#_Qq0UL2Fq651Q0Fjd6TvnYE-faHiOpRlPVQcY_-tA4A",
+    "type": "JwsVerificationKey2020",
+    "controller": "did:example:123",
+    "publicKeyJwk": {
+      "crv": "Ed25519",
+      "x": "VCpo2LMLhn6iWku8MKvSLg2ZAoC-nlOyPVQaO3FxVeQ",
+      "kty": "OKP",
+      "kid": "_Qq0UL2Fq651Q0Fjd6TvnYE-faHiOpRlPVQcY_-tA4A"
+    },
+    "revocation": {
+      <span class="comment">...</span>
+    }
+  }, {
+    "id": "did:example:123#4SZ-StXrp5Yd4_4rxHVTCYTHyt4zyPfN1fIuYsm6k3A",
+    "type": "JwsVerificationKey2020",
+    "controller": "did:example:123",
+    "publicKeyJwk": {
+      "crv": "secp256k1",
+      "x": "Z4Y3NNOxv0J6tCgqOBFnHnaZhJF6LdulT7z8A-2D5_8",
+      "y": "i5a2NtJoUKXkLm6q8nOEu9WOkso1Ag6FTUT6k_LMnGk",
+      "kty": "EC",
+      "kid": "4SZ-StXrp5Yd4_4rxHVTCYTHyt4zyPfN1fIuYsm6k3A"
+    },
+    "revocation": "http://example.com/revocations"
+  }],
+  <span class="comment">...</span>
+}
+</pre>
+
     </section>
 
     <section>
--- a/index.html
+++ b/index.html
@@ -1203,8 +1203,7 @@ methods, which can be used to authenticate or authorize interactions with the
 <a>DID subject</a> or associated parties. The information expressed often
 includes globally unambiguous identifiers and public key material, which can be
 used to verify digital signatures. Other information can be expressed, such as
-status information for the key (for example, whether it is suspended or
-revoked), or other attributes that enable one to determine whether it is a
+status information for the key, or other attributes that enable one to determine whether it is a
 hardware-backed cryptographic key.
       </p>
 
@@ -1310,11 +1309,8 @@ Registry. For a registry of key types and formats, see Appendix
 
       <p>
 If a public key does not exist in the <a>DID document</a>, it MUST be assumed
-the key was revoked or is invalid. The <a>DID document</a> MAY contain revoked
-keys. A <a>DID document</a> containing a revoked key MUST also contain or refer
-to the revocation information for the key (for example, a revocation list). Each
-<a>DID method</a> specification is expected to detail how revocation is
-performed and tracked.
+the key was revoked or is invalid. The <a>DID document</a> MUST NOT contain revoked
+keys.
       </p>
 
       <p>
@@ -1503,47 +1499,6 @@ Caching and expiration of the keys in a <a>DID document</a> is entirely the
 responsibility of <a>DID resolvers</a> and other clients. For more information,
 see Section <a href="#resolution"></a>.
       </p>
-
-<p>
-Revoked keys MUST NOT be present in the <code>publicKey</code> property of a <a>DID document</a>.<br/>
-Revoked keys MAY be present in the <code>revokedKey</code> property of a <a>DID document</a>.
-</p>
-
-<pre class="example nohighlight" title="Revoked public keys">
-{
-  "@context": ["https://www.w3.org/ns/did/v1", "https://w3id.org/security/v1"],
-  "id": "did:example:123456789abcdefghi",
-  <span class="comment">...</span>
-  "publicKey": [],
-  "revokedKey": [{
-    "id": "did:example:123#_Qq0UL2Fq651Q0Fjd6TvnYE-faHiOpRlPVQcY_-tA4A",
-    "type": "JwsVerificationKey2020",
-    "controller": "did:example:123",
-    "publicKeyJwk": {
-      "crv": "Ed25519",
-      "x": "VCpo2LMLhn6iWku8MKvSLg2ZAoC-nlOyPVQaO3FxVeQ",
-      "kty": "OKP",
-      "kid": "_Qq0UL2Fq651Q0Fjd6TvnYE-faHiOpRlPVQcY_-tA4A"
-    },
-    "revocation": {
-      <span class="comment">...</span>
-    }
-  }, {
-    "id": "did:example:123#4SZ-StXrp5Yd4_4rxHVTCYTHyt4zyPfN1fIuYsm6k3A",
-    "type": "JwsVerificationKey2020",
-    "controller": "did:example:123",
-    "publicKeyJwk": {
-      "crv": "secp256k1",
-      "x": "Z4Y3NNOxv0J6tCgqOBFnHnaZhJF6LdulT7z8A-2D5_8",
-      "y": "i5a2NtJoUKXkLm6q8nOEu9WOkso1Ag6FTUT6k_LMnGk",
-      "kty": "EC",
-      "kid": "4SZ-StXrp5Yd4_4rxHVTCYTHyt4zyPfN1fIuYsm6k3A"
-    },
-    "revocation": "http://example.com/revocations"
-  }],
-  <span class="comment">...</span>
-}
-</pre>
 
     </section>
 


index.html
```diff
@@ -1504,6 +1504,47 @@ <h2>Public Keys</h2>
...
 
+<p>
+Revoked keys MUST NOT be present in the <code>publicKey</code> property of a <a>DID document</a>.<br/>
+Revoked keys MAY be present in the <code>revokedKey</code> property of a <a>DID document</a>.
+</p>
+
+<pre class="example nohighlight" title="Revoked public keys">
+{
+  "@context": ["https://www.w3.org/ns/did/v1", "https://w3id.org/security/v1"],
+  "id": "did:example:123456789abcdefghi",
```
dhh1128: What happens if a did with "id" = "did:example?123456789abcdefghi" is *also* present in the publicKeys section, with a different value? I believe this would be valid, since current wisdom is to allow a single "id" value to point to more than one value over the lifespan of the DID. But it seems odd.

index.html
```diff
@@ -1504,6 +1504,47 @@ <h2>Public Keys</h2>
...
+    "type": "JwsVerificationKey2020",
+    "controller": "did:example:123",
+    "publicKeyJwk": {
+      "crv": "Ed25519",
+      "x": "VCpo2LMLhn6iWku8MKvSLg2ZAoC-nlOyPVQaO3FxVeQ",
+      "kty": "OKP",
+      "kid": "_Qq0UL2Fq651Q0Fjd6TvnYE-faHiOpRlPVQcY_-tA4A"
+    },
+    "revocation": {
+      <span class="comment">...</span>
```
dhh1128: It's not clear to me what goes in side the "revocation" tag. Can we expand the example a little bit more than just "..."?

OR13: its not clear to me either :) 

The spec language todays, says it can be embedded or a remote revocation.... But I have never seen an example of either.

I could include GPG Revocation Certificate and an example using https://gpg.jsld.org/

See also: https://debian-administration.org/article/450/Generating_a_revocation_certificate_with_gpg

But as the author of that signature suite, I'm pretty sure nobody other than me is using it right now :)

If there is no better example, I can probably update the GPG Signature suite to provide a revocation certificate example in both embedded and remote form.

index.html
OR13: You mean a key with `id = did:example:123#_Qq0UL2Fq651Q0Fjd6TvnYE-faHiOpRlPVQcY_-tA4A` ? `did:example:123456789abcdefghi` is the DID of the example document... While its possible to use it as an identifier, I belive that should not be allowed, and that it is a seperate issue... 

I think you are hinting at 2 potentially problematic things...

1. What if `did:example:123#_Qq0UL2Fq651Q0Fjd6TvnYE-faHiOpRlPVQcY_-tA4A` is in `authentication`, but also in `revokedKey` with values defined there... doesn't that mean that processing of authentication just became more complicated?

2. What if `did:example:123#_Qq0UL2Fq651Q0Fjd6TvnYE-faHiOpRlPVQcY_-tA4A` is in `revokedKeys`, but is also in `publicKey`... because it was replaced with a new key, when it was revoked, but the new key has the same identifier...

I think that case 2 is illegal, because a did document should never have duplicate identifiers... in other words, if you are rotating  something like `did:example:123#primary`, you cannot retain a copy in `revokedKey`...

Potential solution would be to allow "revoked keys" to be in any section, and add language stating that if any key has a "revocation" property, it is considered revoked... The reason I didn't suggest this is that currently absence of a key indicated revocation, and I wanted to keep that property... not add an or presence of key with "revocation" bit to it....

dhh1128: Yes, you've correctly articulated my two concerns. (Good job; I was struggling to put it into words.)

I strongly prefer to treat absence of a key as equal to revocation, and not having a "revoked keys" section. Maybe the standard has to allow a "revoked keys" section for some use cases, but I hope I never run into one. So I think we're aligned there.

If you are correct that case 2 is illegal (I hope you are), then this is a pretty concrete reason why the asserted behavior goal of allowing #xyz to mean multiple things over its lifetime is a bad idea (at least, if it refers to a key). Maybe we'll get people to reconsider immutability of IDs in DID docs.

Anyway, I don't have any pushback for you; I'm just pointing out subtleties that could trip us up.

OR13: My hope was that by disallowing revoked keys to be present in any section but "revokedKeys" we would at least keep them entirely located there.. and for cases where a DID Method chose to support the concept of revoked keys, say for `assertionMethod`, they would implement their verification logic to handle the  lookup for `did:example:123#vc-key` to look first in `publicKey` and second in `revokedKey`, and for everyone else who didn't use revoked keys, nothing would change.

@msporny @dlongley how does this sound to you?

msporny: -1 to revokedKey (for the reasons stated above). No one is implementing this now, AFAICT.

index.html
```diff
@@ -1504,6 +1504,47 @@ <h2>Public Keys</h2>
 see Section <a href="#resolution"></a>.
       </p>
 
+<p>
+Revoked keys MUST NOT be present in the <code>publicKey</code> property of a <a>DID document</a>.<br/>
+Revoked keys MAY be present in the <code>revokedKey</code> property of a <a>DID document</a>.
```
msporny: -1, this is the wrong thing to do. I think `publicKey` was a mistake and was ok agreeing to it as a compromise. Doubling down on a bad decision and adding `revokedKey` is a realization of exactly what I thought was going to happen if we went down the `publicKey` slippery slope.

There are more than just keys in DID Documents... `publicKeys` are just a special case bucket for `verificationMethod`s. We added it a long time ago because people wanted a single bucket to put keys into that they could refer to from elsewhere in the document. What we really should have done if we wanted a bucket (which was a tenuous compromise to start) is specify a bucket called `verificationMethod`. 

So, -1 to revokedKey because we don't have a strong driving use case for the property.

OR13: I agree with your intention, but I chose this specifically so that people would not need to process `publicKey`, or other verificationMethods differently to understand if things were revoked.

I don't want revokedKeys to be in a did document period... @jandrieu I think you wanted them in the did document.

If we are going to have them in a did document... I want them to be ONLY in one place... like `revokedKey`.

OR13:
Seems like we may want to just remove the concept of revocation, and then try and layer on whatever else is needed...

I think I misinterpreted  https://github.com/w3c/did-core/issues/14#issuecomment-587565105

Seems like we don't want revoked keys... in the did document... at all.

index.html
OR13: As I have revised this content, I request to resolve this thread to avoid confusion.

index.html
OR13: As I have revised this content, I request to resolve this thread to avoid confustion.

index.html
OR13: As I have revised this content, I request to resolve this thread to avoid confusion.

OR13:
@dhh1128 @msporny @jandrieu @csuwildcat @selfissued 

Based on feedback, I have revised this PR to attempt to make it clear that a DID Document MUST NOT contain revoked keys.

Please review.

index.html
```diff
@@ -1310,11 +1309,8 @@ <h2>Public Keys</h2>
 
       <p>
 If a public key does not exist in the <a>DID document</a>, it MUST be assumed
-the key was revoked or is invalid. The <a>DID document</a> MAY contain revoked
-keys. A <a>DID document</a> containing a revoked key MUST also contain or refer
-to the revocation information for the key (for example, a revocation list). Each
-<a>DID method</a> specification is expected to detail how revocation is
-performed and tracked.
+the key was revoked or is invalid. The <a>DID document</a> MUST NOT contain revoked
```
jandrieu: There is not consensus on this second sentence that a DID Document MUST NOT contain revoked keys. There are good use cases for retaining revoked keys if the Controller wishes to communicate that certain keys were once valid but are no longer. The first sentence is fine, IMO. If it isn't in the document, it was revoked or never valid. The second one is overstating the consensus.

OR13: The objective of this PR is to make the concept of revocation clearer... can you point me to any did method that uses revokes keys in a did document today? I suggest we start by agreeing that a did document will not contain revoked keys, and then if someone  (you?) wants to create a PR to define revoked keys, we address that PR separately...  I initially tried to do that here, and then updated this PR based on feedback from  @msporny @csuwildcat  and others....

jandrieu: I can't point you to such methods, but I can point you to the comment thread that seemed to achieve consensus on this matter: https://github.com/w3c/did-core/issues/14#issuecomment-587565105 

The use case that is driving this comes from @ken-ebert: there is a desire to be able to evaluate older credentials based on whether or not a key was valid when the credential was created, even through that key is no longer appropriate for new credentials. However, typing that out, I would question whether or not you can know when a credential was created in order to apply the previously legitimate (and not known to be compromised) key. @ChristopherA has also expressed a desire to be able to state unequivocally that a past key is known to be compromised. This is the inverse of Ken's request, but has a similar construct: in some cases it may be appropriate to communicate some information about specific keys that have been revoked for specific reasons.

That said, my comment is less about the point of value of such annotations than about what appeared to have consensus previously. The consensus was NOT to restrict DID Documents from containing revoked keys, but rather to establish that a key that is not in the DID Document should be considered either revoked or never valid, leaving open the option for others to propose language that would clarify exactly how such a revoked key might be represented.

OR13: I would welcome a PR from @ken-ebert or @ChristopherA defining revocation, but in its absence I think its a terrible idea to allow people to think that a DID Document can contain revoked keys... we should be crystal clear on matters that have such drastic security implications.

jandrieu: I understand your position. But that was not the consensus for which a PR was requested.

Please restrict the PR to the consensus. If you would like to add your own opinion, it would be welcome in another issue where it can be debated appropriately.

OR13: @jandrieu I believe you are misstating "consensus" in this case (as it applies to others, I concede that we clearly disagree :)

Of course, if others object to this PR text, I will concede that I have misinterpreted their preferences as well :)

The goal of this PR is to clarify the concept of "revoked keys"... removing the "MUST NOT contain revoked keys" does not achieve that goal.

Your request in essence, is "leaving the door open to interpretation" for one of the most critical aspects of PKI... with serious security implications.

It is my position that we (except for you) have reached consensus on this matter, and that you, or others who wish to define support for something that everyone else (@msporny @selfissued @csuwildcat @dlongley @msporny , etc...) does not want, should open a PR to do so.

If we are not going to explicitly forbid the inclusion of revoked keys, we would need to define them, and since you have already agreed that nobody is using them... how can we do that?

ChristopherA:
My understanding is that you should never put a revoked key into any of the active key sections (i.e. tag a key as revoked) but that you could put a revoked key into a method specific separate revoked key section with its own annotations. We want this in BTCR, as it is the only way for us to note that a key is not only revoked (i.e. doesn't exist in the key notations of the DID Document), but because it is in the revoked section is explicitly revoked, and has a tag with the reasons, such as compromised.

OR13:
@ChristopherA Thats exactly what this PR attempted to do and @msporny objected. We clearly don't have consensus.

index.html
jandrieu: Regarding consensus, I'm just quoting Manu's assertion that there seemed to be some consensus, so "let's write it up".

I don't have a problem if, in fact, Manu was over-eager on calling for consensus. You and I are apparently not in consensus, so we need to find a way toward that.

You say:
> If we are not going to explicitly forbid the inclusion of revoked keys, we would need to define them

Umm... no. We don't. The language of revoked keys is already there. In fact, your addition of a normative requirement that revoked keys MUST NOT be in the DID Document would require more definition than staying silent on the matter. The fewer constraints the less you are required to define.

My points are twofold. 

First, there was perceived consensus around my own comment on issue 14, cited above. This PR doesn't represent that consensus. That's fine, it seems like there was no consensus so we shouldn't have jumped to a PR yet. Which also means this PR isn't ready.

Second, forbidding revoked keys anywhere is unnecessary overkill at this point. There are identified use cases that want some way to put "old" keys in the DID Document, but there are, as yet, no PRs proposing exactly how that would be done. Those PRs are going to be necessary to enable some form of this capability and it is appropriate to wait until those are written to debate it further so we have something concrete to discuss. However, if you restrict revoked keys now, then you are falsely asserting a consensus that doesn't exist, which will make it harder for those PRs to have an honest hearing.

Both of these options currently lack consensus.

OR13:
Clear lack of consensus closing.

index.html
jonnycrunch: `@ChristopherA has also expressed a desire to be able to state unequivocally that a past key is known to be compromised`

I'm not keen on explicitly stating that a key has been compromised, but I would augment that this PR should say: 

```
If a public key does not exist in the most recent version of the DID document, 
it MUST be assumed  the key was revoked or is invalid. 
```

 There is a need to query older versions of the DID document for validation for a time period that the VC was signed.

OR13: These comments should be made on the issue, so everyone can see them.

PR #238, opened 2020-03-24
Remove JSON-LD Contexts from DID Core

OR13:
They were moved to https://github.com/w3c/did-core-registries/tree/master/contexts

We should ensure that whatever redirected are needed to be setup have been done so before merging this.


iherman:
`package.json` is not used by `respec`, ie, there is no problem there. 

I actually do not know why package.json is there for the moment. @msporny should tell us...

OR13:
It appears that package.json was included here, because did-context (only used by JSON-LD) was being built as an npm package from it... if thats still desirable we can do the same in the registries repo...although I would prefer to package the json schemas as well, and call it something like `did-core-registries-definitions`

msporny:
> It appears that package.json was included here, because did-context (only used by JSON-LD) was being built as an npm package from it... 

Yes, this is correct.

> if thats still desirable we can do the same in the registries repo...although I would prefer to package the json schemas as well, and call it something like did-core-registries-definitions

+1 for still providing an npm package... -1 for bundling the context w/ the schemas. They should be two separate npm packages.

kdenhartog:
Thanks for the clarification. I'm good to see this move forward.

OR13:
@msporny @iherman This PR is blocked, until you can confirm that the redirects are now pointing towards the registries.

iherman:
> @msporny @iherman This PR is blocked, until you can confirm that the redirects are now pointing towards the registries.

is there a resolution to this?

More generally, there has been a number of different comments on a set of redirection that must be done but I do not think there is a list somewhere that has absolute and 100% consensus on what series of redirection are to be done. I would really need that before I do anything, otherwise you can be sure I will screw up something...

OR13:
@iherman AFAIK, the only preventing this merge:

https://www.w3.org/ns/did/v1 -> https://w3c.github.io/did-core-registries/contexts/did-v1.jsonld

everything else can be handled separately.

@msporny can you confirm and approve or request changes and explain?

msporny:
> @msporny can you confirm and approve or request changes and explain?

Confirmed, @iherman, the only redirect that we need to fix is the one @OR13 mentioned above. Once that is done, we can merge this PR.

iherman:
The redirection


https://www.w3.org/ns/did/v1 -> https://w3c.github.io/did-core-registries/contexts/did-v1.jsonld


should be done now, although I am not sure how you can check that (or are the two targets different in content?). Please check as you can.

> Also note that the did-core-registries repo might be changing to did-registries at some undetermined time in the future. :)

Github is very good at redirecting github.com URL-s after a renaming but, afaik, that is not the case for github.io URL-s. In other words, we will have to be careful on synchronizing these steps!

But that is for another day...

msporny:
> should be done now, although I am not sure how you can check that (or are the two targets different in content?). Please check as you can.

Great, thank you @iherman!

Yes, the new content has changes that are detectable from the old content. I've confirmed that the new redirect is working.

msporny:
Normative, multiple positive reviews, sanity checks passed, technical checks passed, merging.

Issue #207, opened 2020-02-21
Add section on extensibility and conformance


jricher:
This document defines how to create and use DIDs and documents that are conformant to a global international standard. In particular, the use of the registry for properties is the only guaranteed method for two independent implementations to be able to work together. However, it's always possible for two specific implementations to agree to use a property or encoding, or for an implementation to use something internally. 

Relates to #205

OR13:
Should this be moved to did core registries?

talltree:
This will be addressed by the Extensibility subsection of Section 6 that I'm writing on the Data Model. I plan to submit the PR this week and will include wording to this effect.

talltree:
Update: please see [PR 241](https://github.com/w3c/did-core/pull/241) for the text of section 6 that addresses this issue. If folks are satisifed, this issue can be closed.

rhiaro:
@jricher does the [additional text about extensibility](https://w3c.github.io/did-core/#extensibility) address this issue?

brentzundel:
If there are no objections raised, we will close this issue in 7 days.

brentzundel:
Marked `pending close` for more than a week with no comment, closing.

Issue #204, opened 2020-02-21
Define terminology for properties and values


jricher:
Since a DID document is an abstract data model that consists of a collection of properties, we need a consistent way to refer to the parts of the property so that parts of the spec (like representations) can use them. For example, given a property like:

```
      <h2>DID Subject</h2>

      <p>
The <a>DID subject</a> is denoted with the <code>id</code> property. The
<a>DID subject</a> is the entity that the <a>DID document</a> is about. That is,
it is the entity identified by the <a>DID</a> and described by the
<a>DID document</a>.
      </p>
```

We need a way to refer to the string `id` as a specific entity. Is this the "property name"? Is this the "property" itself? Is it the "member label" or something else we make up? We also have to talk about things like the "property value" and the "property value data type". These terms all need to be defined and used consistently.

jricher:
In addition, the document needs to define what an "empty" property value is.

kdenhartog:
We discussed this on the call today. I'd be happy to assist in making this PR, but I think we need to bikeshed the terminology a bit. Bringing in the likes of @dmitrizagidulin @msporny @selfissued @dlongley to assist with some color options 😄

rhiaro:
Partly addressed by text in PR #241: https://pr-preview.s3.amazonaws.com/talltree/did-core/pull/241.html#definition

kdenhartog:
@iherman suggested that we not bikeshed this issue and instead use common language already defined. His suggestion was to use [infra](https://infra.spec.whatwg.org/#map-key) definitions which would leave us with "key" and "value". 

See [#241 (comment)](https://github.com/w3c/did-core/pull/241#issuecomment-609369689) for additional context.

Can we please get a rough strawpoll on this on the next call @burnburn @brentzundel and if it's agreed I'll go in and start the PR to update this in the spec?

peacekeeper:
In the PR https://github.com/w3c/did-core/pull/247 about DID Resolution, when talking about metadata, I am using language such as 

> "These sets of metadata are expressed as maps of name-value string pairs, referred to as metadata properties."

I suppose this can perhaps also be improved, depending on the outcome of the discussion in this thread.

burnburn:
> @iherman suggested that we not bikeshed this issue and instead use common language already defined. His suggestion was to use [infra](https://infra.spec.whatwg.org/#map-key) definitions which would leave us with "key" and "value".
> 
[...]

PR #241, opened 2020-03-26
initial text for section 6

talltree:

First draft of the new Section 6 that we agreed to add to the spec after the Amsterdam F2F meeting.

***
<a href="https://pr-preview.s3.amazonaws.com/talltree/did-core/pull/241.html" title="Last updated on Apr 21, 2020, 8:30 PM UTC (dba5548)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/241/fd4c5cf...talltree:dba5548.html" title="Last updated on Apr 21, 2020, 8:30 PM UTC (dba5548)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1227,7 +1227,7 @@ representations.
         </li>
       <li>
 Representations MAY define other extensibility mechanisms including methods for
-decentralized extensions. Such extension mechanisms SHOULD support lossless
+decentralized extensions. Such extension mechanisms MUST support lossless
 conversion into any other conformant representation.
       </li>
       </ol>
--- a/index.html
+++ b/index.html
@@ -1198,7 +1198,7 @@ anything you can represent in the <a>DID document</a> data model can be represen
 compliant representation.
         </li>
         <li>
-The representation MUST have an unambiguous MIME type.
+The representation MUST be associated with an IANA-registered MIME type.
         </li>
         <li>
 The representation MUST define fragment processing rules for its MIME type that
--- a/index.html
+++ b/index.html
@@ -1194,15 +1194,18 @@ Following are the requirements for conformant representations.
       <ol>
         <li>
 A representation MUST define an unambiguous encoding and decoding of all
-properties defined in this specification such that the value of any property
-may be losslessly converted into another representation.
+properties defined in this specification. This definition MUST enable lossless
+conversion of the value of any property into another representation. This means
+anything you can represent in the DID document model can be represented in any
+compliant representation.
         </li>
         <li>
 The representation MUST have an unambiguous MIME type.
         </li>
         <li>
 The representation MUST define fragment processing rules for its MIME type that
-are conformant with this specification.
+are conformant with the fragment processing rules defined in section
+<a href="#fragment"></a> of this specification.
         </li>
       </ol>
     </section>
@@ -1214,10 +1217,11 @@ The data model supports two types of extensibility.
       <ol>
         <li>
 Registry-based extensions are RECOMMENDED for maximum interoperability. The use
-of the registry for properties is the only guaranteed method for two independent
-implementations to be able to work together. Registry-based extensions MUST be
-losslessly convertible between all conformant representations. The official W3C
-extension registry is defined by the DID Core Registry specification.
+of the registry for properties is the only specified method for two independent
+implementations to be able to work together. Registry-based extensions MUST
+enable lossless conversion between all conformant representations. The official
+W3C extension registry is defined by the DID Core Registry specification
+[[DID-CORE-REGISTRIES]].
         </li>
       <li>
 Decentralized extensibility mechanisms MAY be defined by the specification for a
--- a/index.html
+++ b/index.html
@@ -1180,23 +1180,21 @@ representations, and a set of requirements for extensibility.
     <section>
       <h2>Definition</h2>
       <p>
-A <a>DID document</a> consists of a collection of name-value pairs which will be
-referred to as properties in this document. The definitions of each of these
-properties are specified in section <a href="#core-properties"></a>. Specific
+A <a>DID document</a> consists of a collection of property-value pairs. The definitions 
+of each of these properties are specified in section <a href="#core-properties"></a>. Specific
 representations are defined in section <a href="#core-representations"></a>.
       </p>
     </section>
     <section>
       <h2>Representations</h2>
       <p>
-Following are the requirements for conformant representations.
+Following are the requirements for representations.
       </p>
       <ol>
         <li>
 A representation MUST define an unambiguous encoding and decoding of all
-properties defined in this specification. This definition MUST enable lossless
-conversion of the value of any property into another representation. This means
-anything you can represent in the DID document model can be represented in any
+properties and their associated values as defined in this specification. This means
+anything you can represent in the <a>DID document</a> data model can be represented in any
 compliant representation.
         </li>
         <li>
--- a/index.html
+++ b/index.html
@@ -1220,10 +1220,8 @@ The data model supports two types of extensibility.
 For maximum interoperability, it is RECOMMENDED that extensions use the official
 W3C DID Core Registries mechanism [[DID-CORE-REGISTRIES]]. 
 The use of this mechanism for new properties or other
-extensions is the only specified method that ensures that two different representations will
-be able to work together. <a>Producers</a> and <a>consumers</a> of representations MUST process all entries in the DID Core Registries according to the rules associated with the entry in the registry.
-Registry-based extensions MUST enable lossless conversion between all conformant
-representations.
+extensions is the only specified method that ensures that two different
+representations will be able to work together.
         </li>
       <li>
 Representations MAY define other extensibility mechanisms including methods for
--- a/index.html
+++ b/index.html
@@ -1220,7 +1220,7 @@ The data model supports two types of extensibility.
 For maximum interoperability, it is RECOMMENDED that extensions use the official
 W3C DID Core Registries mechanism [[DID-CORE-REGISTRIES]]. 
 The use of this mechanism for new properties or other
-extensions is the only specified method for two independent implementations to
+extensions is the only specified method that ensures that two different representations will
 be able to work together. Representations MUST process any extension in the DID
 Core and Extensions Registries according to the definition in the registry.
 Registry-based extensions MUST enable lossless conversion between all conformant
--- a/index.html
+++ b/index.html
@@ -1221,8 +1221,7 @@ For maximum interoperability, it is RECOMMENDED that extensions use the official
 W3C DID Core Registries mechanism [[DID-CORE-REGISTRIES]]. 
 The use of this mechanism for new properties or other
 extensions is the only specified method that ensures that two different representations will
-be able to work together. Representations MUST process any extension in the DID
-Core and Extensions Registries according to the definition in the registry.
+be able to work together. <a>Producers</a> and <a>consumers</a> of representations MUST process all entries in the DID Core Registries according to the rules associated with the entry in the registry.
 Registry-based extensions MUST enable lossless conversion between all conformant
 representations.
         </li>
--- a/index.html
+++ b/index.html
@@ -1233,9 +1233,9 @@ conversion into any other conformant representation.
       </ol>
       <p class="note">
 It is always possible for two specific implementations to agree out-of-band to
-use a shared extension property or a new representation that is not recorded in
-the [[DID-CORE-REGISTRIES]]. However interoperability between such
-implementations may be less reliable.
+use a mutually understood extension or representation that is not recorded in
+the DID Core Registries [[DID-CORE-REGISTRIES]]; interoperability between such
+implementations and the larger ecosystem will be less reliable.
       </p>
     </section>
 
--- a/index.html
+++ b/index.html
@@ -1169,19 +1169,66 @@ absolute DID URL value of <code>did:example:123456789abcdefghi#key-1</code>.
 
   </section>
 
-  <section class="informative">
+  <section>
     <h1>Data Model</h1>
-
+    <p>
+This specification defines an abstract data model for <a>DID documents</a> so it
+is independent of any specific representation. In this section we provide a
+high-level description of the data model, a set of requirements for
+representations, and a set of requirements for extensibility.
+    </p>
     <section>
       <h2>Definition</h2>
+      <p>
+A <a>DID document</a> consists of a collection of name-value pairs which will be
+referred to as properties in this document. The definitions of each of these
+properties are specified in section <a href="#core-properties"></a>. Specific
+representations are defined in section <a href="#core-representations"></a>.
+      </p>
     </section>
-
+    <section>
+      <h2>Representations</h2>
+      <p>
+Following are the requirements for conformant representations.
+      </p>
+      <ol>
+        <li>
+A representation MUST define an unambiguous encoding and decoding of all
+properties defined in this specification such that the value of any property
+may be losslessly converted into another representation.
+        </li>
+        <li>
+The representation MUST have an unambiguous MIME type.
+        </li>
+        <li>
+The representation MUST define fragment processing rules for its MIME type that
+are conformant with this specification.
+        </li>
+      </ol>
+    </section>
     <section>
       <h2>Extensibility</h2>
-    </section>
-
-    <section>
-      <h2>Representation Requirements</h2>
+      <p>
+The data model supports two types of extensibility.
+      </p>
+      <ol>
+        <li>
+Registry-based extensions are RECOMMENDED for maximum interoperability. The use
+of the registry for properties is the only guaranteed method for two independent
+implementations to be able to work together. Registry-based extensions MUST be
+losslessly convertible between all conformant representations. The official W3C
+extension registry is defined by the DID Core Registry specification.
+        </li>
+      <li>
+Decentralized extensibility mechanisms MAY be defined by the specification for a
+conformant representation. Decentralized extension mechanisms SHOULD be defined
+such that any conformant extension may be losslessly converted into any other
+conformant representation. However, it's always possible for two specific
+implementations to agree to use an extension property or a new representation,
+or for an implementation to use its own properties or representations
+internally.
+      </li>
+      </ol>
     </section>
 
   </section>
--- a/index.html
+++ b/index.html
@@ -1218,8 +1218,8 @@ The data model supports two types of extensibility.
       <ol>
         <li>
 For maximum interoperability, it is RECOMMENDED that extensions use the official
-W3C registries defined by the DID Core and Extensions Registries specification
-[[DID-CORE-REGISTRIES]]. The use of this registry for new properties or other
+W3C DID Core Registries mechanism [[DID-CORE-REGISTRIES]]. 
+The use of this mechanism for new properties or other
 extensions is the only specified method for two independent implementations to
 be able to work together. Representations MUST process any extension in the DID
 Core and Extensions Registries according to the definition in the registry.
--- a/index.html
+++ b/index.html
@@ -1219,23 +1219,27 @@ The data model supports two types of extensibility.
       </p>
       <ol>
         <li>
-Registry-based extensions are RECOMMENDED for maximum interoperability. The use
-of the registry for properties is the only specified method for two independent
-implementations to be able to work together. Registry-based extensions MUST
-enable lossless conversion between all conformant representations. The official
-W3C extension registry is defined by the DID Core Registry specification
-[[DID-CORE-REGISTRIES]].
+For maximum interoperability, it is RECOMMENDED that extensions use the official
+W3C registries defined by the DID Core and Extensions Registries specification
+[[DID-CORE-REGISTRIES]]. The use of this registry for new properties or other
+extensions is the only specified method for two independent implementations to
+be able to work together. Representations MUST process any extension in the DID
+Core and Extensions Registries according to the definition in the registry.
+Registry-based extensions MUST enable lossless conversion between all conformant
+representations.
         </li>
       <li>
-Decentralized extensibility mechanisms MAY be defined by the specification for a
-conformant representation. Decentralized extension mechanisms SHOULD be defined
-such that any conformant extension may be losslessly converted into any other
-conformant representation. However, it's always possible for two specific
-implementations to agree to use an extension property or a new representation,
-or for an implementation to use its own properties or representations
-internally.
+Representations MAY define other extensibility mechanisms including methods for
+decentralized extensions. Such extension mechanisms SHOULD support lossless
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -1169,19 +1169,66 @@ <h2>Relative DID URLs</h2>
...
     <section>
-      <h2>Representation Requirements</h2>
+      <h2>Extensibility</h2>
+      <p>
+The data model supports two types of extensibility.
+      </p>
+      <ol>
+        <li>
+Registry-based extensions are RECOMMENDED for maximum interoperability. The use
+of the registry for properties is the only guaranteed method for two independent
```
kdenhartog: guaranteed seems a bit strong here. The implementations could still encounter bugs. 

Maybe something along the lines of "The use of the registries makes getting two independent implementations to work together far easier."

index.html
```diff
@@ -1169,19 +1169,66 @@ <h2>Relative DID URLs</h2>
...
+implementations to be able to work together. Registry-based extensions MUST be
+losslessly convertible between all conformant representations. The official W3C
```
kdenhartog: Should we define what we mean by losslessly. I assume it can be inferred from the standard definition, but figured the questions was worth getting other opinions on this.

index.html
```diff
@@ -1169,19 +1169,66 @@ <h2>Relative DID URLs</h2>
...
+extension registry is defined by the DID Core Registry specification.
```
kdenhartog: Should we link to the DID Core Registry spec here?

index.html
brentzundel: What about "is the only specified method"
Or "is the only official method"?

index.html
brentzundel: Yes, we should

index.html
```diff
@@ -1169,19 +1169,66 @@ <h2>Relative DID URLs</h2>
...
+        </li>
+      <li>
+Decentralized extensibility mechanisms MAY be defined by the specification for a
```
jricher: This, with the SHOULD above, seems to imply that a representation can choose to ignore the contents of the registry entirely. I think this is dangerous ground and will hurt interoperability. I would rather have a requirement that representations MUST detail with registry contents in some predictable way and MAY have other mechanisms for communicating extensions.

index.html
talltree: @kdenhartog Definitely we should, but when I asked Brent how we would do that, he wasn't sure. So I just posted the PR in the hope that one of the ReSpec and/or W3C experts would tell us how to do that.

index.html
talltree: My rationale was that "lossless" is a well-known concept in digital tech. Wikipedia doesn't have a definition of the word by itself but does have https://en.wikipedia.org/wiki/Lossless_compression which starts:

> Lossless compression is a class of data compression algorithms that allows the original data to be perfectly reconstructed from the compressed data.

So my sense is that we don't have to provide a definition, however I'm fine with doing that if others think it would clarify the spec.

index.html
talltree: @kdenhartog I incorporated this language directly from @jricher in #207. @jricher, what do you think?

index.html
kdenhartog: Yeah, I understood what you meant to just didn't want it to get captured as tribal knowledge was the reason I was checking.

index.html
kdenhartog: I like where Brent went with this. Of note, I'm not formally objecting to use of the word guaranteed, just thought it felt overly strong.

index.html
kdenhartog: I believe We add

```json
    "DID-CORE-REGISTRIES": {
      title: "Decentralized Identifiers Core Registries",
      href: "https://w3c.github.io/did-core-registries/",
      authors: [
      	"Orie Steele",
        "Manu Sporny"
      ],
      status: "ED",
      publisher: "Decentralized Identifier Working Group"
    }
```

  to the JSON called localBiblio in common.js and then we use [[DID-CORE-REGISTRIES]] in the index.html

iherman:
The current draft only defines properties, but does not say anything about what the values of those properties might be. On the other hand, later sections do use a terminology like:

> If a DID document includes a publicKey property, the value of the property MUST be an array of public key objects.

Looking at this from the abstract model point of view, the term 'object' and 'array' is not defined. We should also refer to strings as possible values, see if they are utf-8 or not, etc, etc...

I wonder whether it is not a better approach to refer to something like the [WhatWG Infra](https://infra.spec.whatwg.org/) document instead. Although that document was created for the unambiguous and language independent way of specifying algorithms, the sections on [primitive data types](https://infra.spec.whatwg.org/#primitive-data-types) and [data structures](https://infra.spec.whatwg.org/#data-structures) have a bunch of generic definitions that we can reuse without reinventing the wheel... E.g., a DID document is a "map", the value can be a "string", a "map", or an "ordered set", etc, where all the terms in quotes are defined in infra (or something equivalent).

(I am not bound to infra. But I am afraid we would be forced to a wheel reinvention...)

index.html
jricher: We can define "lossless" in the context of our specification. What we mean is what @talltree said -- you can get data in and out of a representation without losing anything from that representation. That means anything you can represent in the DID document model can be represented in any compliant representation, and you can't pull tricks like "this is a Dates In The Julian Calendar Type that only my system understands" without defining a general model and representation for that across different formats.

index.html
rhiaro: +1 "the only specified method", or perhaps "the most reliable method"

index.html
```diff
@@ -1169,19 +1169,66 @@ <h2>Relative DID URLs</h2>
...
-
     <section>
-      <h2>Representation Requirements</h2>
+      <h2>Extensibility</h2>
+      <p>
+The data model supports two types of extensibility.
+      </p>
+      <ol>
+        <li>
+Registry-based extensions are RECOMMENDED for maximum interoperability. The use
```
rhiaro: > Registry-based extensions ... The official W3C extension registry is defined ...

Implies the [potential] existence of other registries. I'm not sure if this is intended, or the right steer for this paragraph. Though it also says "*the* registry" later rather than "*a* registry" so maybe I'm reading too much in. But I think immediately stating that the W3C registry, rather than the general concept of registry-based extensions, is RECOMMENDED for maximum interop might be a better way to begin.

index.html
```diff
@@ -1169,19 +1169,66 @@ <h2>Relative DID URLs</h2>
...
+of the registry for properties is the only guaranteed method for two independent
+implementations to be able to work together. Registry-based extensions MUST be
+losslessly convertible between all conformant representations. The official W3C
+extension registry is defined by the DID Core Registry specification.
+        </li>
+      <li>
+Decentralized extensibility mechanisms MAY be defined by the specification for a
+conformant representation. Decentralized extension mechanisms SHOULD be defined
+such that any conformant extension may be losslessly converted into any other
+conformant representation. However, it's always possible for two specific
```
rhiaro: > However ... representations internally

Doesn't feel like it fits with the rest of the normative language. It is also not necessarily tied to "decentralized extensibility mechanisms" - implementations that use the registry sometimes could also choose to make out-of-band agreements for certain extensions. Perhaps this sentence can be de-emphasised in a NOTE box? With the caveat about breaking interop/being non-conformant, something like:

> [NOTE] It's always possible for two specific implementations to agree out-of-band to use a common extension property or a new representation that is not recorded in the [[DID Core Registries]]. It is worth noting that these extensions are not considered conformant to the DID Core specification, and interoperability between implementations is less reliable.

index.html
```diff
@@ -1169,19 +1169,66 @@ <h2>Relative DID URLs</h2>
...
+        </li>
+        <li>
+The representation MUST have an unambiguous MIME type.
+        </li>
+        <li>
+The representation MUST define fragment processing rules for its MIME type that
+are conformant with this specification.
+        </li>
+      </ol>
     </section>
```
rhiaro:
```suggestion
        <p>
The core representations are specified in section <a href="#core-representations"></a>.
        </p>
    </section>
```

(I know it says this in the prior section too, but reading the Representations subsection alone, a mention of Core Representations is conspicuously absent.)

index.html
```diff
@@ -1169,19 +1169,66 @@ <h2>Relative DID URLs</h2>
 
   </section>
 
-  <section class="informative">
+  <section>
     <h1>Data Model</h1>
-
+    <p>
+This specification defines an abstract data model for <a>DID documents</a> so it
+is independent of any specific representation. In this section we provide a
```
rhiaro:
```suggestion
This specification defines an abstract data model for <a>DID documents</a>,
independent of any specific representation. This section provides a
```

index.html
peacekeeper: I think I was the one who originally introduced the term "lossless" into the conversation. Before the F2F, one common argument was that a DID document could have a JSON "encoding" and a JSON-LD "encoding", but that argument quietly assumed that you would simply remove or add a `@context` (with that oversimplified approach however, the conversion is not guaranteed to be "lossless"). This is also why I preferred the term "representation" over "encoding".

index.html
peacekeeper: +1 to "the only specified method" or even weaker such as "one specified method". Extensions may be interoperable between different representations through mechanisms that don't require the Registry at all.

index.html
```diff
@@ -1169,19 +1169,66 @@ <h2>Relative DID URLs</h2>
...
+        <li>
+A representation MUST define an unambiguous encoding and decoding of all
+properties defined in this specification such that the value of any property
+may be losslessly converted into another representation.
+        </li>
+        <li>
+The representation MUST have an unambiguous MIME type.
+        </li>
+        <li>
+The representation MUST define fragment processing rules for its MIME type that
```
peacekeeper, line 1204: There's an issue https://github.com/w3c/did-core/pull/241 by @tplooker that asks if there is a guarantee that fragments will consistently identify the same parts of a DID document across different representations, or if each representation is free to use fragments in different ways (e.g. the spec currently mentions JSON Pointer, but that will probably not work in non-JSON-based representations).

index.html
talltree: Made this change in commit [77a9787](https://github.com/w3c/did-core/pull/241/commits/77a97872959692d262c16bbd827f857677ba8f82) to fix this wording.

index.html
talltree: Made this change in commit [77a9787](https://github.com/w3c/did-core/pull/241/commits/77a97872959692d262c16bbd827f857677ba8f82) to add a definition of "lossless" plagiarized from @jricher's comment.

index.html
talltree: Made this change in commit [77a9787](https://github.com/w3c/did-core/pull/241/commits/77a97872959692d262c16bbd827f857677ba8f82)—thank you @kdenhartog!

index.html
talltree: Made this change in commit [77a9787](https://github.com/w3c/did-core/pull/241/commits/77a97872959692d262c16bbd827f857677ba8f82) by specifying that every representation must follow the fragment processing rules defined in section 5.2.6.

index.html
```diff
@@ -1169,19 +1169,66 @@ <h2>Relative DID URLs</h2>
...
+are conformant with this specification.
+        </li>
+      </ol>
     </section>
-
     <section>
-      <h2>Representation Requirements</h2>
+      <h2>Extensibility</h2>
+      <p>
+The data model supports two types of extensibility.
```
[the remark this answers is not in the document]
talltree: Thanks @rhiaro. Made this change in commit [77a9787](https://github.com/w3c/did-core/pull/241/commits/77a97872959692d262c16bbd827f857677ba8f82).

index.html
talltree: Made this change in commit [fdacc41](https://github.com/w3c/did-core/pull/241/commits/fdacc41e349e2a25352fc9ebb6179ee5ea973db1). I revised the language as you suggested.

index.html
talltree: Made this change in commit [fdacc41](https://github.com/w3c/did-core/pull/241/commits/fdacc41e349e2a25352fc9ebb6179ee5ea973db1). You are quite right; I turned that language into a Note.

talltree:
> The current draft only defines properties, but does not say anything about what the values of those properties might be.

@iherman Responding to this [comment](https://github.com/w3c/did-core/pull/241#issuecomment-605031098), I think you make a very good point. As an editor, it would be ideal to reference a precise set of data type definitions like that. Do you want to raise this as a separate issue?

iherman:
>> The current draft only defines properties, but does not say anything about what the values of those properties might be.
>
> @iherman Responding to this comment, I think you make a very good point. As an editor, it would be ideal to reference a precise set of data type definitions like that. Do you want to raise this as a separate issue?

@talltree I think it is indeed a good approach to get the other PR comments settled, merge this PR to put a stake in the ground as far as the overall structure of the document is concerned, and then to raise this is a separate issue. So yes, I will re-raise this once the PR is merged.

index.html
talltree: I agree, @jricher. I revised the wording in commit [fdacc41].(https://github.com/w3c/did-core/pull/241/commits/fdacc41e349e2a25352fc9ebb6179ee5ea973db1)

kdenhartog:
> 
> 
> > > The current draft only defines properties, but does not say anything about what the values of those properties might be.
> > 
> > 
> > @iherman Responding to this comment, I think you make a very good point. As an editor, it would be ideal to reference a precise set of data type definitions like that. Do you want to raise this as a separate issue?
> 
> @talltree I think it is indeed a good approach to get the other PR comments settled, merge this PR to put a stake in the ground as far as the overall structure of the document is concerned, and then to raise this is a separate issue. So yes, I will re-raise this once the PR is merged.

Of note, we've got a issue to decide the names of these name-value pairs going in issue #204

Please pop over there and weigh in on it, and once we decide on one I'm planning to make a pass to update the terminology used.

iherman:
@kdenhartog yes, you are right, #204 is a similar issue. (Sorry not to have realized it, I somehow missed it. Thanks for reminding me.) I believe what I said in https://github.com/w3c/did-core/pull/241#issuecomment-605031098 on [infra](https://infra.spec.whatwg.org/) (or anything similar) would cover that issue, too, insofar on tightening our own terminology.

We can discuss on choosing another, similar formalism but we should avoid a bikeshedding type discussion on what we would use. Personally, I am not a huge fan of the [infra](https://infra.spec.whatwg.org/) style, but that is a matter of taste. _Infra has the huge merit to exist_, to be maintained, and to be used in other W3C documents already. (Not only on strictly speaking "browser topics", see the algorithm section of the [Publication manifest](https://www.w3.org/TR/pub-manifest/#processing-algorithm) as an example.) .

index.html
kdenhartog: Thanks looks good.

index.html
```diff
@@ -1169,19 +1169,77 @@ <h2>Relative DID URLs</h2>
...
-      <h2>Extensibility</h2>
+      <h2>Representations</h2>
+      <p>
+Following are the requirements for conformant representations.
+      </p>
+      <ol>
+        <li>
+A representation MUST define an unambiguous encoding and decoding of all
+properties defined in this specification. This definition MUST enable lossless
+conversion of the value of any property into another representation. This means
```
msporny, line 1198:
```suggestion
A representation MUST define an unambiguous encoding and decoding of all
properties and their associated values as defined in this specification. This means
```

This language is a bit ambiguous... for example, let's say one format supports 64-bit integers and the other is only capable of supporting 32-bit integers. Does that mean that any format that can only support 32-bit integers is not allowed?

It feels like the first sentence is fine, and the second sentence is not necessary... because if you can unambiguously encode/decode all properties in the spec to/from the data model to the representation and back, then you should be able to go to any other representation w/o issue. Perhaps adding "unambiguously and losslessly" or just "losslessly" is all that's needed in the first sentence?

index.html
```diff
@@ -1169,19 +1169,77 @@ <h2>Relative DID URLs</h2>
...
+anything you can represent in the DID document model can be represented in any
+compliant representation.
+        </li>
+        <li>
+The representation MUST have an unambiguous MIME type.
```
msporny, line 1203:
```suggestion
The representation MUST be associated with a MIME type.
```

Not sure what an "unambiguous" MIME type is... does that mean something like "text/plain" vs "text/html"... where text/plain is more ambiguous than text/html?

index.html
```diff
@@ -1169,19 +1169,77 @@ <h2>Relative DID URLs</h2>
...
-      <h2>Representation Requirements</h2>
+      <h2>Extensibility</h2>
+      <p>
+The data model supports two types of extensibility.
+      </p>
+      <ol>
+        <li>
+For maximum interoperability, it is RECOMMENDED that extensions use the official
+W3C registries defined by the DID Core and Extensions Registries specification
+[[DID-CORE-REGISTRIES]]. The use of this registry for new properties or other
```
msporny, line 1224:
```suggestion
W3C DID Core Registries mechanism [[DID-CORE-REGISTRIES]]. 
The use of this mechanism for new properties or other
```

index.html
```diff
@@ -1169,19 +1169,77 @@ <h2>Relative DID URLs</h2>
...
+extensions is the only specified method for two independent implementations to
```
msporny, line 1225:
```suggestion
extensions is the only specified method that ensures that two different representations will
```

msporny: Two independent JSON-LD implementations will work just fine w/o using the registry. The registry is there so that JSON-only folks will continue to interoperate with the JSON-LD folks (as well as other JSON-only folks). It has more to do with representations that it does implementations.

index.html
```diff
@@ -1169,19 +1169,77 @@ <h2>Relative DID URLs</h2>
...
+be able to work together. Representations MUST process any extension in the DID
+Core and Extensions Registries according to the definition in the registry.
```
msporny, line 1227:
```suggestion
be able to work together. <a>Producers</a> and <a>consumers</a> of representations MUST process all entries in the DID Core Registries according to the rules associated with the entry in the registry.
```

Representations can't "process" anything as they're just a representation... but perhaps we can talk about "producers" and "consumers" of representations. Keep in mind that this text probably belongs down in the section of producers and consumers and not here.

index.html
```diff
@@ -1169,19 +1169,77 @@ <h2>Relative DID URLs</h2>
...
+Registry-based extensions MUST enable lossless conversion between all conformant
```
msporny, line 1228: Again, this wording is very difficult to test and analyze for correctness. I get what you're going for here... don't have alternative language, unfortunately. Perhaps just strike the sentence and merge this into the statement about losslessly transforming data model to/from representation?

index.html
```diff
@@ -1169,19 +1169,77 @@ <h2>Relative DID URLs</h2>
...
+representations.
+        </li>
+      <li>
+Representations MAY define other extensibility mechanisms including methods for
+decentralized extensions. Such extension mechanisms SHOULD support lossless
+conversion into any other conformant representation.
```
msporny, line 1234:
```suggestion
decentralized extensions. Such extension mechanisms MUST support lossless
conversion into any other conformant representation.
```

It's MUST in the paragraphs above... downgrading this to a SHOULD here enables the introduction of values that can't be represented in all representations, harming interop. We should be consistent... either MUST in both places, or SHOULD in both places. I'd prefer MUST in both places.

index.html
```diff
@@ -1169,19 +1169,77 @@ <h2>Relative DID URLs</h2>
...
+      </li>
+      </ol>
+      <p class="note">
+It is always possible for two specific implementations to agree out-of-band to
+use a shared extension property or a new representation that is not recorded in
+the [[DID-CORE-REGISTRIES]]. However interoperability between such
+implementations may be less reliable.
```
msporny, line 1241:
```suggestion
It is always possible for two specific implementations to agree out-of-band to
use a mutually understood extension or representation that is not recorded in
the DID Core Registries [[DID-CORE-REGISTRIES]]; interoperability between such
implementations and the larger ecosystem will be less reliable.
```

Removing MAY language so it's not confused to be a normative statement.

index.html
```diff
@@ -1169,19 +1169,77 @@ <h2>Relative DID URLs</h2>
...
+independent of any specific representation. This section provides a
+high-level description of the data model, a set of requirements for
+representations, and a set of requirements for extensibility.
+    </p>
     <section>
       <h2>Definition</h2>
+      <p>
+A <a>DID document</a> consists of a collection of name-value pairs which will be
+referred to as properties in this document. The definitions of each of these
+properties are specified in section <a href="#core-properties"></a>. Specific
```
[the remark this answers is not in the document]
burnburn: @msporny this comment should probably (also) go in issue #204.  I would rather have the bikeshedding there than on a call approving/rejecting this PR.

index.html
burnburn: We cannot make normative statements about what representation definitions MUST contain, because those are constraints on a spec rather than implementations (and thus untestable).  I faced the same challenge in the WebRTC work regarding extensibility of the spec and ultimately wrote the language you see [here](https://www.w3.org/TR/mediacapture-streams/#extensibility).  It might be helpful.

index.html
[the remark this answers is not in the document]
msporny: Done.

index.html
msporny: Yes, let's please do what @burnburn did for WebRTC. That would make me feel much better about the large number of non-testable normative statements we have in the spec.

index.html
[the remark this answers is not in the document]
kdenhartog: Included this option in the poll over on #204 [(comment)](https://github.com/w3c/did-core/issues/204#issuecomment-612574398)

index.html
kdenhartog: Should we say the MIME type should be IANA registered or is that overly constraining?

iherman: > Should we say the MIME type should be IANA registered or is that overly constraining?

Imho, yes, we should.

index.html
talltree: I'm fine with the change, but when I read the WebRTC spec section Dan cited, I was not clear what that meant for this section. I hope we can discuss on the next Editor's call.

index.html
talltree: I'm happy to make this change but I thought the name of the Registries spec was being updated...

index.html
talltree: Yes, good improvement.

index.html
talltree: Agreed it should be MUST in both places.

index.html
talltree: Good suggestion.

index.html
talltree: Agreed. This was address in commit [dba5548](https://github.com/w3c/did-core/pull/241/commits/dba554852486db938ffff4f5baea42d40ec3ad89)

PR #253, opened 2020-04-08
Added DID resolution and dereferencing contracts.

jricher:
***
<a href="https://pr-preview.s3.amazonaws.com/jricher/did-core/pull/253.html" title="Last updated on Apr 20, 2020, 1:58 PM UTC (748fa4c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/253/5bab207...jricher:748fa4c.html" title="Last updated on Apr 20, 2020, 1:58 PM UTC (748fa4c)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -975,7 +975,8 @@ the use case.
   Adding a DID parameter to a <a>DID URL</a> means that the parameter becomes part of
   an identifier for a resource (the <a>DID document</a> or other). Alternatively, the
   <a>DID resolution</a> and the <a>DID URL dereferencing</a> functions can also be influenced
-  by passing input options to a <a>DID resolver</a> that are not part of the DID URL.
+  by passing input metadata to a <a>DID resolver</a> that are not part of the DID URL.
+	(See <a href="#meadata-structure"></a>).
   Such options could for example control caching or the desired encoding of a
   resolution result. This is comparable to HTTP, where certain parameters could
   either be included in an HTTP URL, or alternatively passed as HTTP headers
@@ -2604,21 +2605,21 @@ specification, but all implementations MUST implement a function in the form:
         </p>
         
         <p>
-<code>resolve ( did, input-options ) -> ( did-document, did-document-metadata, did-resolution-metadata )</code>
+<code>resolve ( did, input-metadata ) -> ( did-document, did-document-metadata, did-resolution-metadata )</code>
         </p>
         
         <p>
-The inputs of this function are a <a>DID</a> and a set of input options. The 
+The inputs of this function are a <a>DID</a> and a set of input metadata. The 
 <a>DID</a> is REQUIRED. Note that if the caller of this function wishes to resolve a <a>DID URL</a>, 
 the caller MUST first transform the <a>DID URL</a> into a bare <a>DID</a>, 
-including removal of any fragment. The input options are a map of key-value
+including removal of any fragment. The input metadata are a map of key-value
 string pairs as described in <a href="#metadata-structure"></a>.
-The input options are REQUIRED but the map MAY be empty. Concrete <a>bindings</a> MUST NOT
+The input metadata are REQUIRED but the map MAY be empty. Concrete <a>bindings</a> MUST NOT
 define additional inputs to this function.
         </p>
         
         <p>
-The possible keys for the input options are defined by the DID Core Registry [[?DID-CORE-REGISTRY]]. This specification
+The possible keys for the input metadata are defined by the DID Core Registry [[?DID-CORE-REGISTRY]]. This specification
 defines the following keys and values:
 
             <dl>
@@ -2661,7 +2662,7 @@ The contents of the DID document metadata still needs to be defined within this 
 The <code>DID resolution metadata</code> is returned as a map of key-value string pairs as described in
 <a href="#metadata-structure"></a>. This metadata contains information about the results of
 the resolution process. This metadata typically changes between invocations of the <a>DID resolution</a>
-function. The keys and possible values for this metadata are defined by a registry (TBD). This
+function. The keys and possible values for this metadata are defined by the DID Core Registry [[?DID-CORE-REGISTRY]]. This
 specification defines the following keys and values:
 
             <dl>
@@ -2723,16 +2724,16 @@ not represent.
         </div>
         
         <p>
-The inputs to this function are a <a>DID URL</a> and a set of input options. The
+The inputs to this function are a <a>DID URL</a> and a set of input metadata. The
 <a>DID URL</a> is REQUIRED and MAY be any valid <a>DID URL</a>, including a fragment.
-The input options are a map of key-value string pairs as 
-described in <a href="#metadata-structure"></a>. The input options are REQUIRED but 
+The input metadata are a map of key-value string pairs as 
+described in <a href="#metadata-structure"></a>. The input metadata are REQUIRED but 
 the map MAY be empty. Concrete <a>bindings</a> MUST NOT define additional
 inputs to this function.
         </p>
 
         <p>
-The possible keys for the input options are defined by a registry (TBD).
+The possible keys for the input metadata are defined by the DID Core Registry [[?DID-CORE-REGISTRY]].
         </p>
         
         <p>
@@ -2744,9 +2745,7 @@ from its conformant representation to allow for additional processing.
         
         <p>
 The <a>DID URL dereferencer</a> MUST perform any additional processing indicated
-by the <a>DID URL</a>, such as resolving a fragment into an element. The process
-for completing dereferencing is determined by the mime-type of the <a>DID document</a>
-returned from the <a>DID resolution</a> process. If no additional processing is
+by the <a>DID URL</a>, such as resolving a fragment into an element. If no additional processing is
 required by the <a>DID URL</a>, the <a>DID document</a> is returned as the 
 <code>content</code>.
         </p>
@@ -2756,7 +2755,7 @@ The <code>DID dereference metadata</code> is returned as a map of key-value stri
 <a href="#metadata-structure"></a>. The metadata structure is REQUIRED and MUST NOT be empty.
 This metadata contains information about the results of
 the dereferencing process. This metadata typically changes between invocations of the <a>DID URL dereferencing</a>
-function. The keys and possible values for this metadata are defined by a registry (TBD). This
+function. The keys and possible values for this metadata are defined by the DID Core Registry [[?DID-CORE-REGISTRY]]. This
 specification defines the following keys and values:
 
             <dl>
@@ -2766,7 +2765,7 @@ specification defines the following keys and values:
                 <dt>error</dt>
                 <dd>The error code from the dereferencing process. The <a>DID resolver</a> MUST supply
                     this value when there is an error. The possible values
-                    this field are defined by a registry (TBD). This specification defines the following
+                    this field are defined by the DID Core Registry [[?DID-CORE-REGISTRY]]. This specification defines the following
                     error values:
                     
                     <dl>
@@ -2800,8 +2799,8 @@ Metadata structure
         </h2>
 
         <div class="issue" data-number="58">
-There needs to be a registry established for handling input options and rules defined for 
-how to handle an unknown input option. A similar issue applies for all metadata structures
+There needs to be a registry established for handling input metadata and rules defined for 
+how to handle an unknown input metadata name. A similar issue applies for all metadata structures
 returned from these functions. 
         </div>
         
@@ -2818,7 +2817,7 @@ All <a>bindings</a> MUST provide a way for the caller of the function to input o
 receive the map of key-value string pairs as described in this specification. The 
 exact representation of these metadata structures as part of a protocol or serialization
 is out of scope for this specification. For example, an HTTP-based binding can
-use HTTP headers of an HTTP request to convey input options and HTTP headers of an
+use HTTP headers of an HTTP request to convey input metadata and HTTP headers of an
 HTTP response to convey document and resolver metadata. Alternatively, a fully
 JSON-based binding can use a JSON object to represent the metadata structures.
         </p>
--- a/index.html
+++ b/index.html
@@ -2034,7 +2034,8 @@ the data model, such as XML or YAML.
     <p>
 Producers MUST indicate which representation of a document has been used via a media type
 in the document's metadata. Consumers MUST determine which representation a document is in
-via a media type in the document's metadata. Consumers MUST NOT determine the representation
+via the <code>content-type</code> DID resolver metadata field. (See <a href="#did-resolution"></a>).
+Consumers MUST NOT determine the representation
 of a document through its content alone.
     </p>
 
@@ -2062,8 +2063,8 @@ JSON
         </h2>
 
         <p>
-When producing and consuming DID documents that are in plain JSON (as noted by
-the document metadata), the following rules MUST be followed.
+When producing and consuming DID documents that are in plain JSON (as indicated by
+a <code>content-type</code> of <code>application/json+did</code> in the resolver metadata), the following rules MUST be followed.
         </p>
 
         <section>
@@ -2204,8 +2205,8 @@ JSON-LD
         </p>
 
         <p>
-When producing and consuming DID documents that are in JSON-LD (as noted by
-the document metadata), the following rules MUST be followed.
+When producing and consuming DID documents that are in JSON-LD (as indicated by
+a <code>content-type</code> of <code>application/jsonld+did</code> in the resolver metadata), the following rules MUST be followed.
         </p>
 
         <ul>
@@ -2653,26 +2654,26 @@ can then parse and process the <a>DID document</a> from this byte stream.
         
         <p>
 When a <a>DID document</a> is returned, the DID document metadata is returned as a map of
-key-value string pairs as described in <a href="#metadata-structures"></a>. This metadata 
+key-value string pairs as described in <a href="#metadata-structure"></a>. This metadata 
 contains information about the input <a>DID</a> and the returned <a>DID document</a>. This metadata
 typically does not change between invocations of the <a>DID resolution</a> function. The keys and
 possible values for this metadata are defined by a registry (TBD). This component MAY be empty.
         </p>
 
         <p class="issue" data-number="203">
-The DID document metadata still needs to be defined within this document.
+The contents of the DID document metadata still needs to be defined within this document.
         </p>
 
         <p>
 The DID resolution metadata is returned as a map of key-value string pairs as described in
-<a href="#metadata-structures"></a>. This metadata contains information about the results of
+<a href="#metadata-structure"></a>. This metadata contains information about the results of
 the resolution process. This metadata typically changes between invocations of the <a>DID resolution</a>
 function. The keys and possible values for this metadata are defined by a registry (TBD). This
 specification defines the following keys and values:
 
             <dl>
                 <dt>content-type</dt>
-                <dd>The mime-type of the returned conformant representation of the <a>DID document<a> if successful.
+                <dd>The mime-type of the returned conformant representation of the <a>DID document</a> if successful.
                     The <a>DID resolver</a> MUST supply this value when a <a>DID document</a>
                     is returned. The <a>client</a> MUST use this value when determining how to
                     parse and process the byte stream returned by this function into
@@ -2687,7 +2688,7 @@ specification defines the following keys and values:
                     <dl>
                         <dt>invalid-did</dt>
                         <dd>The <a>DID</a> supplied to the <a>DID resolution</a> function does not
-                            conform to valid syntax.</dd>
+                            conform to valid syntax. (See <a href="#did-syntax"></a>.)</dd>
                             
                         <dt>unauthorized</dt>
                         <dd>The <a>client</a> is not authorized to resolve this <a>DID</a> with
@@ -2710,7 +2711,7 @@ DID URL Dereferencing
 The <a>DID URL dereferencing</a> function dereferences a <a>DID URL</a> into content
 represented by that <a>DID URL</a> based on the <a>DID document</a> represented by
 the <a>DID</a> within the <a>DID URL</a>. This content MAY be a <a>DID document</a>, a 
-service endpoint, or some other data. The details of the implementation of this function
+portion of a <a>DID document</a>, a service endpoint, or some other data. The details of the implementation of this function
 are outside the scope of this specification, but all implementations MUST provide
 a function of the following form:
         </p>
@@ -2722,7 +2723,7 @@ a function of the following form:
         
         <div class="note">
 It would be possible to define dereferencing in terms of a DID URL
-and a DID document, where the resolution process could happen externally
+and a fully parsed and modeled DID document, where the resolution process could happen externally
 to the dereferencer. The problem with this is that a client could potentially
 dereference a URL against a document that the DID within the DID URL does
 not represent.
@@ -2759,7 +2760,7 @@ required by the <a>DID URL</a>, the <a>DID document</a> is returned as the
         
         <p>
 The <code>DID dereference metadata</code> is returned as a map of key-value string pairs as described in
-<a href="#metadata-structures"></a>. This component is REQUIRED and MUST NOT be empty.
+<a href="#metadata-structure"></a>. This component is REQUIRED and MUST NOT be empty.
 This metadata contains information about the results of
 the dereferencing process. This metadata typically changes between invocations of the <a>DID URL dereferencing</a>
 function. The keys and possible values for this metadata are defined by a registry (TBD). This
@@ -2778,7 +2779,7 @@ specification defines the following keys and values:
                     <dl>
                         <dt>invalid-url</dt>
                         <dd>The <a>DID URL</a> supplied to the <a>DID URL dereferencing</a> function does not
-                            conform to valid syntax.</dd>
+                            conform to valid syntax. (See <a href="#did-url-syntax"></a>.)</dd>
                     </dl>
                     
                 </dd>
@@ -2788,8 +2789,9 @@ specification defines the following keys and values:
         
         <p>
 The <code>content</code> of this function's output is a byte stream of data. This byte stream MAY 
-represent a <a>DID document</a> in a conformant representation, the results of calling a service
-endpoint, or something else determined by the <a>DID method</a>. 
+represent a <a>DID document</a> in a conformant representation, a portion of a <a>DID document</a>
+in a conformant representation, the results of calling a service
+endpoint, or something else determined by the <a>DID method</a> and its services.
         </p>
         
         <p>
@@ -2824,7 +2826,8 @@ receive the map of key-value string pairs as described in this specification. Th
 exact representation of these metadata structures as part of a protocol or serialization
 is out of scope for this specification. For example, an HTTP-based binding can
 use HTTP headers of an HTTP request to convey input options and HTTP headers of an
-HTTP response to convey document and resolver metadata.
+HTTP response to convey document and resolver metadata. Alternatively, a fully
+JSON-based binding can use a JSON object to represent the metadata structures.
         </p>
         
         <p>
--- a/index.html
+++ b/index.html
@@ -2623,8 +2623,8 @@ specification, but all implementations MUST implement a function in the form:
 The inputs of this function are a <a>DID</a> and a set of input options. The 
 <a>DID</a> is REQUIRED. Note that if a client wishes to resolve a <a>DID URL</a>, 
 the client MUST first transform the <a>DID URL</a> into a bare <a>DID</a>, 
-including removal of any fragment component. The input options
-are a map of key-value string pairs as described in <a href="#metadata-structure"></a>.
+including removal of any fragment. The input options are a map of key-value
+string pairs as described in <a href="#metadata-structure"></a>.
 The input options are REQUIRED but the map MAY be empty. Concrete <a>bindings</a> MUST NOT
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -2543,18 +2588,269 @@ <h1>
...
+specification, but all implementations MUST implement a function in the form:
+        </p>
+        
+        <p>
+<code>resolve ( did, input-options ) -> ( did-document, did-document-metadata, did-resolution-metadata )</code>
+        </p>
+        
+        <p>
+The inputs of this function are a <a>DID</a> and a set of input options. The 
+<a>DID</a> is REQUIRED and MUST NOT include a fragment component. The input options
```
kdenhartog, line 2622: interesting to see that did-urls must not be included here. We've implemented it compliant with this, but I've found that it would be rather useful for the resolver to handle de-referencing to be quite useful in most cases. Especially when it's being done by a local library rather than remotely. What's your thinking behind forcing this separation? Am I missing a key insight that I should be considering?

index.html
```diff
@@ -2543,18 +2588,269 @@ <h1>
...
+                    <a>DID document</a>.</dd>
+
+                <dt>error</dt>
+                <dd>The error code from the resolution process. The <a>DID resolver</a> MUST supply
+                    this value when there is an error. The possible values
+                    this field are defined by a registry (TBD). This specification defines the following
+                    error values:
+                    
+                    <dl>
+                        <dt>invalid-did</dt>
```
kdenhartog, line 2689: I think it would be useful to include errors around network failure and other failures that occur within the resolution process that could be encompassed together. Something like a Resolution failure error.

index.html
llorllale, line 2622: > MUST NOT include a fragment component

Is there a need to specify this? For starters, `#` is not allowed as per the rules in the generic syntax.

jricher: Yes, it's not technically needed here, but I think it at least warrants a note because of the lazy way people refer to a "DID URL" as a "DID". Being clear that it doesn't have a fragment is helpful.

jricher: It can be done with a single piece of software but it's two different functions with different inputs. What you're talking about is something that can do both dereferencing and resolution, and ultimately the core function of what you're describing is the dereferencing function below, which includes resolution as a sub-step.

index.html
jricher: I agree but didn't want to get into that discussion in this pull request. That's for the WG to decide once the structure is in place. Right now there's not even a way to communicate errors to anything.

index.html
kdenhartog: Ahh ok, that makes sense given the dereference wasn't pulling in a did document as well.

index.html
kdenhartog: Sounds good, makes sense to hold off on that until we get a first cut at it in.

index.html
llorllale: But then now I ask: why not also say that query parameters must also be excluded, or anything else that renders this input non-compliant with the generic syntax.

Anyway, I'm just pointing out words that, to me, distract from the meat of what this text is getting to. I don't really have a strong opinion on whether to include it or exclude it.

jricher: I think it might make sense for it to be an inline note instead of a normative requirement for those reasons.

jricher:
@kdenhartog I'm not sure what you mean by "a document loader". If you already have the DID document, you don't need to do resolution or dereferencing -- those are used to GET the document (or some portion of it). If you're talking about loading an item referred to by a DID from within a document's field, then yes you'd call `resolve` if it's a DID and `dereference` if it's a DID URL, from within the document loader.

index.html
```diff
@@ -2543,18 +2588,269 @@ <h1>
...
+defines the following keys and values:
+
+            <dl>
+                <dt>accept</dt>
+                <dd>The mime-type of the preferred representation of the <a>DID document</a>. The <a>DID resolver</a> MAY 
+                    use this value to determine the representation of the returned <a>DID document</a> if such a 
+                    representation is supported and available.</dd>
+            </dl>
+        </p>
+
```
iherman, line 2639: Isn't it possible to allow for a plural here, ie, a resolver may list several mime types that are accepted (say, JSON _or_ CBOR). This is pretty much like the HTTP Accept header after all.

kdenhartog:
> 
> 
> @kdenhartog I'm not sure what you mean by "a document loader". If you already have the DID document, you don't need to do resolution or dereferencing -- those are used to GET the document (or some portion of it). If you're talking about loading an item referred to by a DID from within a document's field, then yes you'd call `resolve` if it's a DID and `dereference` if it's a DID URL, from within the document loader.

 In regards to what I meant by document loader, it was in reference to the function that I commonly see as a parameter in JSON-LD libraries. As for what you explained, around "loading an item referred to by a did..." this makes sense. That answer aligns with what I've seen, and I wanted to clarify in the event it potentially caused misalignment in other areas of this issue.

index.html
jricher: I agree that for this specific field a plural value makes sense. Supporting it is another story that's less clear. I think we've got a few choices here: 

1. we allow multiple `accept` fields. I am strongly against this because it breaks the model described here, which currently maps nicely to JSON objects and other simple data structures like a Java Map interface
2. We allow rich objects like arrays as the values. I am strongly against this because it breaks the simple model here, and I think keeping the values as plain strings is going to be important in interoperability across multiple platforms and systems where different data structures aren't as prevalent. We also don't want to get into a world of recursive objects for the metadata. 
3. We allow the field itself to define an internal syntax to express multiple values. This is already allowed by the surrounding text, but it's just that this field needs to define what that means. We can do comma separation, space separation, or just take the ABNF from the HTTP Accept header. I am in favor of this approach, but would suggest we add a new issue to define the details outside of this pull request.

index.html
```diff
@@ -2543,18 +2588,271 @@ <h1>
...
+        </h2>
+
+        <div class="issue" data-number="58">
+There needs to be a registry established for handling input options and rules defined for 
+how to handle an unknown input option. A similar issue applies for all metadata structures
+returned from these functions. 
+        </div>
+        
+        <p>
+Metadata structures are maps of key-value string pairs. All keys within this map
```
msporny, line 2818:
```suggestion
Metadata structures are maps of property-value string pairs. All properties within this map
```

msporny: Either property-value or name-value... prefer property-value as that's what's used elsewhere... forget the issue number, but this is being actively discussed.

index.html
```diff
@@ -2543,18 +2588,271 @@ <h1>
...
+MUST be UTF-8 unicode strings. The keys of this map MUST NOT be repeated within 
+a given metadata structure. The keys of this map MUST be compared in a 
+case-insensitive way and the case MAY be transformed. The values of this map
```
msporny, line 2821: -1 to case-insensitive comparison. I understand why it's desired, but modern software practices often do matching based on case sensitivity... this is a strange rule. What is the reason it exists? We're worried about developers not understanding case-sensitive matching?

index.html
```diff
@@ -2543,18 +2588,271 @@ <h1>
...
+        
+        <p>
+<code>DID-CONTENT-TYPE: application/json+did</code>
+        </p>
+        
+        <p>
+In this example, the key <code>content-type</code> has had the string <code>DID-</code>
+prepended to it and the key text is uppercased. When returned from the function, this transformation
+would be reversed to provide the value <code>application/json+did</code> under the key
+<code>content-type</code> to the client. In other words, the client MUST NOT need
```
msporny, line 2850: "client" and "caller" are being used interchangeably in this PR and it's confusing if something different is meant when one is used over the other. If there is a difference, it's so nuanced that I can't tell. Are we talking about clients, callers, or both?

msporny: How would we test this "MUST NOT" normative statement? I'm struggling to think about what test would be written.

terms.html
```diff
@@ -123,12 +123,23 @@
 verifiable data registry. For more information, see [[VC-DATA-MODEL]].
   </dd>
 
+  <dt><dfn>DID resolution</dfn></dt>
+  
+  <dd>
+The function that takes as its input a <a>DID</a> and a set of resolution
+options and returns a <a>DID document</a> (in a conformant serialization)
```
msporny, line 130:
```suggestion
options and returns a <a>DID document</a> in a conforming representation
```

terms.html
```diff
@@ -154,6 +165,27 @@
 by a <a>DID fragment</a>.
   </dd>
 
+  <dt><dfn>DID URL dereferencing</dfn></dt>
+  
+  <dd>
+The function that takes as its input a <a>DID URL</a>, a <a>DID document</a>, 
+and a set of dereferencing options and returns a resource, which may be
```
msporny, line 172:
```suggestion
and a set of dereferencing options and returns a resource, which might be 
```

index.html
```diff
@@ -630,7 +630,51 @@ <h2>DID Documents</h2>
...
+        </dd>
+        <dt><dfn>binding</dfn></dt>
+        <dd>
+A concrete mechanism through which a <a>client</a> invokes a <a>DID resolver</a> or a <a>DID URL dereferencer</a>.
+This could be a local command line tool, a library API, or a network call such as an HTTP(S) request.
+        </dd>
+      </dl>
+      
+      <p>
+The general contracts of the <a>DID resolution</a> and <a>DID URL dereferencing</a> functions
```
msporny, line 674: Using the word "contract" here (and through the rest of the PR) could be confusing. We should use simpler language, like "the inputs to and outputs from the DID resolution and DID URL derefrencing functions are..."

terms.html
```diff
@@ -154,6 +165,27 @@
...
+The dereferencing function then can perform additional processing on the <a>DID document</a> to return
+the dereferenced resource indicated by the <a>DID URL</a>. The contract of this 
+function is defined in <a href="#did-url-dereferencing"></a>.
+  </dd>
+
+  <dt><dfn data-lt"DID URL dereferencers">DID URL dereferencer</dfn></dt>
+  
+  <dd>
+A software and/or hardware system that is capable of executing the <a>DID URL dereferencing</a>
+function, retrieving some form of data (which may or may not be a <a>DID document</a>) for a given
```
msporny, line 184:
```suggestion
function, retrieving some form of data, which might or might not be a <a>DID document</a>) for a given
```

index.html
```diff
@@ -2543,18 +2588,271 @@ <h1>
...
+        <p>
+The inputs of this function are a <a>DID</a> and a set of input options. The 
+<a>DID</a> is REQUIRED. Note that if a client wishes to resolve a <a>DID URL</a>, 
+the client MUST first transform the <a>DID URL</a> into a bare <a>DID</a>, 
+including removal of any fragment component. The input options
+are a map of key-value string pairs as described in <a href="#metadata-structure"></a>.
+The input options are REQUIRED but the map MAY be empty. Concrete <a>bindings</a> MUST NOT
+define additional inputs to this function.
+        </p>
+
```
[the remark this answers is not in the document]
kdenhartog: Given my strong opinion on the subject, I'll plan to attend.

index.html
kdenhartog: issue #204 for anyone reading and looking for reference.

jricher:
@msporny The entire point of a contract is to make normative statements about the edge of the resolution process without making requirements for the internals of the resolution process. That's what this pull request does.

index.html
```diff
@@ -938,18 +982,18 @@ <h2>Generic DID URL Parameters</h2>
...
-  by passing options to a <a>DID resolver</a> that are not part of the DID URL.
+  Adding a DID parameter to a <a>DID URL</a> means that the parameter becomes part of
+  an identifier for a resource (the <a>DID document</a> or other). Alternatively, the
+  <a>DID resolution</a> and the <a>DID URL dereferencing</a> functions can also be influenced
+  by passing input options to a <a>DID resolver</a> that are not part of the DID URL.
   Such options could for example control caching or the desired encoding of a
   resolution result. This is comparable to HTTP, where certain parameters could
   either be included in an HTTP URL, or alternatively passed as HTTP headers
   during the dereferencing process. The important distinction is that DID
   parameters that are part of the <a>DID URL</a> should be used to specify
```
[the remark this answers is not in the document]
jricher: These changes are editorial fixes outside this pull request and therefore I am not inclined to include them in order to keep pull this pull request smaller and more focused. Please submit a separate pull request to change this language instead of addressing it here.

index.html
```diff
@@ -938,18 +982,18 @@ <h2>Generic DID URL Parameters</h2>
...
-  <i>what resource is being identified</i>, whereas <a>DID resolver</a> options
+  <i>what resource is being identified</i>, whereas <a>DID resolution</a> options
   that are not part of the <a>DID URL</a> should be use to control <i>how that
```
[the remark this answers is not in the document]
jricher: Same as above: These changes are editorial fixes outside this pull request and therefore I am not inclined to include them in order to keep pull this pull request smaller and more focused. Please submit a separate pull request to change this language instead of addressing it here.

index.html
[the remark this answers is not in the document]
jricher: I strongly disagree with @msporny here. The language here defines things in terms of a functional API and not a network protocol. As discussed many times this falls under this bullet of our charter:

*  Establish a deterministic mapping between DID method identifiers and the resolution process used to resolve that DID method. 

It can't be deterministic without normative language. The suggestion to just point to a NOTE and "Signal" to people that it's "going to be a real spec some day" is just acrobatics and avoiding the issue, which is not what we had agreed upon in Amsterdam.

index.html
```diff
@@ -2543,18 +2588,271 @@ <h1>
...
+        <p>
+The <a>DID resolver</a> executes the "Read" operation of the <a>DID method</a> as described in
+<a href="#read-verify"></a>. If successful, the outputs of this function MUST contain a <a>DID document</a>, 
+a set of metadata for the DID document (which MAY be empty), and a set of metadata for the resolution process. If
+the function results in an error, the outputs MUST contain a set of metadata for the resolution process.
+        </p>
+        
+        <p>
+The <a>DID document</a> is returned as a byte stream of a conformant representation
+as determined and supported by the <a>DID resolver</a>. This component is REQUIRED unless an error is returned
```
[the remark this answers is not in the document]
jricher: It's referring to "DID document", one of the components in the returned value. Do you have a suggested alternative term? We could just repeat "DID Document" here.

index.html
```diff
@@ -2543,18 +2588,271 @@ <h1>
...
+                </dd>
+            
+            </dl>
+        </p>
+        
+    </section>
+
+    <section>
+        <h2>
+DID URL Dereferencing
```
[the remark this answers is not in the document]
jricher: As above, I strongly disagree with @msporny 's suggested edits and direction.

index.html
jricher: Will leave as is pending resolution of #204 which can change it to be consistent through the document.

index.html
jricher: This is how HTTP header values are compared. I don't see it as a necessary feature as long as a binding is still required to have a deterministic mapping between the key and the binding's representation.

index.html
jricher: The client is defined as the component that calls the function, above. However it is awkward to say the "client of the function". Would appreciate suggested text.

index.html
```diff
@@ -938,18 +982,18 @@ <h2>Generic DID URL Parameters</h2>
         </p>
 
         <p>
-  Adding a DID parameter to a DID URL means that the parameter becomes part of
-  an identifier for a resource (the DID document or other). Alternatively, the
-  DID resolution and the DID URL dereferencing processes can also be influenced
-  by passing options to a <a>DID resolver</a> that are not part of the DID URL.
+  Adding a DID parameter to a <a>DID URL</a> means that the parameter becomes part of
+  an identifier for a resource (the <a>DID document</a> or other). Alternatively, the
```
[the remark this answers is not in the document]
jricher: This was text not changed by this pull request apart from adding the link.

index.html
jricher: I'm fine with that adjustment, I used "contract" here and throughout as it's a well established programming term and was the term by which this WG had been discussing the issue.

msporny:
> @msporny The entire point of a contract is to make normative statements about the edge of the resolution process without making requirements for the internals of the resolution process. That's what this pull request does.

Yes, and after having read two attempts at doing this in the DID Core spec, I'm no longer supportive of doing so in the DID Core spec. It confuses concerns -- the layering is wrong. Let's not do something half-baked, let's put this PR in DID Resolution, get clear about where normative statements should go about resolution, and make that the next focus of the group.

I'll submit some counter-proposals to this one in an attempt to provide some additional constructive input.

index.html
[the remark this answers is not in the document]
msporny: Yes, please repeat "DID Document" so that it's clear what is required.

index.html
msporny: > This is how HTTP header names are compared.

Yeah, and I question whether or not we should let esoteric HTTP things leak into this spec. There are arguments for and against, but that design in HTTP comes from a time when case sensitivity didn't matter as much as it does today.

index.html
msporny: If your argument is that we're merely outlining functions here and not protocol - then I suggest staying away from client. Alternatives include "caller of the function", "when the function is executed" or "consumer of the output".

Out of all of those, I'd prefer "when the function is executed", but it doesn't work here. Something like the following might be better:

> In other words, a consumer of <a>DID resolution</a> output MUST NOT need to know the underlying representation of the metadata used by the implementation of the function.

Doing so doesn't give me any more insight into how we'd write a test in a test suite for this item.

jricher:
This PR does not belong in DID resolution. Without this structure, the rest of this document makes no sense. This is not a layering violation, and without this, there is in fact **no layering** in the document. The edges of everything are too fuzzy, which is why we have endless arguments about what's a parameter on the DID and what goes into the document. We need to have normative language about how the pieces fit together or else nobody will agree on where the definitions of the pieces start and stop, let alone how they might fit together in any fashion.

I believe that the language of this PR strikes a balance of defining what we need in terms of connections without defining at all what goes on within those connections. This is why I pushed back against the language in #247 as it went into too much detail on how to fulfill the contracts, with language on local vs. remote bindings and the discussions of underlying protocol structures. As this PR is written, it simply says you have to take in A and B, and you have to output X, Y, and Z. It says nothing about how you get from one to the other, and all existing DID methods should be able to comply with this pretty easily.

This PR does not define the mechanics of DID resolution beyond what is needed to understand the connection between a DID and its DID Document. If we take this out, or make it non-normative, then we need to remove all of section 9 (Methods) as well, especially 9.2 (Method Operations), as that goes into even greater depth than this at defining the required functions for DID Method implementations and creates just as much of a "protocol", if not more so. To be clear, I believe removing section 9 would be an egregious mistake, as I believe that not having the clear normative definitions that I am proposing here would be an egregious mistake.

Finally, If you have specific concerns about the requirements upon method implementations posed by these function definitions, then I would like to know what those are as I believe that such details can be worked through.

I look forward to reading and reviewing your counter-proposals and continuing this discussion.

index.html
jricher: I'm fine with having more precise definitions to use here, as **client** is a fairly generic term. 

I'm not sure what you mean about the test suite. These function definitions actually make it much easier to write a test suite, and in fact make it easier to test the entire rest of the specification. You simply have the test suite implement a hook for `resolve` which is then implemented by whatever adapters you need to call different methods to do the actual resolution. The test suite code then doesn't need to change whether it's testing an implementation of the input of a function or the implementation of an output of a function or the implementation of tying those two together. So your test suite would test this by sending in a set of standard or common inputs and expecting a set of standard or common outputs. If those work, it passes. The test suite doesn't care what happens underneath as long as the contract is fulfilled, which is the very nature of a test suite.

index.html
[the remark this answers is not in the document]
mavarley: The language in this section seems abstract enough to define what is required for DID Resolution without defining a protocol (client-server) although that can certainly be interpreted or derived. It seems like some language like this defines what is meant/required for DID Resolution, and then a NOTE can be used to be more specific in what a DID Resolution protocol may be. My $0.02.

index.html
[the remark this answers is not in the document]
jricher: Done.

index.html
jricher: I'm not convinced that it matters more or less today, and so I'd be fine with rolling that back if the WG wants to go that way.

index.html
jricher: I've removed the term "client" from this section, even though it's used elsewhere in the document.

dlongley:
At a high level, there are two important pieces of "DID resolution" that must be defined in specs:

1. The thing a DID resolves to is a DID Document.
2. When two different pieces of software resolve a DID, they must get the *same* DID Document, and to do that they follows steps X, Y, and Z.

I believe we're chartered to do the former and not the latter. At this point, I think this PR only does the first (maybe we still need a tweak here or there, but it's on target). I also think that people will be able to write software that would pass tests for the new normative language here that could fail tests for the second piece. And that's a good thing; it means we've created the right separation.

We need to, as a group, stop saying we're not chartered to work on DID Resolution ... because I think that's very confusing. We are chartered to work on a piece of it, just not all of it.

kdenhartog:
First off, thank you to everyone on the call that helped discuss this further. I think we're working towards alignment in a positive manner. After we got off the call I started thinking about some new angles of how we could approach this riffing on the idea of "_a did MUST identify a did document that contains an id name-value pair with the value being that same did_".

If we went down this road, it at least makes it clear that we've got a "deterministic mapping" between the identifier and the data model that it identifies. The edge cases I think everyone here wants to avoid are ones like `urn:did-like:123 => { "@id": "did:example:123" }` and `did:example:123 => {}`. As far as I can tell, there's nothing invalid about these deterministic, but clearly out in left field, mappings. I'm sure we also REALLY don't want to see things like this did:example:123#key1 => 
```
{ 
   "@id": "did:example:123#endpoint1", 
   "type": "CredentialRepositoryService",
   "serviceEndpoint": "https://repository.example.com/service/8377464"
}
```

because then we've gotten to a point where interoperability is going to be nearly impossible between different implementations without behind the scenes coordination and we're allowing so much ambiguity that it leaves me scratching my head and wondering what do dids actually identify?

With that in mind, I think reasonably speaking what we're all after here is dids, did-urls, and did documents that are a means to an end, not the end in and of itself. To do that, I just can't seem to understand how it's valuable to us to leave so much room for ambiguity that did methods become their own silos. I think statements like:

> The DID resolution function resolves a DID into a DID document by using the "Read" operation of the applicable DID method. The details of how this process is accomplished is outside the scope of this specification, but all implementations MUST implement a function in the form:

are incredibly useful for reducing the ambiguity in this spec and removing the edge cases like the ones above. So, let's make that the main objective here while keeping the spec simple and allow us to deliver the value that we all recognize dids and did documents, are useful for.

msporny:
There is a line in our current charter that was a topic of debate in our DID WG special call today. Namely, this one:

> Establish a deterministic mapping between DID method identifiers and the resolution process used to resolve that DID method.

Interpretations of what that sentence means have been varied and have spilled into the DID Resolution discussion. If the DID WG takes a consistent (and official) position on the interpretation of that statement, it'll help us figure out how to bring this DID Resolution discussion to a close.

Hindsight being 20/20, that statement probably should have said something more along these lines:

Define a concrete, generalized, and extensible DID resolution process for DID Method implementers that transforms a DID into a DID Document.

Specifically, this means that:

* It is in scope for the DID WG to define a concrete process that takes a DID as input and provides a DID Document as output.

* It is in scope to make that process take in options and provide back a document along with different classes of metadata (e.g., subject, document, and resolution metadata).

* It is in scope to add resolution tests to the test suite that exercise the generalized DID resolution process in the specification on concrete DID Method implementations.

* It is out of scope to normatively define DID Method specific details of implementing DID resolution.

* It is out of scope to normatively define DID Resolution protocols and representation formats.

* It is out of scope to test concrete DID Resolution protocols and data formats beyond the necessary process to demonstrate interoperability between the test suite and an implementation.

If we can get consensus on the above, it'll provide clarity around exactly how far we can take this DID Resolution discussion and exactly what is in scope and out of scope.

My request is that the Chairs put this (or some variation) to the WG during the next call where getting consensus on this would be an appropriate use of everyone's time.

/cc @brentzundel @burnburn @iherman

index.html
jricher: I've removed the term contract from the pull request.

talltree:
Manu, I think you did a great job of interpreting what was meant by that
line in the charter. So +1 to your proposal.

=Drummond

kdenhartog:
> 
> 
> At a high level, there are two important pieces of "DID resolution" that must be defined in specs:
> 
>     1. The thing a DID resolves to is a DID Document.
> 
>     2. When two different pieces of software resolve a DID, they must get the _same_ DID Document, and to do that they follows steps X, Y, and Z.
> 
> 
> I believe we're chartered to do the former and not the latter. At this point, I think this PR only does the first (maybe we still need a tweak here or there, but it's on target). I also think that people will be able to write software that would pass tests for the new normative language here that could fail tests for the second piece. And that's a good thing; it means we've created the right separation.
> 
> We need to, as a group, stop saying we're not chartered to work on DID Resolution ... because I think that's very confusing. We are chartered to work on a piece of it, just not all of it.

While I'd like to see 2 get answered too so we can avoid examples like I provided in my previous comment, I could live with this approach and think it's still a huge step in the right direction.

peacekeeper:
One thought about the metadata section.. I agree with this language:

> All bindings MUST provide a way for the caller of the function to input or receive the map of key-value string pairs as described in this specification. The exact representation of these metadata structures as part of a protocol or serialization is out of scope for this specification. For example, an HTTP-based binding can use HTTP headers of an HTTP request to convey input metadata and HTTP headers of an HTTP response to convey document and resolver metadata. Alternatively, a fully JSON-based binding can use a JSON object to represent the metadata structures. 

However, I don't like the `DID-CONTENT-TYPE` example very much. It goes too much into implementation, and I don't think it's a particularly realistic example (more likely a concrete HTTP-based binding will just use the standard HTTP `Content-Type` header). Can we remove that whole example?

Also, I think an important point in my original PR https://github.com/w3c/did-core/pull/247 that may have gotten a bit lost here is that metadata may be linked to the representation of the DID document itself. The metadata may be part of the DID document, or there may be a higher-level data structure that contains both the DID document and its metadata (this is what the DID Resolution spec currently calls a [DID Resolution Result](https://w3c-ccg.github.io/did-resolution/#did-resolution-result)). I can create a suggestion to add this idea back into the text, should only be a single sentence.

msporny:
> Let's get this first layer agreed too, before we talk about how to do parsing, or validation.

@jricher, this request came about after the DID WG Editor/Chair discussion last week. The result of the conversation last week was mixed, where some felt like we didn't get anywhere, and others felt like there was more clarity.

The end result is that none of the Editors felt like they could channel what you wanted and would rather see you take the current spec and define just the interfaces. Note that this would mean that there would be no mention of algorithms, normative statements around processing, what is processed where, etc. Meaning, large chunks of your original PR would have to be taken out.

The hope is that once we can get a very minimal thing on the inputs and outputs agreed to, we can decide how far to take things.

Do you feel like you would be able to put such a PR together by the special topic call this week?

jricher:
@OR13 I started with it all together because all of the parts depend on each other in ways that are easy to miss when you only see the pieces. Those interlinks are important, but nonetheless I see the value in minimizing a change set. Manu already tried to split it into parts, and a small piece of it of that has already made it into the specification. I agree that there shouldn't be an algorithm, which is why this pull request doesn't have one listed in here. #263 is nearly the same text for the resolution section, but #286 is a separate pull request alltogether -- it added an algorithm, removed normative constructs, changed data requirements, and otherwise deeply altered the content and of this PR. It's a very different beast.

As for the fetch spec, that's the pattern I tried to use here, but without tying it to a specific programming language or platform (which is what fetch can do). I'll note that the entire fetch spec is normative and algorithmic, though, so I'm not sure how you can describe it otherwise. The function statement doesn't mean anything without the additional text surrounding it that describes what it does and how it works. 

My pull request here did not try to define all actions and error conditions, but it did try to define a common set of data structures and ways to tie them together. This way, when the group decides what errors and actions need to be taken, we have a place to put them and a way to talk about them. This is why there isn't a strict processing algorithm but there is a strict definition of what the data types are, and what goes where.

@msporny I will gladly work with the editors to incorporate the changes that I and many others in the working group believe this spec needs to be workable. I do not think it makes sense to work on things in isolation, though. I only think it makes sense if all the pieces are available, even if they're in separate pull requests. I am very worried about a non-normative piece of guidance being accepted for this and all of the issues marked as resolved, as I have stated previously.

I look forward to continued work and discussion.

dmitrizagidulin:
@OR13 
> I would love to see just the interface defined first...no algorithm, not a lot of normative spec text, just a signature... similar to https://fetch.spec.whatwg.org/#body-mixin

+1, very much agree to the WHATWG fetch() approach to the resolution layers.

msporny:
PR #331 has been merged. @jricher, can we close this PR now?

peacekeeper:
I agree that after merging https://github.com/w3c/did-core/pull/331, this can probably be closed. If there is anything in this PR that we still want to get into the spec (e.g. language around bindings and transformations), then this can still be added on top of the current spec in a separate PR. @jricher do you agree?

Also, I created a new issues to track DID URL dereferencing: https://github.com/w3c/did-core/issues/364. This PR here has some content about DID URL dereferencing that can potentially be re-used in a new PR to address this issue.

msporny:
Closing this in 48 hours if we don't hear back from the PR author.

msporny:
No response from original PR author, closing.

Issue #276, opened 2020-05-11
Is "verification method" relevant to inbound (as opposed to outbound) communication?


dhh1128:
The spec currently describes verification methods as follows:

>Verification methods "provide a set of mechanisms enabling a DID controller to prove control of the DID." (Abstract)

>Verification method: A set of parameters required to independently verify a proof, such as an identifier for a public/private key pair that would be used in the proof. (section 2)

Both of these explanations convey the idea that the DID controller is proving the legitimacy of an action to an outside audience. However, when we apply the notion of verification methods to the problem domain of general communication, some friction shows up. For example, how does Alice say that one or more of her keys are appropriate targets for encryption of __inbound__ communication (sent to her as opposed to emitted by her)?

We could create an item off the root of the DID doc called, for example, "readEncrypted", and Alice could define or reference there all the public keys that are usable as encryption targets when talking to her. Such an item would look and behave very much like a verification method block like "keyAgreement", but it wouldn't be a method for external parties to verify actions by Alice. If anything, it would be the opposite -- a method for Alice to verify that something from external parties was really intended for her. And we could explain this either by broadening the definition of "verification method" or by explaining the section as being like one for a verification method, but not fitting in the category.

Alternatively, we could say that "keyAgreement" is the right section to use for this, and make the assumption that the set of keys Alice uses to encrypt outbound is always exactly equal to the set of keys that Alice uses to decrypt inbound. This assumption is clearly not guaranteed to be true (a sensor could emit encrypted data on Alice's behalf, but be incapable of receiving anything; a sink could consume encrypted data inbound, but never emit anything) -- but perhaps it's close enough?

Generalizing still further, do we need to contemplate inbound actions in general, of which decryption is just a special case? If so, what do we call the mechanism for constraining or specifying or configuring them? "Verification method" feels a bit odd, if we take the concept that far. Something like "configuration method" might fit better.

I'm not necessarily arguing for the broader term, but I do need a way to specify inbound encryption keys, and I'd like the spec to use language that lets a mental model for such things to fit nicely.

dhh1128:
Cross-referencing #268

dhh1128:
Based on the resolution of [this issue](https://github.com/decentralized-identity/didcomm-messaging/pull/48#issuecomment-628204677) over in the DIDComm repo, I'm now comfortable using `keyAgreement` as the container for all the keys used to do DH and share symmetric keys for encryption in DIDComm. 

[...]

Issue #283, opened 2020-05-13
Verification method block should be a first citizen like public keys


awoie:
Public keys in the public key section can be referenced in other sections of the DID Doc, e.g., `authentication`. That is great because it makes the DID Doc smaller in case the same public key is used for multiple purposes. 

Verification methods should have the same option. 

As per https://www.w3.org/TR/did-core/#authentication:
> Each verification method MAY be embedded or referenced. 

ATM, referencing verification methods other than public keys is not possible. IMO, it would make more sense to introduce `verificationMethod` as a new first citizen block. That block would list the verification methods, so that they can be referenced in other sections of the DID Doc, e.g., authentication, in the same way as public keys.

Another approach would be to just rename the `publicKey` section in `verificationMethod` which might be even the better approach.

An example of a verification method that would benefit from this approach is `ethereumAddress`.

awoie:
@OR13 @dlongley In case there is also a way to achieve that behaviour, please provide an example and we can close that issue. It might be even worth to provide an example in the specification. I could probably also live with having an `ethereumAddress` array similar to `publicKey` and just referring to those elements within the `authentication` section.

OR13:
This is essentially a request to rename `publicKey` `verificationMethod`... which IMO is an excellent idea for a number of reasons... 

In an ideal world, we don't reuse keys for things, so there is no size benefit to linking, since each key is declared only in the block for which it is used....

In the real world, its extremely common to declare a key once in `publicKey` section, and then reference it from all the other locations...

For DID Methods that don't have a default public key... like `did:ethr` this creates an awkward situation where we are listing `ethereumAddress` under `publicKey`.... when it's clearly not one.

IMO, this is what a `did:ethr` which has not been updated should look like:

```
{
    "@context": [
        "https://www.w3.org/ns/did/v1",
        "https://identity.foundation/blockchain-identity/v1"
    ],
    "id": "did:ethr:0x74fd35392f1c2a97c6080e0394b1995b0e63c8bf",
    "verificationMethod": [
      {
        "id": "did:ethr:0x74fd35392f1c2a97c6080e0394b1995b0e63c8bf#0",
        "type": "EcdsaSecp256k1RecoveryMethod2020",
        "controller": "did:ethr:0x74fd35392f1c2a97c6080e0394b1995b0e63c8bf",
        "ethereumAddress": "0x74fd35392f1c2a97c6080e0394b1995b0e63c8bf"
      }
    ],
    "authentication": [
[...]

PR #284, opened 2020-05-13
Amplify/broaden def of 'verification method'

dhh1128:
Signed-off-by: Daniel Hardman <daniel.hardman@gmail.com>

See https://github.com/w3c/did-core/issues/276

### DIFF
--- a/terms.html
+++ b/terms.html
@@ -269,14 +269,14 @@ signer possessed the associated private key.
     </p>
     <p>
 "Verification" and "proof" in this definition are intended to apply broadly. A public
-key may be used during Diffie-Hellman key exchange to negotiate a shared symmetric
+key might be used during Diffie-Hellman key exchange to negotiate a shared symmetric
 key for encryption. This guarantees the integrity of the key agreement process. It
 is thus another type of verification method, even though descriptions of the process
-may not use the words "verification" and "proof."
+might not use the words "verification" and "proof."
     </p>
     <p>
-Verification methods may take many parameters. An example of this might be a set of five
-cryptographic keys from which any three must contribute to a threshold signature. Methods
+Verification methods might take many parameters. An example of this is a set of five
+cryptographic keys from which any three are required to contribute to a threshold signature. Methods
 need not be cryptographic. An example of this might be the contact information for a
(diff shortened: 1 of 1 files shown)


peacekeeper:
The only aspect of this PR I find problematic is that it adds a lot of text to the Terminology section. All other terms in that section consist of a single paragraph, whereas this one has three. Perhaps most of this new text should be moved elsewhere (especially if a new `verificationMethod` core property is introduced - as suggested in https://github.com/w3c/did-core/issues/283).

terms.html
```diff
@@ -261,9 +261,26 @@
...
-proof.
+    <p>
+A set of parameters that can be used to independently verify a proof according to
+a particular method. For example, a public key can be used as a verification
+method with respect to a digital signature; in such usage, it verifies that the
+signer possessed the associated private key.
+    </p>
+    <p>
+"Verification" and "proof" in this definition are intended to apply broadly. A public
+key may be used during Diffie-Hellman key exchange to negotiate a shared symmetric
```
msporny, line 272:
```suggestion
key might be used during Diffie-Hellman key exchange to negotiate a shared symmetric
```

terms.html
```diff
@@ -261,9 +261,26 @@
...
+key for encryption. This guarantees the integrity of the key agreement process. It
+is thus another type of verification method, even though descriptions of the process
+may not use the words "verification" and "proof."
```
msporny, line 275:
```suggestion
might not use the words "verification" and "proof."
```

terms.html
```diff
@@ -261,9 +261,26 @@
...
+    </p>
+    <p>
+Verification methods may take many parameters. An example of this might be a set of five
```
msporny, line 278:
```suggestion
Verification methods might take many parameters. An example of this is a set of five
```

terms.html
```diff
@@ -261,9 +261,26 @@
...
+cryptographic keys from which any three must contribute to a threshold signature. Methods
```
msporny, line 279:
```suggestion
cryptographic keys from which any three are required to contribute to a threshold signature. Methods
```

PR #286, opened 2020-05-17
Add normative DID Resolution section. (option #2)

msporny:
This PR refactors PR #263 by doing the following things:

* Takes a less-prosey / more algorithmic approach to establish how the inputs are provided and checked to the DID Resolution process and how the outputs are received and validated from the DID Resolution process.
* Further aligns how properties defined in DID Core are reflected in DID Spec Registries.
* Splits the resolution options, DID document metadata, and resolution metadata sections into their own subsections and points to the DID Spec Registries in each to ensure the normative requirements flow down from the top-level algorithm.
* Uses stronger and more accurate, testable normative wording while clarifying and/or eliminating some redundant statements.

This is a WIP, do not merge.

Human readable preview is here:

https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/286.html#did-resolution

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/286.html" title="Last updated on May 18, 2020, 1:10 PM UTC (108da3d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/286/200cff0...108da3d.html" title="Last updated on May 18, 2020, 1:10 PM UTC (108da3d)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -652,6 +652,10 @@ A <a>DID resolver</a> is a software and/or hardware component that takes a
 DID document</a> (and associated metadata) as output in a process
 called <a>DID resolution</a>.
       </p>
+    </section>
+
+    <section>
+      <h2>DID Dereferencers</h2>
       <p>
 The inputs and outputs of the <a>DID resolution</a> process are defined in
 <a href="#did-resolution"></a>. Additional considerations for implementing a
--- a/index.html
+++ b/index.html
@@ -739,6 +739,21 @@ hardware and conforms to this specification if it consumes
 when consuming non-conforming <a>DIDs</a> or <a>DID Documents</a>.
     </p>
 
+    <p>
+A <dfn>conforming DID resolver</dfn> that is conformant with this specification
+MUST be capable of performing the <a>DID resolution</a> process, as described in
+<a href="#did-resolution"></a> for at least one <a>DID method</a> and MUST
+return a <a>conforming DID document</a> in at least one conformant
+representation.
+    </p>
+
+    <p>
+A <dfn>conforming DID URL dereferencer</dfn> that is conformant with this
+specification MUST implement the <a>DID URL dereferencing</a> process, as
+described in <a href="#did-url-dereferencing"></a> for at least one conformant
+representation.
+          </p>
+
   </section>
 
   <section>
@@ -2694,13 +2709,6 @@ functions is out of scope for this specification, but some considerations for
 implementors are available in [[?DID-RESOLUTION]].
     </p>
 
-    <p>
-All conformant <a>DID resolvers</a> MUST implement the <a>DID resolution</a> function
-for at least one <a>DID method</a> and be able to return a <a>DID document</a> in
-at least one conformant representation. All conformant <a>DID URL dereferencers</a>
-MUST implement dereferencing for at least one conformant representation type.
-    </p>
-
     <section>
         <h2>
 DID Resolution
@@ -2738,7 +2746,6 @@ defines the following keys and values:
                     representation is supported and available.</dd>
             </dl>
         </p>
-
 
         <p>
 The <a>DID resolver</a> executes the "Read" operation of the <a>DID method</a> as described in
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -2694,12 +2711,196 @@ <h1>
...
+The <a>DID resolution</a> process takes a <a>DID</a> and resolution options as
+input, and produces a <a>DID document</a> and resolution metadata as output. The
+process utilizes the "Read" operation of the applicable <a>DID method</a>, as
+described in <a href="#read-verify"></a>. The <a>DID method</a>-specific details
+of how this process is accomplished is outside the scope of this specification.
+        </p>
+
+        <section>
+          <h3>
+Generic DID Resolution Algorithm
```
tplooker, line 2725: Is there a particular reason why we need the term "Generic" here, would "DID Resolution Algorithm" suffice?

index.html
```diff
@@ -2694,12 +2711,196 @@ <h1>
...
+          </h3>
+
+          <p>
+The generic DID resolution algorithm takes a <a>conforming DID</a>
```
tplooker, line 2729: Same comment w.r.t the above, about the usage of "Generic"

index.html
talltree: @tplooker Actually, I am strong proponent of constantly reinforcing that the DID spec defines **generic** DID syntax and **generic** DID resolution because that makes it clear that it is up to DID method specifications to define **specific** DID syntax (their own method name and method-specific identifier) and **specific** DID resolution operations (how their specific DID method performs the CRUD operations). @rhiaro and I are trying to use the same "generic/specific" terminology in our rewrite of the frontmatter sections (PRs coming shortly).

index.html
talltree: @tplooker See my reply above.

peacekeeper:
Hmm I understand the goal is to have more testable wording, but this PR introduces an actual algorithm, when I thought the idea was to only define the contract/interface? Note that we already have such an algorithm in [[?DID-RESOLUTION]]: https://w3c-ccg.github.io/did-resolution/#resolving

index.html
```diff
@@ -2694,12 +2711,196 @@ <h1>
...
+            </li>
+            <li>
+If successful, store the resulting <a>DID document</a>
+(<em>didDocument</em>), the DID Document Metadata
+(<em>didDocumentMetadata</em>), and the resolution metadata
+(<em>resolutionMetadata</em>).
+            </li>
+            <li>
+Ensure that <em>didDocument</em> is a <a>conforming DID document</a>. If the
+<em>didDocument</em> does not conform, an <code>invalid did document</code>
```
kdenhartog, line 2772: Can we talk about what a "conforming" did document would be? For example, would `did:example:123 => { "@id":"did:example:456" }` be consider valid even though the returned did document is valid, but doesn't match the input identifier?

msporny: > Can we talk about what a "conforming" did document would be or would that fall out of scope? 

The specification defines what a "conforming DID Document" is... so yes, it's in scope.

> For example, would did:example:123 => { "@id":"did:example:456" } be consider valid even though the returned did document is valid, but doesn't match the input identifier?

Per the current spec, it's invalid because of `@id` -- we don't allow that use because it goes against what the JSON-only folks want.

That the id doesn't map to the DID Document is valid as there is no normative requirement in the spec that states that the resulting DID Document MUST contain an `id` that matches the DID that was sent as the input of the DID Resolution process -- a DID Method can choose to give you back something totally different.

peacekeeper:
@jricher I mostly agree with your points but am not sure about the first one. My understanding has also been that the resolve() function includes certain (method-independent) parsing and validation steps.

jricher:
@peacekeeper Parsing and validation of the input DID and input options, yes. These need to be method-independent, for the sake of having a common function. But not parsing and validation of the output. It should generate valid output, but it shouldn't *parse* the document or output metadata. Otherwise, we can't define a method-independent way of describing the results of the function, and it's not meaningful.

msporny:
Closing since the group is now converging on attempting to get PRs #295, #296, #297, #298, #299, and #300 into the spec.

PR #287, opened 2020-05-18
Trivial nit wrt a reference

philarcher:
The reference to DID-RESOLUTION in line 1006 was only in single square brackets, not double. So this PR just doubles up those [] characters.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/287.html" title="Last updated on May 18, 2020, 8:08 AM UTC (ea6dc25)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/287/9a7cad0...ea6dc25.html" title="Last updated on May 18, 2020, 8:08 AM UTC (ea6dc25)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1003,7 +1003,7 @@ time. Note that this parameter might not be supported by all <a>DID methods</a>.
 
         <p>
 Additional considerations for processing these parameters are discussed in
-[?DID-RESOLUTION].
+[[?DID-RESOLUTION]].
         </p>
 
         <p>


peacekeeper:
Actually now that I look at this, it may have been better to merge it into master instead of this branch; but we should be able to fix this later.

peacekeeper:
Actually now that I look at this again, it may have been better to merge it into master instead of this branch; but we should be able to fix this later.

PR #295, opened 2020-05-27
Define simple type-less resolution function

jricher:
This adapts the DID Resolution functional definition defined by #253 into a single type-less function with no additional requirements.

***
<a href="https://pr-preview.s3.amazonaws.com/jricher/did-core/pull/295.html" title="Last updated on May 27, 2020, 9:27 PM UTC (8515500)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/295/200cff0...jricher:8515500.html" title="Last updated on May 27, 2020, 9:27 PM UTC (8515500)">Diff</a>

PR #296, opened 2020-05-27
Define resolution function with data types

jricher:
This adapts the DID Resolution functional definition defined by #253 into a single typed function with requirements for the function signature and implementation conformance. 

This builds on #295.

***
<a href="https://pr-preview.s3.amazonaws.com/jricher/did-core/pull/296.html" title="Last updated on Jun 9, 2020, 4:23 PM UTC (38c9969)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/296/200cff0...jricher:38c9969.html" title="Last updated on Jun 9, 2020, 4:23 PM UTC (38c9969)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2684,9 +2684,9 @@ Resolution
     </h1>
 
     <p>
-This section defines the inputs and outputs of the <a>DID resolution</a> and
-<a>DID URL dereferencing</a> processes. The exact implementation of these
-functions is out of scope for this specification, but some considerations for
+This section defines the inputs and outputs of <a>DID resolution</a> and
+<a>DID URL dereferencing</a>. These functions are defined in an abstract way.
+Their exact implementation is out of scope for this specification, but some considerations for
 implementors are discussed in [[?DID-RESOLUTION]].
     </p>
 
@@ -2704,7 +2704,7 @@ DID Resolution
 The <a>DID resolution</a> function resolves a <a>DID</a> into a <a>DID document</a>
 by using the "Read" operation of the applicable <a>DID method</a>. (See <a href="#read-verify"></a>.)
 The details of how this process is accomplished are outside the scope of this
-specification, but all conformant implementations MUST implement a function in the following form:
+specification, but all conformant implementations MUST implement a function which has the following abstract form:
         </p>
 
         <p><code>
--- a/index.html
+++ b/index.html
@@ -2767,6 +2767,14 @@ the <a>DID document</a> changes, as it represents data about the <a>DID document
 If the resolution is unsuccessful, this output MUST be an empty <a href="metadata-structure">metadata structure</a>.
             </dd>
         </dl>
+
+        <p>
+<a>DID resolver</a> implementations MUST NOT alter the signature of this function in any way. <a>DID resolver</a>
+implementations MAY map the <code>resolve</code> function to a method-specific internal function to perform the
+actual <a>DID resolution</a> process. <a>DID resolver</a> implementations <a>DID resolver</a> implementations MAY implement 
+and expose additional functions with different signatures in addition to the <code>resolve</code> function specified here.
+        </p>
+
     </section>
 
     <section>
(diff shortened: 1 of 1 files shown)


dlongley:
A valid DID resolver should have to implement "abstractResolve(did)". This abstract function can work across N different DID methods (N different DID Read operations). Additionally, a DID resolver may implement "concreteResolve(did)" which *wraps* "abstractResolve(did)" and outputs a DID Document (and whatever else) in a *concrete* representation.

I think if we take this approach we can fix the conflation with the current "resolve" function mixing both abstract and concrete concepts (it needs to be split apart -- and it is insufficient to point at a particular DID Read operation as the only abstract part of this process).

index.html
```diff
@@ -2687,13 +2687,94 @@ <h1>
 This section defines the inputs and outputs of the <a>DID resolution</a> and
 <a>DID URL dereferencing</a> processes. The exact implementation of these
 functions is out of scope for this specification, but some considerations for
```
peacekeeper, line 2689:
```suggestion
This section defines the inputs and outputs of <a>DID resolution</a> and
<a>DID URL dereferencing</a>. These functions are defined in an abstract way.
Their exact implementation is out of scope for this specification, but some considerations for
```

index.html
```diff
@@ -2687,13 +2687,94 @@ <h1>
...
 
     <section>
         <h2>
 DID Resolution
         </h2>
+        <p>
+The <a>DID resolution</a> function resolves a <a>DID</a> into a <a>DID document</a>
+by using the "Read" operation of the applicable <a>DID method</a>. (See <a href="#read-verify"></a>.)
+The details of how this process is accomplished are outside the scope of this
+specification, but all conformant implementations MUST implement a function in the following form:
```
peacekeeper, line 2707:
```suggestion
specification, but all conformant implementations MUST implement a function which has the following abstract form:
```

index.html
```diff
@@ -2687,13 +2687,94 @@ <h1>
...
+        <p>
+The input variables of this function MUST be as follows:
+        </p>
+
+        <dl>
+            <dt>
+did
+            </dt>
+            <dd>
+A conformant <a>DID</a> as a single string. This is the <a>DID</a> to resolve.
```
kdenhartog, line 2724: Do we want to specify whether it MUST be a DID URI or a DID URL here? In general, I've operated under the assumption resolution accepts only DID URIs and dereferencing accepts DID URLs.

peacekeeper: I don't think we're defining "DID URI" anywhere? Resolution takes a DID, dereferencing accepts a DID URL.

msporny:
@jricher could you please process the comments from @peacekeeper so we can get this PR merged?

peacekeeper:
@dlongley perhaps what you call `abstractResolve()` is really `resolve()`, and what you call `concreteResolve()` is actually `dereference()` ? If you try reading your own comment above with those terminology replacements, does it make sense to you?

This way, `resolve()` would return an abstract DID document (not a byte stream), and `dereference()` would return a concrete representation, which to me feels consistent with how URI dereferencing works in general.

index.html
jricher: Resolution always takes a DID.

kdenhartog: Just to clarify terminology first, when you're saying DID, do you mean just `did:example:123`? That's what I meant by DID URI, so I think we'd be in alignment then. However, I found a case that would conflict with my original assumption. The case where we're referring to a specific version of a did document with the version query parameter could use some clarity for this.

For example, if I used `did:example:123?version=1` would I need to parse `version=1` from the identifier and pass that via the options then or could I pass the full DID URL with the version query parameter to the resolution function?

peacekeeper: @kdenhartog 

> would I need to parse `version=1` from the identifier and pass that via the options then or could I pass the full DID URL with the version query parameter to the resolution function?

You can only pass a DID to `resolve()`. But you can pass a DID URL to `dereference()`. In that latter case, the DID parameters are parsed from the DID URL and then passed as `did-resolution-input-metadata` (aka "options") to `resolve()`, exactly like you said.

Example:

`dereference(did-url="did:example:123?version=1", did-dereference-input-metadata=null)`

will internally call

`resolve(did="did:example:123", did-resolution-input-metadata=[version="1"])`

See the first step in the [DID URL Dereferencing algorithm section of the DID Resolution spec](https://w3c-ccg.github.io/did-resolution/#dereferencing-algorithm), which says this:

_All generic DID parameters as well as method-specific DID parameters of the input DID URL MUST be passed as input metadata properties to the DID Resolution algorithm._

PR #297, opened 2020-05-27
Define resolution function with data types and property values

jricher:
This adapts the DID Resolution functional definition defined by #253 into a single typed function with requirements for the function signature and implementation conformance. Property values for input and output metadata structures are defined here, with pointers to the DID registry for management of properties. Requirements for when and how to return each value are also added.

This builds on #295 and #296.

***
<a href="https://pr-preview.s3.amazonaws.com/jricher/did-core/pull/297.html" title="Last updated on May 27, 2020, 9:28 PM UTC (db3383d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/297/200cff0...jricher:db3383d.html" title="Last updated on May 27, 2020, 9:28 PM UTC (db3383d)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2767,6 +2767,14 @@ the <a>DID document</a> changes, as it represents data about the <a>DID document
 If the resolution is unsuccessful, this output MUST be an empty <a href="metadata-structure">metadata structure</a>.
             </dd>
         </dl>
+
+        <p>
+<a>DID resolver</a> implementations MUST NOT alter the signature of this function in any way. <a>DID resolver</a>
+implementations MAY map the <code>resolve</code> function to a method-specific internal function to perform the
+actual <a>DID resolution</a> process. <a>DID resolver</a> implementations <a>DID resolver</a> implementations MAY implement 
+and expose additional functions with different signatures in addition to the <code>resolve</code> function specified here.
+        </p>
+
     </section>
 
     <section>
--- a/index.html
+++ b/index.html
@@ -2711,6 +2711,62 @@ specification, but all conformant implementations MUST implement a function in t
 resolve ( did, did-resolution-input-metadata ) <br>
 &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;-&gt; ( did-resolution-metadata, did-document-stream, did-document-metadata )
         </code></p>
+
+        <p>
+The input variables of this function MUST be as follows:
+        </p>
+
+        <dl>
+            <dt>
+did
+            </dt>
+            <dd>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -2687,13 +2687,191 @@ <h1>
...
+        <p>
+The <a>DID resolution</a> function resolves a <a>DID</a> into a <a>DID document</a>
+by using the "Read" operation of the applicable <a>DID method</a>. (See <a href="#read-verify"></a>.)
+The details of how this process is accomplished are outside the scope of this
+specification, but all conformant implementations MUST implement a function in the following form:
+        </p>
+
+        <p><code>
+resolve ( did, did-resolution-input-metadata ) <br>
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;-&gt; ( did-resolution-metadata, did-document-stream, did-document-metadata )
```
davidlehn, line 2712: Is this syntax for these functions described somewhere?  Hard to tell looking at this if this can be an async function (in languages that do that).  A JavasScript implementation would probably return a Promise here.  Is the output a tuple or map with named properties?  Seems kind of a fuzzy description for a "MUST".  In the JSON-LD API spec we used WebIDL for these API descriptions.  Has some rough edges but seemed to work good enough and has some respec support.  The [LoadDocumentCallback](https://www.w3.org/TR/json-ld11-api/#remote-document-and-context-retrieval) API has a similar pattern to this.

jricher: Those details depend on your language. It's pretty clear in any implementation what "call a function" and "return a value" mean, and it's different for each one. 

Here, we're not specifying if it's synchronous or not, so a promise-based resolution would be fine. We're also not specifying how multiple outputs are returned, so a tuple, or an encapsulating object, or a bunch of pointers will all be fine.

davidlehn: I think it's confusing to use language like "all conformant implementations MUST implement a function in the following form".  In a spec I read this as something strict.  But here it involves guessing what the ad-hoc function call description means, and understanding it can be sync/async, use tuples, objects, or pointers (for parameters and results maybe?), and I'm guessing the actual names are not required either since many languages can't handle `-` in identifiers.  What form is it that must be followed?

msporny:
PR #331 has been merged. @jricher, can we close this PR now?

peacekeeper:
@jricher do you agree we can close this now that https://github.com/w3c/did-core/pull/331 has been merged (since that PR also includes the new incremental commit from this PR here)?

PR #298, opened 2020-05-27
Define resolution function with data types, property values, and simple metadata structures

jricher:
This adapts the DID Resolution functional definition defined by #253 into a single typed function with requirements for the function signature and implementation conformance. Property values for input and output metadata structures are defined here, with pointers to the DID registry for management of properties. Requirements for when and how to return each value are also added. 

Simple metadata structures are defined here, without detail for conformance or transmission.

This builds on #295, #296, and #297.

***

### :boom: Error: 500 Internal Server Error :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on May 27, 2020, 9:28 PM UTC)_.

<details>
<summary>More</summary>

PR Preview relies on a number of web services to run. There seems to be an issue with the following one:

:rotating_light: [Spec Generator](https://www.w3.org/2015/labs/) - Spec Generator is the web service used to build specs that rely on ReSpec.

:link: [Related URL](https://labs.w3.org/spec-generator/?type=respec&url=https%3A%2F%2Frawcdn.githack.com%2Fjricher%2Fdid-core%2F1b2f1523e079a2c09f7dad531aa87b132c6092cb%2Findex.html%3FisPreview%3Dtrue)

```

😭  Sorry, there was an error generating the HTML. Please report this issue!
[36mSpecification: https://rawcdn.githack.com/jricher/did-core/1b2f1523e079a2c09f7dad531aa87b132c6092cb/index.html?isPreview=true&publishDate=2020-05-27[39m
[36mReSpec version: 25.6.0[39m
[36mFile a bug: https://github.com/w3c/respec/[39m
[36mError: Error: Evaluation failed: Timeout: document.respecIsReady didn't resolve in 19587ms.[39m
[36m    at ExecutionContext._evaluateInternal (/u/spec-generator/node_modules/puppeteer/lib/ExecutionContext.js:122:13)[39m
[36m    at runMicrotasks (<anonymous>)[39m
[36m    at processTicksAndRejections (internal/process/task_queues.js:97:5)[39m
[36m    at async ExecutionContext.evaluate (/u/spec-generator/node_modules/puppeteer/lib/ExecutionContext.js:48:12)[39m
[36m    at async generateHTML (/u/spec-generator/node_modules/respec/tools/respecDocWriter.js:128:12)[39m
[36m    at async fetchAndWrite (/u/spec-generator/node_modules/respec/tools/respecDocWriter.js:95:18)[39m
[...]

PR #299, opened 2020-05-27
Define resolution function with data types, property values, and full metadata structures

jricher:
This adapts the DID Resolution functional definition defined by #253 into a single typed function with requirements for the function signature and implementation conformance. Property values for input and output metadata structures are defined here, with pointers to the DID registry for management of properties. Requirements for when and how to return each value are also added. 

Simple metadata structures are defined here, including requirements for conformance, transformation, and transmission.

This builds on #295, #296, #297, #298, and #300.

***
<a href="https://pr-preview.s3.amazonaws.com/jricher/did-core/pull/299.html" title="Last updated on May 28, 2020, 2:48 PM UTC (72acb94)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/299/200cff0...jricher:72acb94.html" title="Last updated on May 28, 2020, 2:48 PM UTC (72acb94)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2909,6 +2909,29 @@ Property values MAY be repeated under different property names within a given se
 Property definitions MAY impose additional constraints on the property value, so long as
 those constraints allow for unambiguous representation as a single string.
         </p>
+        
+        <p>
+Properties within a <a href="metadata-structure">metadata structure</a> MAY be represented by an
+implementation in a variety of ways in an underlying protocol or system.
+For example, an implementation could represent a <a href="metadata-structure">metadata structure</a> with
+a set of HTTP-style headers, a JSON object, or an in-memory data map.
+Implementations MUST NOT require callers of functions to know the details of the underlying representation
+of the <a href="metadata-structure">metadata structure</a> in order to call the function.
+        </p>
+
+        <p>
+Properties within a <a href="metadata-structure">metadata structure</a> MAY be transformed by an
+implementation in order to facilitate their transmission in an underlying protocol or implementation.
+Any transformations MUST be bidirectional and lossless. 
+Implementations MUST NOT expose these transformations to callers of the <a>DID resolution</a> function and
+MUST NOT require callers of the <a>DID resolution</a> function to know about the transformations in order to call the function.
+For example, an implementation could prepend the string <code>DID-Document-</code> to the names of all <a href="did-document-metadata-properties"></a>
+to differentiate them in its underlying protocol, but the caller of the <a>DID resolution</a> function would refer to them only by their defined
+property names without the prefix.
+The details of such transformations are method-specific and are outside the scope of this specification,
+but considerations for implementors are discussed in [[?DID-RESOLUTION]].
+        </p>
+
     </section>
 
   </section>
--- a/index.html
+++ b/index.html
@@ -2884,6 +2884,16 @@ DID URL Dereferencing
         <h2>
 Metadata Structure
         </h2>
+
+        <p>
+A <a href="metadata-structure">metadata structure</a> is a common and interoperable data structure used to
+communicate information alongside a <a>DID</a> and <a>DID document</a>, and is specifically
+used for the <a>DID resolution</a> function definition in this specification.
+Use of a <a href="metadata-structure">metadata structure</a> is intended to compliment the information available
+within a <a>DID</a> or <a>DID document</a>, and therefore is best suited to carrying
+metadata as an input or output to the function process.
+        </p>
+        
     </section>
 
   </section>
--- a/index.html
+++ b/index.html
@@ -2767,6 +2767,14 @@ the <a>DID document</a> changes, as it represents data about the <a>DID document
 If the resolution is unsuccessful, this output MUST be an empty <a href="metadata-structure">metadata structure</a>.
             </dd>
         </dl>
+
+        <p>
+<a>DID resolver</a> implementations MUST NOT alter the signature of this function in any way. <a>DID resolver</a>
+implementations MAY map the <code>resolve</code> function to a method-specific internal function to perform the
+actual <a>DID resolution</a> process. <a>DID resolver</a> implementations <a>DID resolver</a> implementations MAY implement 
+and expose additional functions with different signatures in addition to the <code>resolve</code> function specified here.
+        </p>
+
     </section>
 
     <section>
--- a/index.html
+++ b/index.html
@@ -2711,6 +2711,62 @@ specification, but all conformant implementations MUST implement a function in t
 resolve ( did, did-resolution-input-metadata ) <br>
 &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;-&gt; ( did-resolution-metadata, did-document-stream, did-document-metadata )
         </code></p>
+
+        <p>
+The input variables of this function MUST be as follows:
+        </p>
+
+        <dl>
+            <dt>
+did
+            </dt>
+            <dd>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -2706,6 +2884,54 @@ <h2>
...
+of the <a href="metadata-structure">metadata structure</a> in order to call the function.
+        </p>
+
+        <p>
+Properties within a <a href="metadata-structure">metadata structure</a> MAY be transformed by an
+implementation in order to facilitate their transmission in an underlying protocol or implementation.
+Any transformations MUST be bidirectional and lossless. 
+Implementations MUST NOT expose these transformations to callers of the <a>DID resolution</a> function and
+MUST NOT require callers of the <a>DID resolution</a> function to know about the transformations in order to call the function.
+For example, an implementation could prepend the string <code>DID-Document-</code> to the names of all <a href="did-document-metadata-properties"></a>
```
OR13, line 2928: I'm having trouble groking this section, and wording like:

"MUST NOT require callers of the <a>DID resolution</a> function to know about the transformations in order to call the function."

I feel like this language is not necessary, so long as the response interface is defined... callers are required to understand the interface... right?

As a DID Method author, I define support for JSON / CBOR / JSON-LD... resolve returns a representation (singular)... we believe that the representation can be transformed in a lossless bi-directional manner if we believe that the contract interface and registry are all that is needed to support that....This feels like we are now saying that DID Method implementers are responsible for ensuring that resolve supports all 3... instead of relying on the registry and the WG to make that transparent.... using post resolution middleware... in other words, all DID Methods are now required to produce all representations?... 

I guess what I am trying to say is: "Any transformations MUST be bidirectional and lossless. "... I would like to see a proof that its possible for at least JSON /JSON-LD/CBOR, before I see normative text saying its a requirement... I'm not sure I believe its possible.... and if its not, then we will end with 3 non interoperable DID Representations... Yikes!

jricher: So this is the metadata, not the document itself. The metadata is not in JSON/CBOR/etc. The metadata is a map of strings: so a Javascript hash object, a python dict, a java Map<String, String>, etc. 

The transformation in question here is to and from that data structure into whatever underlying format there needs to be. That's one of the reasons that I think it's important for the metadata structure to be a LIMITED data structure in terms of Map[String -> String] and not objects, arrays, numbers, or anything else. This is representable in any possible underlying system in a deterministic way. 

The reversible transforms basically just means that if your protocol needs to use "DID-Resolution-Content-Type" to represent a field, you can't have someone call you with "DID-Resolution-Content-Type", you have to let them call you with "content-type", and you need to return "content-type" in the resulting map. This needs to be a strict requirement, but I'm struggling how to make it normatively enforceable. Thus the "MUST NOT expose" language above as an attempt to get to that.

I think a number of people are getting lost on this point, especially that the metadata is intended to be completely independent of the underlying representation. Is there a way to rephrase this to make it more clear what's going on here?

llorllale: This sections feels to me like it's delving a little bit too deep into the internals of the `resolve` function. It seems out of scope of did-core.

> if your protocol needs to use "DID-Resolution-Content-Type" to represent a field, you can't have someone call you with "DID-Resolution-Content-Type", you have to let them call you with "content-type", and you need to return "content-type" in the resulting map.

Understood. However, I still question its pertinence to did-core. This belongs in a transport-specific did-resolution spec.

jricher: I can see how it can be read that way, but that's not the intent. The intent isn't to describe how any mapping can happen, but instead put limits on what a mapping can do. The real requirement here is that if a property is named `foo-bar`, then I should be able to call this function with a property named `foo-bar` and get the expected behavior (defined by the property) regardless of how the implementation represents `foo-bar` under the hood. I'll put in an alternative phrasing for this requirement that might capture it better.

index.html
```diff
@@ -2706,6 +2884,54 @@ <h2>
...
+Implementations MUST NOT require callers of functions to know the details of the underlying representation
+of the <a href="metadata-structure">metadata structure</a> in order to call the function.
+        </p>
+
+        <p>
+Properties within a <a href="metadata-structure">metadata structure</a> MAY be transformed by an
+implementation in order to facilitate their transmission in an underlying protocol or implementation.
+Any transformations MUST be bidirectional and lossless. 
+Implementations MUST NOT expose these transformations to callers of the <a>DID resolution</a> function and
+MUST NOT require callers of the <a>DID resolution</a> function to know about the transformations in order to call the function.
```
jricher, line 2927:
```suggestion
Implementations MUST accept metadata properties by their registered names as inputs and MUST produce metadata properties by their registered names as outputs, regardless of any internal transformation that could occur within the implementation.
```

jricher:
Since there were a lot of comments on the transformation section but not the requirements section, I've made another pull request that includes the requirements but not the transformation language in case that's a better place to make the cut: #300

index.html
```diff
@@ -2687,13 +2687,191 @@ <h1>
...
+        <p>
+The <a>DID resolution</a> function resolves a <a>DID</a> into a <a>DID document</a>
+by using the "Read" operation of the applicable <a>DID method</a>. (See <a href="#read-verify"></a>.)
+The details of how this process is accomplished are outside the scope of this
+specification, but all conformant implementations MUST implement a function in the following form:
+        </p>
+
+        <p><code>
+resolve ( did, did-resolution-input-metadata ) <br>
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;-&gt; ( did-resolution-metadata, did-document-stream, did-document-metadata )
```
dmitrizagidulin, line 2712: What format is the `did-resolution-metadata` argument? (I assume the `did-document-stream` is in the format that the metadata specifies.)

jricher: The `did-resolution-metadata` is a map of strings. It's not in any particular serialized format as it's not expected to be parsed outside of this function, just read/used. This could be a javascript hash, a python dict, or a java Map<String, String>, or any number of other hash-table-style data structures that can store strings under unique string keys. The goal was to be a simple enough data structure that we wouldn't need to agree on a serialization format -- that's why it doesn't allow arrays, objects, numbers, or other things that would need serialization/parsing rules.

jricher: That would be fine with me. I would suggest doing that kind of work as a separate pull request as it's more editorial at that point.

msporny:
PR #331 has been merged. @jricher, is there any part of this PR that needs to be reworked or applied to another metadata PR or can we close this PR now?

index.html
```diff
@@ -2706,6 +2884,54 @@ <h2>
...
+        </p>
+        
+        <p>
+A <a href="metadata-structure">metadata structure</a> consists of a set of properties.
+Each property has a single name and a single value.
+The name of the property MUST be a single string. 
+Property names MUST be unique within a given set. 
+Each property name MUST map unambiguously to a single value within the set.
+Property names MUST be compared by the exact byte value.
+The value of a property MUST be a single string.
```
selfissued, line 2904: Properties must be able to be any JSON type - including arrays, booleans, objects, and yes strings.

OAuth metadata contains all of these.  There's no reason to believe that DID metadata won't.

peacekeeper:
I feel like this PR has been superseded by https://github.com/w3c/did-core/pull/331 and https://github.com/w3c/did-core/pull/347. I think we should close this. If there is anything in this PR that we want to get into the spec (e.g. language around lossless transformation of metadata in underlying protocols), then this can still be added on top of the current spec in a separate PR.  @jricher do you agree?

msporny:
Closing this in 48 hours if we don't hear back from the PR author.

msporny:
No response from original PR author, closing.

PR #305, opened 2020-05-31
Remove proof as it is specific to Linked Data Proofs. Fixes #293

### LINKED ISSUE #293: Remove `proof`
`proof` should not be in DID Core as it is a way to protect content not used by all DID methods. Propose removing it altogether, understanding that it will still be available for those that want to use it via the DID Spec Registries.

rhiaro:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/305.html" title="Last updated on May 31, 2020, 2:57 PM UTC (d846737)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/305/490cccb...d846737.html" title="Last updated on May 31, 2020, 2:57 PM UTC (d846737)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1967,56 +1967,6 @@ property, as outlined in Section <a href="#created"></a>.
       </pre>
     </section>
 
-    <section>
-      <h2>Proof</h2>
-
-      <p>
-A <code>proof</code> on a <a>DID document</a> is cryptographic proof of the
-integrity of the <a>DID document</a> according to either the:
-      </p>
-
-      <ul>
-        <li>
-<a>DID subject</a>, as defined in Section <a href="#did-subject"></a>
-        </li>
-
-        <li>
-<a>DID controller</a>, as defined in Section
-<a href="#authorization-and-delegation"></a>, if present.
-        </li>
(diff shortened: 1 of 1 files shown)


rhiaro:
If there are objections to removing `proof` altogether, suggest we mark it as at-risk.

peacekeeper:
I approve this if @dmitrizagidulin approves it too :)

msporny:
The reasoning for removing `proof` is not that it's not useful. The use of `proof` to protect the contents of the DID Document is very useful... but it's a part of the Linked Data Security stack. This property comes from a time when we expected DID Documents to be signed on all ledgers, but that's not how things have played out. All that said, it's still possible that DID Documents /can/ be signed on ledger using the Linked Data Security stack... but the DID Core spec should be silent on how DID Documents are protected on-ledger. That is a DID Method implementation specific detail. Therefore, we should remove `proof` from the core spec and let it live in the Linked Data Security specs.

msporny:
Normative, multiple positive reviews, no objections, merging.

PR #315, opened 2020-06-14
Remove interop registries appendix.

msporny:
Remove the interop registries appendix as we refer to the registries throughout the document now and the appendix duplicates content earlier in the specification as well as in the section on interop.


msporny:

PR #316, opened 2020-06-14
Remove largely ficticious real world example.

msporny:
The real world example that is in an appendix in the specification is largely fictitious and should be removed. At present, we have a number of other examples that have been added to the specification that do a better, more precise, job of demonstrating what various properties look like when they're used. We can add another more complete example in the future if the group feels we need it, but at present, we probably don't need one.


msporny:

PR #320, opened 2020-06-15
Makes use of property/name/value consistent

rhiaro:
per #204.

(Also fixes spacing and markup in CBOR section.)

### DIFF
--- a/index.html
+++ b/index.html
@@ -1194,7 +1194,8 @@ representations, and a set of requirements for extensibility.
     <section>
       <h2>Definition</h2>
       <p>
-A <a>DID document</a> consists of a collection of property-value pairs. The
+A <a>DID document</a> consists of a collection of properties, which are
+name-value pairs (ie. a property name, and a property value). The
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1368,8 +1371,8 @@ <h2>Verification Methods</h2>
       </p>
 
       <p>
-In order to maximize global interoperability, the verification method properties
-SHOULD be registered in the [[DID-SPEC-REGISTRIES]].
+In order to maximize global interoperability, the verification method property
+names SHOULD be registered in the [[DID-SPEC-REGISTRIES]].
```
msporny, line 1375:
```suggestion
In order to maximize global interoperability, the verification method 
SHOULD be registered in the [[DID-SPEC-REGISTRIES]].
```

msporny: Technically, we're registering more than just the property name, so keeping this less specific for now.

msporny:
Again, couldn't rebase+merge because of multi-branch merges. Had to squash+merge.

PR #321, opened 2020-06-16
Put all term dfns in terms.html

### LINKED ISSUE #311: Move term definitions into the term document
The current document includes <dfn>term</dfn> definition in the body of the document. 

The terms.html file refers (externally) to those terms, which happens to work in the DID Core respec, but kills the term filtering in the use case document.

All defined terms need to be moved to the terms.html file.

rhiaro:
Also replaces references to 'this specification' in the terms file with 'the DID Core specification' (and a link), for when the terms file is included elsewhere (ie. use cases).

Fixes #311.

### DIFF
--- a/index.html
+++ b/index.html
@@ -651,13 +651,13 @@ implementations of a given DID method must be interoperable for that method).
       </p>
 
      <p>
-A <a>conforming DID</a> is any concrete expression of the rules specified in
+A <dfn>conforming DID</dfn> is any concrete expression of the rules specified in
 Section <a href="#identifier"></a> and MUST comply with relevant normative
 statements in that section.
      </p>
 
       <p>
-A <a>conforming DID document</a> is any concrete expression of the data
+A <dfn>conforming DID document</dfn> is any concrete expression of the data
 model described in this specification and MUST comply with the relevant
 normative statements in Sections <a href="#data-model"></a> and  <a
 href="#core-properties"></a>. A serialization format for the conforming document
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -667,24 +667,22 @@ <h2>
...
       <p>
-A <dfn>consumer</dfn> is any algorithm realized as software and/or hardware and
-conforms to this specification if it consumes <a>conforming DID</a>s or
-<a>conforming DID Document</a>s. A <a>consumer</a> that is conformant with this
-specification MUST produce errors when consuming non-conforming <a>DIDs</a> or
-<a>DID Documents</a>.
+A <a>conforming consumer</a> is any algorithm realized as software and/or
+hardware and conforms to this specification if it consumes <a>conforming
+DID</a>s or <a>conforming DID document</a>s. A <a>conforming consumer</a> MUST
+produce errors when consuming non-conforming <a>DIDs</a> or <a>DID documents</a>.
```
msporny, line 685: Hmm, don't think I agree with moving conformance statements to terminology. I do understand that "conforming DID document" is a term we use throughout the specification, but it has more to do with conformance than it does with terminology. To put it another way, I think it's fine for certain things to only appear in the sections that define them if no other section points to them. The conformance statements fall into that category, IMHO. 

To put it another way, do we think that someone in some other group is going to pull in the "conforming DID document" definition into their spec? I hope not... I'm not sure it adds to another spec to refer to stuff as a "conforming this" or a "conforming that".

index.html
```diff
@@ -2039,7 +2037,7 @@ <h2>Updated</h2>
       </p>
 
       <dl>
-          <dt><dfn>updated</dfn></dt>
+          <dt>updated</dt>
```
msporny, line 2040: I'd like to understand the thought process behind removing the dfns, here. One good argument for removing them is that we don't cite them or link to them from elsewhere in the specification? We may want to discuss why not before removing them. It would be nice to link to these items if we mention them elsewhere in the specification... that was the intent long ago, and then I never got around to writing the ReSpec extension to do that like we do for the `<a>reference</a>` stuff.

terms.html
```diff
@@ -7,6 +7,13 @@
 
 <dl class="termlist">
 
+  <dt><dfn>authentication</dfn></dt>
+  <dd>
+Authentication is a mechanism by which an entity can prove it is the <a>DID
+subject</a>, using one or more of a set of verification methods (such as, but
+not limited to, public keys).
+  </dd>
+
```
msporny, line 16: We need to improve this definition, it's inaccurate... /cc @kdenhartog @talltree @dlongley

index.html
rhiaro: I agree. I moved them all because the definition for DID resolver refers to "conforming DID document", so when the terms.html is pulled into other specs (ie. the use cases), the internal ref breaks there. Probably easiest just to remove the link around "conforming DID document" in the DID resolver definition and put them back?

index.html
rhiaro: Right, they're not cited anywhere

index.html
rhiaro: Updated to do [that](https://github.com/w3c/did-core/pull/321#discussion_r443184308) ^

index.html
rhiaro: I also find it a bit odd to have definitions around property names. In my head I differentiate between definitions for terms/concepts and the specific property names, and I think it makes more sense to link to the section in which they are defined than to put them in as term definitions.

msporny:
Editorial, multiple positive reviews, changes requested and made, no objections, merging.

PR #322, opened 2020-06-17
Allow resolve() to return abstract DID document (option 1)

peacekeeper:
Building on https://github.com/w3c/did-core/pull/296 and https://github.com/w3c/did-core/pull/297, this is one of two proposals how `resolve()` can return a DID document in an abstract form.

- Proposal 1 https://github.com/w3c/did-core/pull/322 **(this one)**: `resolve()` always returns a DID document in an abstract form, and `dereference()` always returns a DID document or other content as a byte stream of a concrete representation.
- Proposal 2 https://github.com/w3c/did-core/pull/323: `resolve()` can return either a DID document in an abstract form or as a byte stream of a concrete representation, depending on whether the `accept` input metadata property is present.

### DIFF
--- a/index.html
+++ b/index.html
@@ -3013,7 +3013,7 @@ implementations MUST implement a function which has the following abstract form:
 
         <p><code>
 resolve ( did, did-resolution-input-metadata ) <br>
-&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;-&gt; ( did-resolution-metadata, did-document-stream, did-document-metadata )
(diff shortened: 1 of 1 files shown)


dlongley:
I prefer #323 -- I'm concerned here with introducing more things that others might find objectionable.

peacekeeper:
Replaced by https://github.com/w3c/did-core/pull/331 per discussion on the [2020-06-18 DID WG Topical Call on DID Resolution contracts](https://www.w3.org/2019/did-wg/Meetings/Minutes/2020-06-18-did-topic).

PR #323, opened 2020-06-17
Allow resolve() to return abstract DID document (option 2)

peacekeeper:
Building on https://github.com/w3c/did-core/pull/296 and https://github.com/w3c/did-core/pull/297, this is one of two proposals how `resolve()` can return a DID document in an abstract form.

- Proposal 1 https://github.com/w3c/did-core/pull/322: `resolve()` always returns a DID document in an abstract form, and `dereference()` always returns a DID document or other content as a byte stream of a concrete representation.
- Proposal 2 https://github.com/w3c/did-core/pull/323 **(this one)**: `resolve()` can return either a DID document in an abstract form or as a byte stream of a concrete representation, depending on whether the `accept` input metadata property is present.

### DIFF
--- a/index.html
+++ b/index.html
@@ -3034,7 +3034,9 @@ did-resolution-input-metadata
             <dd>
 A <a href="metadata-structure">metadata structure</a> consisting of input
 options to the <code>resolve</code> function in addition to the <code>did</code>
-itself. This input is REQUIRED, but the structure MAY be empty.
+itself.
+Properties defined by this specification are in <a href="#did-resolution-input-metadata-properties"></a>.
+This input is REQUIRED, but the structure MAY be empty.
             </dd>
         </dl>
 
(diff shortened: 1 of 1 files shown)


msporny:
This PR feels like it checks all the right boxes wrt. the discussion happening in the WG... feeling positive that this one might make it through. Thank you @peacekeeper!

index.html
```diff
@@ -3089,6 +3095,98 @@ <h2>
...
+This specification defines the following common properties.
+            </p>
+
+            <dl>
+                <dt>
+content-type
+                </dt>
+                <dd>
+The mime-type of the returned <code>did-document</code>.
+If the resolution is successful, and if the <code>did-document</code> is returned in the form of a byte stream, this property is REQUIRED.
```
kdenhartog, line 3136: I'm not sure I understand what byte stream means and it seems to be the important nuance of this discussion. The one thing I understand is that because we've got an abstract model we can't assume it will always be a JSON object because it could be CBOR as well or some other form and a byte stream is the least common denominator of them all. 

However, this language leaves me confused as to what I would need to do to practically implement around this. Could someone clarify which type in a programming language this is referring to? For example, in typescript is this mean I should expect a Uint8Array in and need to encode it using a CBOR to JSON converter if I want to move it around in memory as JSON? 

What if I expect the did document to be JSON(-LD)? Should I expect to utf8 encode the bytes first and then JSON.parse() the string I just got?

As the language currently stands in this I'm left a bit confused as to these types of implementation considerations, so I'd like some further clarification. With that in mind, I recognize some of this has to stay relatively vague to make it compatible with an abstract data model.

peacekeeper:
Replaced by https://github.com/w3c/did-core/pull/331 per discussion on the [2020-06-18 DID WG Topical Call on DID Resolution contracts](https://www.w3.org/2019/did-wg/Meetings/Minutes/2020-06-18-did-topic).

Issue #122, opened 2019-11-25
When is a DID subject not a DID controller (if ever)?


rhiaro:
By default, the DID subject is a controller of its own DID (this is not stated in the spec but I think everyone agrees on this).

When there is explicitly a `controller` property at the top level of the DID document, pointing to a controller other than the DID subject, does this *add to* a list of controllers which still includes the DID subject, or does it *override* such that the DID subject is no longer considered a controller?

If there is a case for both or the latter of these being an option, how could it ever be shown that the DID subject is *not* a controller of the DID?

TallTed:
I think there are cases for both scenarios, which could be covered by something like -- 

* When `controller` property is not present, the DID subject is the only controller of the DID. 
* When `controller` property is present, its value identifies all controllers of the DID. If the DID subject is not included therein, the DID subject is not a controller of the DID.

ewelton:
I think this is a great question for clarification, and I feel like I may have missed a large chunk of the historical conversation here so I'd like to offer up some ignorance in hopes that it be quickly put to rest.

It really feels like the winds of DIDs are now pointing *only* to a certain class of enabled sentient sapiens.

Once upon a time I thought it might be possible for my cats to have multiple DIDs - one for their medical records, one for their instragram accounts, and a separate one for Litter - like Tinder, but for cats.

I also once had an idea that I'd tag all the books on the banana pancake trail with dids, so they could be tracked via QR codes (kinda as a playground for rugged wearing pepper-and-ink or scantrust kinda QR tech - SICPA has some tech in play too)

Perhaps this is entirely dealt with by Guardianship and Ownership constructs - or that DIDs are now *only* about establishing communication with active software agents.  Is the idea of having a DID for a book, where the control of that DID moves along with the controller while allowing the book to be referenced (via DID) sensible, or is that an abuse of DIDs in the current thinking?

I'm asking as a point of clarification rather than advancing this specific question forward so that I can respond to the question itself with minimal lateral confusion.

jandrieu:
Responding to @rhiaro's opening:
> By default, the DID subject is a controller of its own DID 

Depending on your meaning, I would not say this is the default. It *is* the most common case that most of us care about, but one should not assume that the subject is the controller.

In discussion (although not necessarily in the Use Cases & Requirements document yet), the most common cases where the subject is not the controller are devices and dependents, or more commonly for both: IoT and guardianship situations.

[...]

PR #326, opened 2020-06-21
Relationship between `controller` property and DID controller

rhiaro:
Issue #122 has a lot of back-and-forth about the relationship between the `controller` property and the actual DID controller, aka the entity that can edit the DID document. 

The original issue was about whether we can assume the DID subject (aka an entity that can authenticate itself with verification material contained in the DID doc.. I think?) is default controller in the absence of a `controller` property. (Short answer to that being 'no'.)

I'm not sure there's consensus, so I've added a NOTE presenting one set of perspectives from that thread, which is basically that the `controller` property is indicative but ultimately who/what can or cannot edit a DID doc, and under what circumstances, is up to the DID method, `controller` property or no.

I'm not sure if there's normative language to be developed from this, happy to iterate.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1830,10 +1830,8 @@ malicious activity by an attacker. See Section
 
       <p>
 A DID document MAY include a <code>controller</code> property to indicate
-that there are <a>DID controller(s)</a> other than the <a>DID subject</a>.
-If so:
+the <a>DID controller(s)</a>. If so:
       </p>
-
       <dl>
           <dt><dfn>controller</dfn></dt>
           <dd>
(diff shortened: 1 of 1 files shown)


iherman:
See:

* [Preview](https://cdn.statically.io/gh/w3c/did-core/rhiaro-122-controller/index.html)
* [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/did-core/&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/did-core/rhiaro-122-controller/index.html)

PR #331, opened 2020-06-23
Support resolve() and resolveStream() functions

peacekeeper:
Replaces https://github.com/w3c/did-core/pull/322 and https://github.com/w3c/did-core/pull/323. Builds on https://github.com/w3c/did-core/pull/296 and https://github.com/w3c/did-core/pull/297.

During the [2020-06-18 DID WG Topical Call on DID Resolution contracts](https://www.w3.org/2019/did-wg/Meetings/Minutes/2020-06-18-did-topic) we had consensus that we should have two abstract functions for DID resolutions - one called `resolve` that returns a DID document as the abstract data model, and one called `resolveStream` that returns a DID document as a byte stream in one of the conformant representations.

After merging this, we can add a few remaining improvements that have been discussed, such as compatibility with the [Infra standard](https://infra.spec.whatwg.org/), and definition of metadata structures.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/331.html" title="Last updated on Jul 2, 2020, 9:45 PM UTC (2627f95)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/331/2110e3f...2627f95.html" title="Last updated on Jul 2, 2020, 9:45 PM UTC (2627f95)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -3034,7 +3034,9 @@ did-resolution-input-metadata
             <dd>
 A <a href="metadata-structure">metadata structure</a> consisting of input
 options to the <code>resolve</code> function in addition to the <code>did</code>
-itself. This input is REQUIRED, but the structure MAY be empty.
+itself.
+Properties defined by this specification are in <a href="#did-resolution-input-metadata-properties"></a>.
+This input is REQUIRED, but the structure MAY be empty.
             </dd>
         </dl>
 
@@ -3052,6 +3054,9 @@ relating to the results of the <a>DID resolution</a> process. This structure is
 REQUIRED and MUST NOT be empty. This metadata typically changes between
 invocations of the <code>resolve</code> function as it represents data about the
 resolution process itself.
+Properties defined by this specification are in <a href="#did-resolution-metadata-properties"></a>.
+If the resolution is successful, this structure MUST contain a <code>content-type</code> property containing the mime-type <a>representation</a> of the <code>did-document-stream</code> in this result.
+If the resolution is not successful, this structure MUST contain an <code>error</code> property describing the error.
             </dd>
             <dt>
 did-document-stream
@@ -3076,6 +3081,7 @@ between invocations of the <code>resolve</code> function unless the <a>DID
 document</a> changes, as it represents data about the <a>DID document</a>. If
 the resolution is unsuccessful, this output MUST be an empty <a
 href="metadata-structure">metadata structure</a>.
+Properties defined by this specification are in <a href="#did-document-metadata-properties"></a>.
             </dd>
         </dl>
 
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -3070,25 +3087,121 @@ <h2>
...
+This specification defines the following common properties.
+            </p>
+            
+            <dl>
+                <dt>
+accept
+                </dt>
+                <dd>
+The MIME type of the caller's preferred <a href="#representations">representation</a> of the <a>DID document</a>. The <a>DID resolver</a> implementation
+MAY use this value to determine the representation contained in the returned <code>did-document-stream</code> if such
```
msporny, line 3126:
```suggestion
SHOULD use this value to determine the representation contained in the returned <code>did-document-stream</code> if such
```

index.html
```diff
@@ -3070,25 +3087,121 @@ <h2>
...
+The possible properties within this structure and their possible values are defined by [[DID-SPEC-REGISTRIES]].
+This specification defines the following common properties.
+            </p>
+
+            <dl>
+                <dt>
+content-type
+                </dt>
+                <dd>
+The mime-type of the returned <code>did-document-stream</code>.
```
msporny, line 3148:
```suggestion
The MIME type of the returned <code>did-document-stream</code>.
```

index.html
```diff
@@ -3070,25 +3087,121 @@ <h2>
...
+This property is REQUIRED if resolution is successful and if the <code>resolveStream</code> function was called.
+It MUST NOT be present if the <code>resolve</code> function was called.
+The value of this property MUST be the mime-type of one of the conformant <a href="#representations">representations</a>.
```
msporny, line 3151:
```suggestion
The value of this property MUST be the MIME type of one of the conformant <a href="#representations">representations</a>.
```

index.html
```diff
@@ -3070,25 +3087,121 @@ <h2>
...
+Properties defined by this specification are in <a href="#did-document-metadata-properties"></a>.
             </dd>
         </dl>
 
         <p>
-<a>DID resolver</a> implementations MUST NOT alter the signature of this
-function in any way. <a>DID resolver</a> implementations MAY map the
-<code>resolve</code> function to a method-specific internal function to perform
+<a>DID resolver</a> implementations MUST NOT alter the signature of these
+functions in any way. <a>DID resolver</a> implementations MAY map the
```
dlongley, line 3102: > <a>DID resolver</a> implementations MUST NOT alter the signature of these
functions in any way.

I understand the intent here but I imagine it may trip up implementers ... at least until we get more specific by using, for example, Infra. Then it may be ok to say this. The fact that we already have a normative statement saying that resolvers must implement this function could also be seen as enough and could avoid issues here.

index.html
dlongley: We may want to tighten this up a bit more and say that it MUST be used to determine the representation if used with `resolveStream` and MUST be ignored if used with `resolve`. That would better fall in line with other statements made elsewhere. Or -- we could just say what the format is here and leave those other normative statements in one place with their respective functions.

index.html
peacekeeper: I think this sentence originally came from @jricher . I'd be fine with removing.

index.html
peacekeeper: I agree with tightening this and would probably also prefer MUST. We can add an error code `unsupported-representation` if the representation is not supported (this would be comparable to an HTTP 406 error).

index.html
dlongley: +1 to just remove it -- since I was also thinking about how there may be some implementations that could provide additional optional parameters, for example, that would not run afoul of any interoperability concerns here. We don't need to be this prescriptive.

index.html
jricher: "SHOULD" is fine but "MUST" is too far here.

index.html
jricher: The goal is to ensure interop. If you have additional functions (which might be how you handle optional parameters) then that's fine.

index.html
```diff
@@ -3004,20 +3004,25 @@ <h2>
...
+        </code></p>
+
+        <p><code>
+resolveStream ( did, did-resolution-input-metadata ) <br>
 &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;-&gt; ( did-resolution-metadata, did-document-stream, did-document-metadata )
         </code></p>
 
         <p>
-The input variables of this function MUST be as follows:
+The input variables of these functions MUST be as follows:
```
jricher, line 3025: I would recommend splitting the definition of these functions into separate subsections with different descriptions about their inputs and outputs instead of trying to put everything in one block as they are here. This is possible to do as an editorial change after this PR is merged, but the issue shouldn't be lost.

index.html
peacekeeper: I'll accept the "SHOULD" suggestion for now, we can later raise a separate issue/PR to change it if desired.

index.html
msporny: Agreed, let's do this in a separate PR given that this has taken so long to get into the spec. Let's set down the base, editorial changes will be easier after that's done.

msporny:
Normative, lots of WG debate in special topic calls, lots of requested and applied changes, we have consensus with no objections, merging.

PR #342, opened 2020-07-03
[WIP] Update specification to use INFRA

msporny:
This PR updates the specification to use the INFRA specification https://infra.spec.whatwg.org/ in an attempt to formalize the abstract data model and concrete representations of that abstract data model.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/342.html" title="Last updated on Jul 14, 2020, 12:02 AM UTC (1664a6f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/342/9d8e245...1664a6f.html" title="Last updated on Jul 14, 2020, 12:02 AM UTC (1664a6f)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -705,6 +705,14 @@ Terminology
     <div data-include="terms.html" data-oninclude="restrictReferences">
     </div>
 
+    <p>
+In addition to the terminology above, this specification also uses terminology
+from the [[INFRA]] specification to formally define the abstract data model.
+When [[INFRA]] terminology is used, such as <a
+data-cite="INFRA#strings">string</a>, <a data-cite="INFRA#lists">list</a>, and
+<a data-cite="INFRA#maps">map</a>, it is linked directly to that specification.
+    </p>
+
   </section>
 
   <section>
--- a/index.html
+++ b/index.html
@@ -790,7 +790,7 @@ used, for example, to identify a specific part of a <a>DID document</a>.
       <p>
 This following is the ABNF definition using the syntax in [[!RFC5324]]. It
 builds on the <code>did</code> scheme defined in <a href="#did-syntax"></a>. The
-<a data-cite="!rfc3986#section-3.3"><code>path-abempty</code></a> and  <a
+<a data-cite="!rfc3986#section-3.3"><code>path-abempty</code></a> and <a
 data-cite="!rfc3986#section-3.5"><code>fragment</code></a> components are
 identical to the ABNF rules defined in [[!RFC3986]], and the
 <code>did-query</code> component is derived from the  <a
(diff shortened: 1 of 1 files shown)


iherman:
@msporny, some editorial comments:

- We have moved to the usage of infra in another specification lately (in [publication manifests](https://www.w3.org/TR/pub-manifest/)). Because infra is not yet a household name, we made it explicit in the spec that we are dependent, in terminology, of infra. That has proven to be useful for our readers. Although you do link to the infra terminology from within the text (which is great) it may be useful, e.g., in the Terminology section, to add a note that we do really on that standard. 

- Infra has a separate section on the [relationships with JSON](https://infra.spec.whatwg.org/#parse-json-into-infra-values). It may be worth checking that those are compatible with our spec in terms of our usage of JSON (ie, section 6) and maybe making this relationship explicit in our text (ie, that we abide to the rules described in infra). I do not expect any problems there, but checking is always a good idea...

msporny:
> We have moved to the usage of infra in another specification lately (in [publication manifests](https://www.w3.org/TR/pub-manifest/)). Because infra is not yet a household name, we made it explicit in the spec that we are dependent, in terminology, of infra. That has proven to be useful for our readers. Although you do link to the infra terminology from within the text (which is great) it may be useful, e.g., in the Terminology section, to add a note that we do really on that standard.

Done in 66c0d5daf13b730ce47ec223cd3b030a57664031.

msporny:
> * Infra has a separate section on the [relationships with JSON](https://infra.spec.whatwg.org/#parse-json-into-infra-values). It may be worth checking that those are compatible with our spec in terms of our usage of JSON (ie, section 6) and maybe making this relationship explicit in our text (ie, that we abide to the rules described in infra). I do not expect any problems there, but checking is always a good idea...

Done in e656eddce2e74a4acf7e065ed243f769a967413a.

I checked, nothing looks out of place, however, we're now repeating content in INFRA (by reference) and should consider eliminating the bulleted lists in the specification and just referring to the Infra/ECMA specifications. Our production/consumption rules are effectively only repeating a very small subset of the requirements in ECMA.

index.html
```diff
@@ -1299,11 +1316,11 @@ <h2>Verification Methods</h2>
         <dt>verificationMethod</dt>
         <dd>
 If a <a>DID document</a> includes a <code>verificationMethod</code> property,
-the value of the property MUST be an unordered set of verification methods,
-where each verification method is described by a set of properties.
-The set of properties MUST include the <code>id</code>, <code>type</code>,
-<code>controller</code>, and specific verification method properties, and
-MAY include additional properties.
+the value of the property MUST be a <a data-cite="INFRA#lists">list</a> of
```
dlongley, line 1319: Should this be an ordered set instead? Same everywhere else that was converted from "set" => "INFRA list".

dmitrizagidulin:
👍

index.html
msporny: Done in 1664a6fca40a0cf8a7df0fa7908609fb3ddabb16.

PR #343, opened 2020-07-05
Added missing reference to the diagram in ECHIDNA

iherman:
The reference to the diagram was missing, as reported in 

https://lists.w3.org/Archives/Public/public-did-wg/2020Jul/0009.html


iherman:
And we can now enjoy the diagram in https://www.w3.org/TR/did-core/#architecture-overview :-)

PR #347, opened 2020-07-14
Add Metadata section based on infra data structures.

msporny:
This PR attempts to write some base text for the metadata section based on infra data structures. It attempts to strike the right balance by strongly suggesting that string-string values should be used. However, if non-string values are required, then infra types that are loss-lessly expressible across all representations must be used. Hopefully this strikes the right compromise.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/347.html" title="Last updated on Jul 28, 2020, 1:16 PM UTC (64e5f41)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/347/78f1148...64e5f41.html" title="Last updated on Jul 28, 2020, 1:16 PM UTC (64e5f41)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -3462,18 +3462,27 @@ Input and output metadata is often involved during the <a>DID Resolution</a>,
 <a>DID URL Dereferencing</a>, and other DID-related processes. The structure
 used to communicate this metadata MUST be a <a data-cite="INFRA#maps">map</a> of
 properties. Each property name MUST be a <a data-cite="INFRA#string">string</a>.
-Each property value SHOULD be a <a data-cite="INFRA#string">string</a>. In the
-rare cases where a non-string value is required, the value and all non-string
-sub-values MUST be expressed as a <a data-cite="INFRA#maps">map</a>, <a
-data-cite="INFRA#lists">list</a>, <a data-cite="INFRA#boolean">boolean</a>,
-number, or  <a data-cite="INFRA#null">null</a> and it MUST be possible to
-losslessly represent the value across all supported representations as
-described in <a href="#core-representations"></a>.
+Each property value MUST be a <a data-cite="INFRA#string">string</a>,
+<a data-cite="INFRA#maps">map</a>, <a data-cite="INFRA#lists">list</a>,
+<a data-cite="INFRA#boolean">boolean</a>, or  <a data-cite="INFRA#null">null</a>.
+The values within any complex data structures such as maps and lists 
+MUST be one of these data types as well. 
+All metadata property definitions MUST define the value type, including any additional
+formats or restrictions to that value (for example, a string formatted as a date or as a decimal integer).
+It is RECOMMENDED that property definitions use strings for values where possible.
         </p>
 
         <p>
-The following example demonstrates a JSON-encoded metadata structure sent to
-the <a href="#did-resolution">DID Resolution process</a>.
+All implementations of functions that use metadata structures as either input or output MUST
+be able to fully represent all data types described here in a deterministic fashion. As inputs and
+outputs using metadata structures are defined in terms of data types and not their serialization,
+the method for representation is internal to the implementation of the function and is out of 
+scope of this specification.
+        </p>
+
+        <p>
+The following example demonstrates a JSON-encoded metadata structure represented
+within a <a href="#did-resolution">DID Resolution process</a>.
         </p>
 
         <pre class="example" title="JSON-encoded DID resolution input metadata example">
@@ -3483,8 +3492,18 @@ the <a href="#did-resolution">DID Resolution process</a>.
         </pre>
 
         <p>
+This example corresponds to a metadata structure of the following format:
+        </p>
+
+        <pre class="example" title="DID resolution input metadata example">
+«[
+  "accept" → "application/did+ld+json"
+]»
+        </pre>
+
+        <p>
 The next example demonstrates a JSON-encoded metadata structure that might be
-returned from the <a href="#did-resolution">DID Resolution process</a> if a
+used in the return of a <a href="#did-resolution">DID Resolution process</a> if a
 DID was not found.
         </p>
 
@@ -3492,6 +3511,16 @@ DID was not found.
 {
   "error": "not-found"
 }
+        </pre>        
+
+        <p>
+This example corresponds to a metadata structure of the following format:
+        </p>
+
+        <pre class="example" title="DID resolution output metadata example">
+«[
+  "error" → "not-found"
+]»
         </pre>
 
     </section>
--- a/index.html
+++ b/index.html
@@ -3456,6 +3456,21 @@ DID URL Dereferencing
         <h2>
 Metadata Structure
         </h2>
+
+        <p>
+Input and output metadata is often involved during the <a>DID Resolution</a>,
+<a>DID URL Dereferencing</a>, and other DID-related processes. The structure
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -3456,6 +3456,44 @@ <h2>
         <h2>
 Metadata Structure
         </h2>
+
+        <p>
+Input and output metadata is often involved during the <a>DID Resolution</a>,
+<a>DID URL Dereferencing</a>, and other DID-related processes. The structure
+used to communicate this metadata MUST be a <a data-cite="INFRA#maps">map</a> of
+properties. Each property name MUST be a <a data-cite="INFRA#string">string</a>.
+Each property value SHOULD be a <a data-cite="INFRA#string">string</a>. In the
```
selfissued, line 3465: The "SHOULD be a string" language is the worst of both worlds.  It encourages developers to write code that assumes strings, that will then break when other object types are used.  Please delete the prejudicial text.

index.html
```diff
@@ -3456,6 +3456,44 @@ <h2>
...
+rare cases where a non-string value is required, the value and all non-string
+sub-values MUST be expressed as a <a data-cite="INFRA#maps">map</a>, <a
+data-cite="INFRA#lists">list</a>, <a data-cite="INFRA#boolean">boolean</a>,
+number, or  <a data-cite="INFRA#null">null</a> and it MUST be possible to
+losslessly represent the value across all supported representations as
+described in <a href="#core-representations"></a>.
```
selfissued, line 3471: Please simplify the text by saying that any representable value, such as strings, booleans, arrays, objects, and nulls may be the value.

index.html
msporny: @jricher we're waiting on feedback from you on this one. The discussion has boiled down to your position vs. @selfissued's position. I made an attempt at a compromise, which doesn't work for @selfissued, so it'll be up to both of you to hash out what you want the "Each property MUST/SHOULD be a string. " sentence to state... or if the sentence should be deleted entirely.

index.html
msporny: @jricher waiting on feedback from you on this one too. Need you to hash this out with @selfissued.

index.html
jricher: I agree that "SHOULD" is really bad in this case as it will lead to premature optimization on the part of consumers, which will break in the way that Mike recommends. My goal was to separate the data structure of the metadata from the serialization of the DID document itself. They should not be tied together, and the reference to core representations here is misleading.

index.html
```diff
@@ -3456,6 +3456,44 @@ <h2>
...
+        </p>
+
+        <p>
+The following example demonstrates a JSON-encoded metadata structure sent to
+the <a href="#did-resolution">DID Resolution process</a>.
+        </p>
```
jricher, line 3477: The resolution process should absolutely NOT accept something "JSON-encoded". It should accept a "map". Serialization is internal to the process, on either input or output from the function.

index.html
jricher, line 3471:
```suggestion
Input and output metadata is often involved during the <a>DID Resolution</a>,
<a>DID URL Dereferencing</a>, and other DID-related processes. The structure
used to communicate this metadata MUST be a <a data-cite="INFRA#maps">map</a> of
properties. Each property name MUST be a <a data-cite="INFRA#string">string</a>.
Each property value MUST be a <a data-cite="INFRA#string">string</a>,
<a data-cite="INFRA#maps">map</a>, <a data-cite="INFRA#lists">list</a>,
<a data-cite="INFRA#boolean">boolean</a>, or  <a data-cite="INFRA#null">null</a>.
The values within any complex data structures such as maps and lists 
MUST be one of these data types as well. 
All metadata property definitions MUST define the value type, including any additional
formats or restrictions to that value (for example, a string formatted as a date).
It is RECOMMENDED that property definitions use strings for values where possible.
```

index.html
```diff
@@ -3456,6 +3456,44 @@ <h2>
...
+The next example demonstrates a JSON-encoded metadata structure that might be
+returned from the <a href="#did-resolution">DID Resolution process</a> if a
+DID was not found.
+        </p>
+
+        <pre class="example" title="JSON-encoded DID resolution output metadata example">
+{
+  "error": "not-found"
+}
+        </pre>
```
jricher, line 3495:
```suggestion
        <p>
All implementations of functions that use metadata structures as either input or output MUST
be able to fully represent all data types described here in a deterministic fashion. As inputs and
outputs using metadata structures are defined in terms of data types and not their serialization,
the method for representation is internal to the implementation of the function and is out of 
scope of this specification.
        </p>

        <p>
The following example demonstrates a JSON-encoded metadata structure represented
within a <a href="#did-resolution">DID Resolution process</a>.
        </p>

        <pre class="example" title="JSON-encoded DID resolution input metadata example">
{
  "accept": "application/did+ld+json"
}
        </pre>

        <p>
This example corresponds to a metadata structure of the following format:
        </p>

        <pre class="example" title="DID resolution input metadata example">
«[
  "accept" → "application/did+ld+json"
]»
        </pre>

        <p>
The next example demonstrates a JSON-encoded metadata structure that might be
used in the return of a <a href="#did-resolution">DID Resolution process</a> if a
DID was not found.
        </p>

        <pre class="example" title="JSON-encoded DID resolution output metadata example">
{
  "error": "not-found"
}
        </pre>        

        <p>
This example corresponds to a metadata structure of the following format:
        </p>

        <pre class="example" title="DID resolution input metadata example">
«[
  "error" → "not-found"
]»
        </pre>
```

index.html
selfissued: @jricher - First, I'm mostly good with your text above.  But one question: INFRA numbers are noticeability missing from your text.  I expect there to be use cases where you want to express a number, such as a timeout value, in metadata.  Why not allow that?

FYI, I'd be fine with the restriction that numbers be integers that are small enough that they can be represented exactly in all INFRA representations.

jricher: Infra doesn't have a native representation for numbers, so I left it off the list of allowed elements. A property would need to define a specific way to serialize a numeric value. We could call that out here.

jricher: I added "or as a decimal integer" to the example of type value restrictions to call out numbers specifically as a missing format.

selfissued: I agree that we need the "or decimal integer" clause, since some metadata will need to contain integers.

jricher: Definitely agree, and I think we could potentially define a set of common string formats for metadata properties to re-use, but those will probably come into light once we have more actual metadata properties defined.

index.html
msporny: Applied suggestions from @jricher, looks like we have consensus as @selfissued seems to agree in the comments. Not seeing any objections.

index.html
msporny: Applied suggestions from @jricher, looks like we have consensus as @selfissued seems to agree in the comments. Not seeing any objections.

index.html
msporny: Applied your suggestions verbatim.

msporny:

msporny:
For what it's worth, I'm exploring how we add "number" to Infra here: https://github.com/whatwg/infra/issues/87

skyevo938 (approved):

index.html
cezary13k: leave as @cezary13k/ @ cybercezaryc13

index.html
cezary13k: leave as @cezary13k /coppilot

Issue #199, opened 2020-02-18
Clarification on what DIDs might identify


brentzundel:
I think there is agreement in the group that DIDs may be used to identify entities such as people, agents, IoT devices, etc. 
What about using DIDs to identify things that are incapable of interacting using cryptographic keys? What if there is no controller property or public keys at all? Could a DID document simply contain, or point to, some content?

Could the following have DIDs?:
- the moon
- a schema, context, or verifiable credential
- an item in a database
- insured assets

Apologies if this is a re-hash of an already concluded conversation. I searched but did not find.

jandrieu:
There are two ideas here. 

The first, yes,  one should be able to use DIDs to identify subjects like the moon,  e.t.c.  which is especially useful for VCs.

However,  using a DID document to publish information about the subject is an anti pattern i actively discourage.  IMO, a DID document should only contain information needed for secure interactions with the subject.  We can't prevent people adding assertions that the moon is made of cheese, but it's really a violation of the moon's privacy' so let's not encouraged that.

iherman:
@jandrieu,  just to clarify, though: while I agree that assertions about the details of the moon are an anti-pattern, it _is_ acceptable to add a reference (I guess as a service) to a home page on the Web authoritatively describing the subject. (I use the 'authoritatively' in a loose sense; e.g. is to refer to my personal Web site, for example.)

dlongley:
I agree with @jandrieu.

@iherman,

> just to clarify, though: while I agree that assertions about the details of the moon are an anti-pattern, it is acceptable to add a reference (I guess as a service) to a home page on the Web authoritatively describing the subject.

In principle, yes, but the same PII rules apply. In my view, I think which statements can/should be made about a DID subject in a DID Document largely depend on who can resolve the DID to get the DID Document -- and the nature of the system that stores these statements, including how it "forgets" such statements when desired. These two qualifiers could perhaps be summed up by saying "it depends on the DID method".

[...]

PR #348, opened 2020-07-15
add representation property

talltree:

***
<a href="https://pr-preview.s3.amazonaws.com/talltree/did-core/pull/348.html" title="Last updated on Aug 14, 2020, 3:15 AM UTC (0c92a62)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/348/4dedfe5...talltree:0c92a62.html" title="Last updated on Aug 14, 2020, 3:15 AM UTC (0c92a62)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1300,7 +1300,7 @@ expected to match the <a>DID</a> that was resolved.
         <p>
         Appendix A classifies the <a>resources</a> identified by <a>DIDs</a>
         into three broad categories. In the first two, the <a>DID subject</a>
-        subject is not an information resource but an entity that exists
+        is not an information resource but an entity that exists
         independent of a digital <a>representation</a> on a network. In the
         third, the <a>DID subject</a>  is an information resource that has a
         digital <a>representation</a>. Common examples of this type of resource
--- a/index.html
+++ b/index.html
@@ -1288,50 +1288,6 @@ expected to match the <a>DID</a> that was resolved.
       </p>
         </section>
 
-        <section>
-        <h2>sameSubject</h2>
-        <p>
-        A <a>DID controller</a> MAY assert that two or more <a>DIDs</a> identify
-        the same <a>DID subject</a> using the <code>sameSubject</code> property. This
-        property is OPTIONAL.
-        </p>
-
-        <dl>
-        <dt>sameSubject</dt>
-        <dd>
-        The value of <code>sameSubject</code> MUST be a
-        <a data-cite="INFRA#lists">list</a> where each item in the list is a
-        <a>DID</a> that conforms to the rules in Section <a href="#did-syntax"></a>.
-        </dd>
-      </dl>
-        <p>
-        From a semantic standpoint, the <a>DID controller</a> is asserting that
-        each <a>DID</a> in the list identifies the same <a>DID subject</a> as
-        the <a>DID</a> described by the <a>DID document</a> (i.e., the value of
-        the <code>id</code> property).
-        </p>
-        <p class="note">
-        WARNING: A <code>sameSubject</code> assertion that two or more <a>DIDs</a>
-        identify the same <a>DID subject</a> does not prove this assertion is
-        true. A malicious <a>DID controller</a> may attempt to impersonate a
-        <a>DID subject</a> by making a false <code>sameSubject</code> assertion.
-        Therefore it is STRONGLY RECOMMENDED that a requesting party obtain
-        independent verification of a <code>sameSubject</code> assertion. One
-        method for doing this is to resolve the <code>sameSubject</code> <a>DID</a>
-        to its own <a>DID document</a> and then verify that it contains the
-        inverse <code>sameSubject</code> assertion (i.e., a backpointer). See
-        the <a href="#security-considerations"></a> and
-        <a href="#privacy-considerations"></a> sections for more information.
-        </p>
-        <p class="note">
-        A <code>sameSubject</code> assertion is not the same as an
-        <code>owl:sameAs</code> assertion. In other words, it does not assert
-        that two or more RDF graphs represented as <a>DID documents</a> can be
-        merged. It only asserts that two or more <a>DIDs</a> identify the same
-        <a>DID subject</a>.
-        </p>
-        </section>
-
       <section>
         <h2>representation</h2>
         <p>
--- a/index.html
+++ b/index.html
@@ -1238,27 +1238,25 @@ The following sections define the properties in a <a>DID document</a>,
 including whether these properties are required or optional.
     </p>
 
-    <section>
-      <h2>
-Identifiers
-      </h2>
-      <p>
+    <p class="note">
 Identifiers are used in the data model to identify specific instances of people,
 organizations, devices, keys, services, and things in general. Identifiers are
-typically URLs, or more generally, <a>URIs</a>. Non-<a>URI</a>-based identifiers
-are not advised because, while the data model supports them, they are not easy
-to use with other Internet-based identifiers.
-      </p>
-    </section>
+typically URLs, or more generally, <a>URIs</a> (and in some cases specifically
+<a>DIDs</a>). Non-<a>URI</a>-based identifiers are not advised because, while
+the data model supports them, they are not easy to use with other Internet-based
+identifiers.
+    </p>
 
     <section>
       <h2>DID Subject</h2>
 
+      <section>
+        <h2>id</h2>
+
       <p>
 The <a>DID subject</a> is denoted with the <code>id</code> property. The
-<a>DID subject</a> is the entity that the <a>DID document</a> is about. That is,
-it is the entity identified by the <a>DID</a> and described by the
-<a>DID document</a>.
+<a>DID subject</a> is the entity identified by the <a>DID</a> and described by
+the <a>DID document</a>.
       </p>
 
       <p>
@@ -1288,6 +1286,128 @@ fully resolved <a>DID document</a> always contains a valid <code>id</code>
 property. The value of <code>id</code> in the resolved <a>DID document</a> is
 expected to match the <a>DID</a> that was resolved.
       </p>
+        </section>
+
+        <section>
+        <h2>sameAs</h2>
+        <p>
+        A <a>DID controller</a> MAY assert that two or more <a>DIDs</a> identify
+        the same <a>DID subject</a> using the <code>sameAs</code> property. This
+        property is OPTIONAL.
+        </p>
+
+        <dl>
+        <dt>sameAs</dt>
+        <dd>
+The value of <code>sameAs</code> MUST be a <a
+data-cite="INFRA#lists">list</a> where each item in the list is a <a>DID</a>
+that conforms to the rules in Section <a href="#did-syntax"></a>.
+        </dd>
+      </dl>
+        <p>
+        From a semantic standpoint, the <a>DID controller</a> is asserting that
+        each <a>DID</a> in the list identifies the same <a>DID subject</a> as
+        the <a>DID</a> described by the <a>DID document</a> (i.e., the value of
+        the <code>id</code> property).
+        </p>
+        <p class="note">
+        WARNING: An assertion of semantic equivalence using the
+        <code>sameAs</code> property does not prove this assertion is true.
+        A malicious <a>DID controller</a> may attempt to impersonate a
+        <a>DID subject</a> by making a false <code>sameAs</code> assertion.
+        Therefore it is STRONGLY RECOMMENDED that a requesting party obtain
+        independent verification of a <code>sameAs</code> assertion. One method
+        of doing this is to resolve the <code>sameAs</code> <a>DID</a> to its
+        own <a>DID document</a> and then verify that it contains the inverse
+        <code>sameAs</code> assertion (i.e., a backpointer). See the
+        <a href="#security-considerations"></a> and
+        <a href="#privacy-considerations"></a> sections for more information.
+        </p>
+        </section>
+
+      <section>
+        <h2>representation</h2>
+        <p>
+       As described in Appendix A, a <a>DID</a> is an abstract identifier that
+       does not resolve directly to a <a>representation</a> of the resource
+       identified by the <a>DID</a> (the <a>DID subject</a>). Rather the
+       <a>DID resolution</a> process produces a <a>DID document</a> which in
+       turn describes the <a>DID subject</a>.
+        </p>
+        <p>
+        Appendix A classifies the <a>resources</a> identified by <a>DIDs</a>
+        into three broad categories. In the first two, the <a>DID subject</a>
+        subject is not an information resource but an entity that exists
+        independent of a digital <a>representation</a> on a network. In the
+        third, the <a>DID subject</a>  is an information resource that has a
+        digital <a>representation</a>. Common examples of this type of resource
+        include web pages, PDF files, data schemas, or digital files of any
+        type.
+        </p>
+        <p>
+        In this third case, a <a>DID controller</a> MAY use the
+        <code>representation</code> property to: 1) include a <a>representation</a>
+        of the <a>DID subject</a> directly within the <a>DID document</a>,
+        and/or 2) reference a <a>representation</a> of the <a>DID subject</a>
+        directly from the <a>DID document</a>. Since an information resource
+        may have multiple <a>representations</a>, these two options are not
+        mutually exclusive.
+        </p>
+
+        <dl>
+        <dt>representation</dt>
+        <dd>
+        The <code>representation</code> property MAY have three subproperties:
+        <code>type</code>, <code>uri</code>, and <code>value</code>.
+        </dd>
+        </dl>
+
+        <dl>
+        <dt>type</dt>
+        <dd>
+        The <code>type</code> property MUST be a <a>URI</a> that asserts the
+        MIME type of the <a>representation</a>.
+        </dd>
+        </dl>
+
+        <dl>
+        <dt>uri</dt>
+        <dd>
+        The <code>uri</code> property MUST be a <a>URI</a>. This <a>URI</a>
+        SHOULD be a reference to a <a>representation</a> of the <a>DID subject</a>
+        that is not contained within the <a>DID document</a> but at some other
+        network location.
+        </dd>
+        </dl>
+
+        <dl>
+        <dt>value</dt>
+        <dd>
+        The <code>value</code> property contains a <a>representation</a> of
+        the <a>DID subject</a>. This value MUST first be encoded in the MIME
+        type indicated by the <code>type</code> subproperty. If necessary, the
+        value MUST then be re-encoded according to the requirements of the
+        <a>DID document</a> <a>representation</a> itself. See Section
+        <a href="#core-representations"></a>.
+        </dd>
+      </dl>
+
+        <p>
+        If both the <code>uri</code> property and the <code>value</code>
+        property are present and non-empty, the value of the <code>uri</code>
+        property SHOULD resolve to the same <a>representation</a> contained in
+        the <code>value</code> property.
+        </p>
+
+        <p>
+        If there is more than one <code>representation</code> property present,
+        each instance of the property describes a different <a>representation</a>
+        of the same <a>DID subject</a>. Each instance SHOULD have a different
+        value of the <code>type</code> property if present.
+        </p>
+
+      </section>
+
     </section>
 
     <section>
--- a/index.html
+++ b/index.html
@@ -1289,19 +1289,19 @@ expected to match the <a>DID</a> that was resolved.
         </section>
 
         <section>
-        <h2>sameAs</h2>
+        <h2>sameSubject</h2>
         <p>
         A <a>DID controller</a> MAY assert that two or more <a>DIDs</a> identify
-        the same <a>DID subject</a> using the <code>sameAs</code> property. This
+        the same <a>DID subject</a> using the <code>sameSubject</code> property. This
         property is OPTIONAL.
         </p>
 
         <dl>
-        <dt>sameAs</dt>
+        <dt>sameSubject</dt>
         <dd>
-The value of <code>sameAs</code> MUST be a <a
-data-cite="INFRA#lists">list</a> where each item in the list is a <a>DID</a>
-that conforms to the rules in Section <a href="#did-syntax"></a>.
+        The value of <code>sameSubject</code> MUST be a
+        <a data-cite="INFRA#lists">list</a> where each item in the list is a
+        <a>DID</a> that conforms to the rules in Section <a href="#did-syntax"></a>.
         </dd>
       </dl>
         <p>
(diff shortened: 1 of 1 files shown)


talltree:
While I would normally propose two new properties into two separate PRs, @brentzundel wisely suggested that both of these fit neatly as subsections of [DID Subject](https://w3c.github.io/did-core/#did-subject) under [Core Properties](https://w3c.github.io/did-core/#core-properties). So I submitted them as one PR. 

A suggestion to commenters: put `sameAs` or `representation` at the start of your comment so as to make it clear which property you are commenting on.

The `representation` property addresses issue #199 . Note that the PR references a proposed Appendix A that I will submit (hopefully tomorrow)as a separate PR.

The `sameAs` property addresses the need to able to assert semantic equivalence of two or more DIDs (i.e., that they all identify the same DID subject). This has been an issue since we put it in the Future Work section 3+ years ago. More recently it has been raised by @OR13 and @peacekeeper.

iherman:
Reading the text with my Semantic Web hat on, the immediate question is "is this `sameAs` _THE_ `sameAs`?". Ie, does the DID JSON-LD context map this term to `owl:sameAs` (just like `type` is mapped on `rdf:type`)?

I presume the answer is (or should be) 'no': it is a term defined for the sole usage in DID Documents for DID subjects. (E.g., I presume it is not acceptable to use `sameAs` to connect, say, key identifiers.) But, I believe, it is worth making this very explicit in the document to avoid people like me asking stupid questions...

(One could bikeshed on whether the term `sameAs` is, under the circumstances, a judicious choice for the aforementioned reasons. But it is difficult to find a better one.)

index.html
```diff
@@ -1288,6 +1286,128 @@ <h2>DID Subject</h2>
...
+        A <a>DID controller</a> MAY assert that two or more <a>DIDs</a> identify
+        the same <a>DID subject</a> using the <code>sameAs</code> property. This
+        property is OPTIONAL.
+        </p>
+
+        <dl>
+        <dt>sameAs</dt>
+        <dd>
+The value of <code>sameAs</code> MUST be a <a
+data-cite="INFRA#lists">list</a> where each item in the list is a <a>DID</a>
```
OR13, line 1303: we might want to relax this to be a URI...

`did:example:123 owl:sameAs https://example.com/person/456`

index.html
```diff
@@ -1288,6 +1286,128 @@ <h2>DID Subject</h2>
...
+        <dd>
+        The <code>representation</code> property MAY have three subproperties:
+        <code>type</code>, <code>uri</code>, and <code>value</code>.
+        </dd>
+        </dl>
+
+        <dl>
+        <dt>type</dt>
+        <dd>
+        The <code>type</code> property MUST be a <a>URI</a> that asserts the
```
OR13, line 1368: is this a URI? I would expect `application/json`, `application/ld+json`... prefer to see examples here.

OR13:
I obviously disagree, see the comment manu linked :)

From the charter:

> Recommend a data model and syntax(es) for the expression of Decentralized Identifier Documents, including _one or more_ core vocabularies.

we have to remember here that not everyone thinks about linked data the same way... and there is absolutely nothing wrong with using OWL with DIDs... every triple in an RDF representation of a DID Document has the same security issues.... there is nothing different about sameAs... except that its from a vocabulary set thats already widely used...

all the statements about a subject in a did document are the same kind of thing... statements... made by a controller.

`did:example:123  capabilityInvocation did:example:123#456`
`did:example:123  ccg:alsoKnownAs https://twitter.com/@Lord_Voldemort7`
`did:example:123  owl:sameAs https://twitter.com/@Lord_Voldemort7`

Each of these is only as "trustworthy" as the DID Method + the DID Controller's opsec... anyone can claim anything in their decentralized identity document... thats the whole point of being free...

We are not "Solving Identity" on the Web... we can't make statements like "did:example:123 is/is not provably lord voldemort"... and we can't stop did:example:123 from claiming to be a dog, or a software package.... we can recommend for or against it... I'm in favor of running towards this problem by using OWL, and addressing this issue directly, not creating a new kind of equivalence... which is again only as trustworthy as the controller and did method.

If my use case is I want to use OWL with DIDs... does my using `owl:sameAs` prevent you from using `ccg:alsoKnownAs` ?

AFAIK, this might be the first DID Document JSON-LD vocabulary term that is not primarily defined and maintained by the W3C CCG. 

Its also fundamental to the public DID use cases that @csuwildcat and others are so excited about: https://rubenverborgh.github.io/WebFundamentals/semantic-web/#timbl-linked-data

See also: 

- https://en.wikipedia.org/wiki/Semantic_reasoner
- http://dig.csail.mit.edu/2006/Papers/AAAI/
- https://pdfs.semanticscholar.org/360e/ee82db5317b22754bd94429fa78133eec00e.pdf
- https://github.com/semantalytics/awesome-semantic-web#reasoners
- https://github.com/mdesalvo/RDFSharp

I'm not opposed to supporting more than 1 kind of equivalence relationship... but specifically blocking support for existing semantic reasoning / software out of the gate feels like a bad move...  OWL exists already... DIDs are new... let's build a bridge not an island.

talltree:
First, I was not aware of [the CCG Security Vocab thread on @OR13 's sameAs PR](https://github.com/w3c-ccg/security-vocab/pull/45) when I posted this PR. Thanks to @msporny for pointing it out.

As I posted there, I think I understand the concerns @msporny articulates on that thread. And I also appreciate @OR13 's desires to be able to pursue full owl:sameAs semantics.

So let me just clarify that the purpose of this PR (despite the proposed `sameAs` name for the property) was just to establish a standard way for a DID controllers to assert that one DID _identifies the same DID subject_ as a different DID. Full stop. No additional semantics or DID document equivalence or RDF graph merging were implied.

So I'm wide open to giving this property a different name, and leaving `sameAs` for a separate PR for a separate property that would convey the full OWL semantic graph meaning (if @OR13 wants to pursue that).

Since I don't personally like the human semantics of the term "alsoKnownAs" (it feels too colloquial to me), I'd like to propose the property name `sameSubject`.

Thoughts?

talltree:
> This is excellent, I think we need to discuss:
> 
> 1. should sameAs work with any IRI (I suggest yes).

@OR13 First I just want to clarify that my answer to your question assumes (per the comment I just posted) that we will split into two different properties—one called `sameAs` for the full OWL semantic graph equivalence that you are seeking and one called something else (I have proposed `sameSubject`) to assert only that two DIDs identify the same DID subject.

This PR only covers the latter (I will update it to the new property name as soon as we settle on it).

Now, in that context, my answer is "No" to allowing IRIs (or other URIs that are not DIDs) because I think it is very important to restrict the `sameSubject` property to identification by DIDs. Ironically, the reasons do in fact go deep into semantic web questions. But a full explanation demands more time that I can give this right now—I plan to address this with a proposed Appendix that I'm currently drafting. I'll update the comment thread as soon as I post it.

talltree:
> This is excellent, I think we need to discuss:
> 
> 1. should sameAs work with any IRI (I suggest yes).
> 2. should type be a value from this list: https://www.iana.org/assignments/media-types/media-types.xhtml

@OR13 on #2, I agree that the value SHOULD be from that list. I'm wary of making it a MUST as a DID controller may have a reason to assert a type that has not yet been registered. What do you think?

talltree:
> I think there is general consensus to support limited "alsoKnownAs" use cases.

@msporny Would you be okay with `sameSubject` as the property name (vs. `alsoKnownAs`)?

OR13:
> > This is excellent, I think we need to discuss:
> > 
> > 1. should sameAs work with any IRI (I suggest yes).
> 
> @OR13 First I just want to clarify that my answer to your question assumes (per the comment I just posted) that we will split into two different properties—one called `sameAs` for the full OWL semantic graph equivalence that you are seeking and one called something else (I have proposed `sameSubject`) to assert only that two DIDs identify the same DID subject.
> 
> This PR only covers the latter (I will update it to the new property name as soon as we settle on it).
> 
> Now, in that context, my answer is "No" to allowing IRIs (or other URIs that are not DIDs) because I think it is very important to restrict the `sameSubject` property to identification by DIDs. Ironically, the reasons do in fact go deep into semantic web questions. But a full explanation demands more time that I can give this right now—I plan to address this with a proposed Appendix that I'm currently drafting. I'll update the comment thread as soon as I post it.

interested to here more about this, I would assume that such a restriction is not actually legal for JSON-LD / OWL / RDF...

https://www.w3.org/TR/json-ld11/#iris
https://www.w3.org/TR/owl-ref/#sameAs-def
https://w3c.github.io/did-core/#json-ld

When I add these up... I feel like making a restriction to DIDs only for `owl:sameAs`... is a bad idea... but I could see such an argument for `sameSubject`.

talltree:
@OR13 to be clear, I'm not proposing to restrict `sameAs` to just DIDs. Only `sameSubject`. 

I agree with you that if a `sameAs` property will have the same semantics as owl:sameAs, then the domain needs to be all IRIs as it is in RDF/OWL.

talltree:
I just revised the PR to change the name of the `sameAs` property to the `sameSubject` property. I also added a note to explain that the semantics of the `sameSubject` assertion are limited to asserting that two or more DIDs identify the same DID subject. They do NOT assert `owl:sameAs` semantics, i.e., the ability to merge the RDF graphs represented by the respective DID docs.

That's not to say that `owl:sameAs` semantics do not have value or should not be pursued by those who might need them, but that's out of scope for this PR.

index.html
talltree: @OR13 Yes, that's a relative URI. I updated the PR to include your suggestion that the value SHOULD come from the IANA media type registry.

index.html
talltree: @OR13 Per the other discussions on this topic, I changed the property name to `sameSubject`. For reasons I will explain in a new Appendix I am composing, I favor keeping the value of `sameSubject` restricted to DIDs.

index.html
```diff
@@ -1288,6 +1286,137 @@ <h2>DID Subject</h2>
...
+        <p>
+        From a semantic standpoint, the <a>DID controller</a> is asserting that
+        each <a>DID</a> in the list identifies the same <a>DID subject</a> as
+        the <a>DID</a> described by the <a>DID document</a> (i.e., the value of
+        the <code>id</code> property).
+        </p>
+        <p class="note">
+        WARNING: A <code>sameSubject</code> assertion that two or more <a>DIDs</a>
+        identify the same <a>DID subject</a> does not prove this assertion is
+        true. A malicious <a>DID controller</a> may attempt to impersonate a
```
msporny, line 1316:
```suggestion
        true. A malicious <a>DID controller</a> might attempt to impersonate a
```

index.html
```diff
@@ -1288,6 +1286,137 @@ <h2>DID Subject</h2>
...
+        <a>DID subject</a> by making a false <code>sameSubject</code> assertion.
+        Therefore it is STRONGLY RECOMMENDED that a requesting party obtain
```
msporny, line 1318:
```suggestion
        Therefore it is strongly advised that a requesting party obtain
```

index.html
```diff
@@ -1288,6 +1286,137 @@ <h2>DID Subject</h2>
...
+        independent verification of a <code>sameSubject</code> assertion. One
+        method for doing this is to resolve the <code>sameSubject</code> <a>DID</a>
+        to its own <a>DID document</a> and then verify that it contains the
+        inverse <code>sameSubject</code> assertion (i.e., a backpointer). See
```
msporny, line 1322:
```suggestion
        inverse <code>sameSubject</code> assertion, also known as a back pointer. See
```

index.html
```diff
@@ -1288,6 +1286,137 @@ <h2>DID Subject</h2>
...
+        the <a href="#security-considerations"></a> and
+        <a href="#privacy-considerations"></a> sections for more information.
+        </p>
+        <p class="note">
+        A <code>sameSubject</code> assertion is not the same as an
+        <code>owl:sameAs</code> assertion. In other words, it does not assert
+        that two or more RDF graphs represented as <a>DID documents</a> can be
+        merged. It only asserts that two or more <a>DIDs</a> identify the same
+        <a>DID subject</a>.
```
msporny, line 1331:
```suggestion
        <a>DID subject</a>. Implementers are strongly advised against the use of <code>owl:sameAs</code> when processing <a>DID documents</a> as its use could expand the attack surface in a way that was not considered by the Working Group.
```

index.html
```diff
@@ -1288,6 +1286,137 @@ <h2>DID Subject</h2>
...
+        </p>
+        </section>
+
+      <section>
+        <h2>representation</h2>
```
msporny, line 1336: This property is begging to be abused from a GDPR perspective. :)

There are better and more specific properties in schema.org... `image`, `url`, etc. We shouldn't advise implementers to express this sort of information in the DID Document... it should be done externally.

Please split the `representation` PR into a separate PR... as the `sameSubject` PR is far less controversial. I'm concerned that this PR is going to be held up because of the `representation` PR... they should progress on different timelines.

index.html
```diff
@@ -1288,6 +1286,137 @@ <h2>DID Subject</h2>
...
+        <dd>
+        The <code>type</code> property MUST be a <a>URI</a> that asserts the
+        media type of the <a>representation</a>. The value of this property
+        SHOULD be a media type from the
+        <a href="https://www.iana.org/assignments/media-types/media-types.xhtml">IANA Media Type list </a>.
+        </dd>
+        </dl>
+
+        <dl>
+        <dt>uri</dt>
```
msporny, line 1383: This should be `id`

index.html
```diff
@@ -1288,6 +1286,137 @@ <h2>DID Subject</h2>
...
+        <dl>
+        <dt>representation</dt>
+        <dd>
+        The <code>representation</code> property MAY have three optional
+        subproperties: <code>type</code>, <code>uri</code>, and <code>value</code>.
+        </dd>
+        </dl>
+
+        <dl>
+        <dt>type</dt>
```
msporny, line 1373: `type` is being re-used in a way that is not compliant with the DID spec... this really should be `mediaType` and the value isn't a URI.

index.html
```diff
@@ -1288,6 +1286,137 @@ <h2>DID Subject</h2>
...
+        the <a>DID subject</a>. This value MUST first be encoded in the media
+        type indicated by the <code>type</code> subproperty. If necessary, the
+        value MUST then be re-encoded according to the requirements of the
+        <a>DID document</a> <a>representation</a> itself. See Section
+        <a href="#core-representations"></a>.
+        </dd>
+      </dl>
+
+        <p>
+        If both the <code>uri</code> property and the <code>value</code>
```
msporny, line 1405: Hrm, no, we should be using data URIs if we want to embed values directly... and we should put that value in `id` (see above). Also see https://en.wikipedia.org/wiki/Data_URI_scheme for explanation on data URIs.

index.html
```diff
@@ -1288,6 +1286,128 @@ <h2>DID Subject</h2>
...
+        the <code>id</code> property).
+        </p>
+        <p class="note">
+        WARNING: An assertion of semantic equivalence using the
+        <code>sameAs</code> property does not prove this assertion is true.
+        A malicious <a>DID controller</a> may attempt to impersonate a
+        <a>DID subject</a> by making a false <code>sameAs</code> assertion.
+        Therefore it is STRONGLY RECOMMENDED that a requesting party obtain
+        independent verification of a <code>sameAs</code> assertion. One method
+        of doing this is to resolve the <code>sameAs</code> <a>DID</a> to its
```
brentzundel:
```suggestion
        of doing this would be to resolve each <code>sameAs</code> <a>DID</a> to its
```

index.html
```diff
@@ -1288,6 +1286,128 @@ <h2>DID Subject</h2>
...
+       As described in Appendix A, a <a>DID</a> is an abstract identifier that
+       does not resolve directly to a <a>representation</a> of the resource
+       identified by the <a>DID</a> (the <a>DID subject</a>). Rather the
+       <a>DID resolution</a> process produces a <a>DID document</a> which in
+       turn describes the <a>DID subject</a>.
+        </p>
+        <p>
+        Appendix A classifies the <a>resources</a> identified by <a>DIDs</a>
+        into three broad categories. In the first two, the <a>DID subject</a>
+        subject is not an information resource but an entity that exists
```
brentzundel, line 1347:
```suggestion
        is not an information resource but an entity that exists
```

index.html
```diff
@@ -1288,6 +1286,128 @@ <h2>DID Subject</h2>
...
+        </dd>
+        </dl>
+
+        <dl>
+        <dt>value</dt>
+        <dd>
+        The <code>value</code> property contains a <a>representation</a> of
+        the <a>DID subject</a>. This value MUST first be encoded in the MIME
+        type indicated by the <code>type</code> subproperty. If necessary, the
+        value MUST then be re-encoded according to the requirements of the
```
brentzundel, line 1398: an example of this would be very helpful

talltree:
> Let's split this PR up into two PRs... one for `sameSubject` and another for `representation`... I expect they will progress on different timelines, with `sameSubject` being less controversial than `representation`.

@msporny Agreed. I probably should have done that at the outset. I will try to do that by EOD tomorrow (Wed 22 July).

I'll wait to do the new PR before responding to some of your comments on `representation` (all of which I believe there are good answers to).

talltree:
Revised PR to remove the sameSubject property, which is now addressed by PR #355.

index.html
kdenhartog: +1

talltree:
> For the representation property can, I'm not really sure I'm following the purpose of this. The only thing I'm coming up with is when the DID Subject is a online resource, but in that case I'm not understanding the advantages of using a DID rather than something like a URN. Could you provide me any additional use cases for this so I can understand it and reason about it better?

@kdenhartog You nailed it—the whole reason for this PR is that a DID can, in certain cases, be much better than a URN. Just take every use case for a URN, and add these advantages of a DID:

1. **No centralized registry required.** Almost all URNs require centralized or federated registry services (the exception is URNs like UUIDs that are not resolvable at all).
1. **Cryptographically verifiability.** The DID controller of the immutable object that serves as a DID subject—for example a schema—is able to publish a series of versions of that schema that each can be cryptographically verified as published by that same DID controller. While not all persistent digital objects need that (there is a whole class of immutable objects the Decentralized Semantics folks call "ungoverned objects" that are just "fire and forget" and are never versioned). But if you need verifiable provenance of an immutable object, a DID is perfect.
1. **Standardized resolution.** First of all, not all URNs are resolvable (e.g., UUIDs). But if they are resolvable, DIDs have a significant advantage because every URN scheme needs to define not just its URN identifier format and resolution protocol, _but also its own metadata format_. Check out the [IANA URN Namespaces registry](https://www.iana.org/assignments/urn-namespaces/urn-namespaces.xhtml) for a sense of how many resolution protocols and metadata formats are involved. 

While it is true that each DID method also has its own concrete resolution protocol, the abstract DID resolution contract—and most importantly the **resolution result** (a DID document)—is standardized for all DIDs.

That's a major advantage over URNs. Furthermore, with the `representation` property, we are uniting persistent, resolvable, cryptographically verifiable identification of the world of digital objects (what the Semantic Web folks call "information objects") with persistent, resolvable, cryptographically verifiable identification of the world of all non-digital objects. So we can use one comprehensive standard for both.

How cool is that?

index.html
peacekeeper: I agree with that, but also want to mention that this PR seems to serve a similar purpose as various web linking and discovery technologies such as JRD and the HTML <link> element. Both of those use `type` to declare a media type, not an RDF class.

OR13: if we have an opportunity to pick a better name, we should... `id` and `type` are used all over the place... +1 for not continuing to overload them.... `Content-Type` isn't `type`... https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Type

msporny: Waiting on this to be resolved to do a re-review of this PR.

index.html
msporny: I think @OR13 is stating that using `type` is the wrong property to use here?

kdenhartog:
> > For the representation property can, I'm not really sure I'm following the purpose of this. The only thing I'm coming up with is when the DID Subject is a online resource, but in that case I'm not understanding the advantages of using a DID rather than something like a URN. Could you provide me any additional use cases for this so I can understand it and reason about it better?
> 
> @kdenhartog You nailed it—the whole reason for this PR is that a DID can, in certain cases, be much better than a URN. Just take every use case for a URN, and add these advantages of a DID:
> 
>     1. **No centralized registry required.** Almost all URNs require centralized or federated registry services (the exception is URNs like UUIDs that are not resolvable at all).
> 
>     2. **Cryptographically verifiability.** The DID controller of the immutable object that serves as a DID subject—for example a schema—is able to publish a series of versions of that schema that each can be cryptographically verified as published by that same DID controller. While not all persistent digital objects need that (there is a whole class of immutable objects the Decentralized Semantics folks call "ungoverned objects" that are just "fire and forget" and are never versioned). But if you need verifiable provenance of an immutable object, a DID is perfect.
> 
>     3. **Standardized resolution.** First of all, not all URNs are resolvable (e.g., UUIDs). But if they are resolvable, DIDs have a significant advantage because every URN scheme needs to define not just its URN identifier format and resolution protocol, _but also its own metadata format_. Check out the [IANA URN Namespaces registry](https://www.iana.org/assignments/urn-namespaces/urn-namespaces.xhtml) for a sense of how many resolution protocols and metadata formats are involved.
> 
> 
> While it is true that each DID method also has its own concrete resolution protocol, the abstract DID resolution contract—and most importantly the **resolution result** (a DID document)—is standardized for all DIDs.
> 
> That's a major advantage over URNs. Furthermore, with the `representation` property, we are uniting persistent, resolvable, cryptographically verifiable identification of the world of digital objects (what the Semantic Web folks call "information objects") with persistent, resolvable, cryptographically verifiable identification of the world of all non-digital objects. So we can use one comprehensive standard for both.
> 
> How cool is that?

In this case, I'd put myself at a +0 then. I'm not thinking we'll use this for resource descriptions at this point, but I expect that down the road it will be useful for implementers who are expanding on the ways in which DIDs. I'd be much more willing to move to a +1 if we saw more use cases adopting DIDs in this manor.

index.html
```diff
@@ -1258,6 +1258,52 @@ <h1>Core Properties</h1>
...
+      <li>
+<code><a>capabilityInvocation</a></code>: defined in <a href="#capabilityinvocation"></a>.
+      </li>
+      <li>
+<code><a>service</a></code>: defined in <a href="#service-endpoints"></a>.
+Sub-properties include <code>id</code>, <code>type</code> and
+<code><a>serviceEndpoint</a></code>.
+      </li>
+      <li>
+<code><a>created</a></code>: defined in <a href="#created"></a>.
```
OR13: did we agree that created and updated go in the did document? I recall strongly objecting to that.

msporny: We did not agree to put `created` and `updated` in the DID Document. We left it open that some DID Methods may choose to do that, but if they do, it's a statement asserted by the `controller` of said document.

talltree:
I just pushed a rewrite of this property based on input from this thread and the discussion on today's DID WG special topic call on the proposed DID Appendices (see #373). This rewrite removes any direct references to those proposed Appendices so it can now stand on its own (though the proposed Appendices may still be helpful for understanding this property).

talltree:
This PR has been superseded by the proposed `type` property in #410. I am going to mark this PR "pending close".

brentzundel:
No comments since marked pending close, closing

PR #351, opened 2020-07-16
URI not URL in section 1.1

philarcher:
I think section 1.1 has a typo where it says that `did` is the scheme - surely that should be URI?
This PR changes that single character (from L to I)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/351.html" title="Last updated on Jul 16, 2020, 11:40 AM UTC (adfc367)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/351/78f1148...adfc367.html" title="Last updated on Jul 16, 2020, 11:40 AM UTC (adfc367)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -320,7 +320,7 @@ A <a>DID</a> is a simple text string consisting of three parts, the:
       </p>
         <ul>
           <li>
-URL scheme identifier (<code>did</code>)
+URI scheme identifier (<code>did</code>)
           </li>
           <li>
 Identifier for the <a>DID method</a>


dlongley:
Perhaps relevant: https://www.w3.org/International/wiki/IRIStatus

talltree:
> Perhaps relevant: https://www.w3.org/International/wiki/IRIStatus

@dlongley In this case the debate around IRIs and URLs isn't relevant. The spec we're conforming to with the DID spec is RFC 3986 which is the URI spec. So Phil is correct that we should make this change. I'm approving it.

dlongley:
@talltree, from the link:

> Standardize on the term URL. URI and IRI are just confusing. In practice a single algorithm is used for both so keeping them distinct is not helping anyone. URL also easily wins the search result popularity contest.

dlongley:
We don't have to agree or follow that particular advice, I'm just pointing out that this has been raised in a few places now -- and it might be to our detriment to use the term URI at the end of the day.

msporny:
> I think section 1.1 has a typo where it says that did is the scheme - surely that should be URI?

It's not a typo :)

Now, before you throw that chair at me... let me explain, by pointing to the goals in the WHATWG URL specification:

https://url.spec.whatwg.org/#goals

Namely:

"Standardize on the term URL. URI and IRI are just confusing. In practice a single algorithm is used for both so keeping them distinct is not helping anyone. URL also easily wins the search result popularity contest."

I find that goal the most compelling, though all the other goals are compelling as well. I say this as a card-carrying member of the URI/IRI camp: https://www.w3.org/TR/2014/REC-json-ld-20140116/#iris

It's time to move on -- the precise nature of URI/IRI has never helped me explain this stuff to a non-standards geek developer... usually, it's done the opposite as I've had to forcefully insert the pedantic differences between an URL and a URI and a IRI, which tend to be unnecessary. As the WHATWG URL spec states... in practice, there is only one algorithm... the URL parsing algorithm.

I have been waiting to bring this up in the WG, knowing it may kick off some sort of URI/IRI vs. URL holy war... so before people start throwing spears, I do want to highlight that I made a pass through the specification and we could replace almost every instance of URI with URL (including adding a sentence or two about the distinction we're making and pointing to the WHATWG URL specification).

I know that this will probably not be very popular with folks that have been in this space for decades and have put in the effort of understanding the distinction. It's a distinction that's not serving us well (other than serving as an interesting historical origin story).

Thoughts?

index.html
```diff
@@ -320,7 +320,7 @@ <h2>
       </p>
         <ul>
           <li>
-URL scheme identifier (<code>did</code>)
+URI scheme identifier (<code>did</code>)
```
msporny, line 323: This is not a typo, see this comment:

https://github.com/w3c/did-core/pull/351#issuecomment-665055361

philarcher:
Hmmm... having written and deleted seven replies 'cos they're some version of rude, insulting, inflammatory, unhelpful, probably questionable - you get the picture - I'll desist from further comment. A fairy just died.

iherman:
> I know that this will probably not be very popular with folks that have been in this space for decades and have put in the effort of understanding the distinction. It's a distinction that's not serving us well (other than serving as an interesting historical origin story).

I am one of those guys, and no, I am not happy but... but I know that we have to yield to realities and I have argued, in some other WG-s, to use URL-s and not URI-s. As @philarcher said, a fairy died, and it has been dead for a while, actually. 

But, regardless of the fact that we lost this battle out there in general, I am not sure whether it is that simple to change it in _this_ document. First of all, as @talltree said, we refer to the URI spec in the formal specification. We do not refer to the WhatWG document, and we register the 'did' scheme as part of the IETF. Ie, a DID is a URI in that context.

Maybe more importantly, if we say a _DID_ is a URL, then what is a _DID URL_? Isn't that the source of a major confusion? We spent quite a lot of time separating the concept of a DID from that of a DID URL, and we may have re-start the bikeshedding discussion on what to call a DID URL... I do not think we want that.

My (personal, not W3C!) conclusion is that yes, a fairy has died but, in this document, it makes the document _much_ clearer if the strict separation of a DID and DID URL is kept, meaning that a DID is not a URL but... a URI.

jonnycrunch:
[RFC 3986 section-1.1.3](https://tools.ietf.org/html/rfc3986#section-1.1.3) regarding the URL, URN and URI states that  Future specifications and related documentation should use the general term "URI" rather than the more restrictive terms "URL" and "URN".  Given we are really defining a new class of decentralized identifiers, seems reasonable that we don't get constrained with a subset of thinking.

talltree:
I want to reinforce that, since we are defining a URI scheme specification based on [RFC 3986](https://www.ietf.org/rfc/rfc3986.txt) (the URI specification), I don't think it is possible for use to avoid using the term "URI".

However this thread convinces me that I would be very supportive of adding a Note, perhaps right after our first use of the term "URI" in the spec, to explain what @dlongley and others on the thread have pointed out about the prevalence of the term URL—and also explaining why we do not use "URL" exclusively. 

Thoughts?

msporny:
> However this thread convinces me that I would be very supportive of adding a Note, perhaps right after our first use of the term "URI" in the spec, to explain what @dlongley and others on the thread have pointed out about the prevalence of the term URL—and also explaining why we do not use "URL" exclusively.

Yes, we should do this.

msporny (approved):
Alright, looks like this is going in since many in the group believe we are defining a URI scheme, among other things.

We'll look for a future PR as to why we use URI and URL throughout the spec.

PR #357, opened 2020-07-24
Add 'relative-ref' DID URL parameter. Fixes #349.

### LINKED ISSUE #349: Add relative-ref DID URL parameter
Since matrix parameters were removed from the DID URL syntax, one consequence is that an extra DID URL parameter is needed that can contain an encoded service endpoint path and query string. This has been originally mentioned by @dlongley in https://github.com/w3c/did-core/issues/159#issuecomment-598286814 and used again in an example use case described by @OR13 in https://github.com/w3c/did-core/issues/327. This DID URL parameter would always be used in conjunction with the `service` parameter which is already in the spec.

Example

`
did:ex:123?service=files&relative-ref=%2Fmyresume%2Fdoc%3Fversion%3Dlatest%23intro
`

peacekeeper:
This fixes https://github.com/w3c/did-core/issues/349

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/357.html" title="Last updated on Jul 29, 2020, 8:53 AM UTC (73e89f7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/357/6988450...73e89f7.html" title="Last updated on Jul 29, 2020, 8:53 AM UTC (73e89f7)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1030,9 +1030,7 @@ Also discuss hierarchical method namespaces in DID parameter names.
         <p>
 A <a>DID path</a> is identical to a generic URI path and MUST conform to the <a
 data-cite="!rfc3986#section-3.3"><code>path-abempty</code></a> ABNF rule in
-[[!RFC3986]]. A <a>DID path</a> SHOULD be used to address <a>resources</a>
-available through a <a>service endpoint</a>. For more information, see Section
-<a href="#service-endpoints"></a>.
+[[!RFC3986]].
         </p>
 
         <p>
--- a/index.html
+++ b/index.html
@@ -901,6 +901,21 @@ The associated value MUST be an
 <a data-lt="ascii string">ASCII string</a>.
               </td>
             </tr>
+
+            <tr>
+              <td>
+<code>relative-ref</code>
+              </td>
+              <td>
+A relative URI reference according to <a
+data-cite="RFC3986#section-4.2">RFC3986 Section 4.2</a>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -901,6 +901,21 @@ <h2>Generic DID URL Parameters</h2>
...
             </tr>
+
+            <tr>
+              <td>
+<code>relative-ref</code>
+              </td>
+              <td>
+A relative URI reference according to <a
+data-cite="RFC3986#section-4.2">RFC3986 Section 4.2</a>
+that identifies a resource at a <a>service endpoint</a>, which is selected from
```
Fak3, line 912: Section 3.2.3 about DID Path says:

> A DID path SHOULD be used to address resources available through a service endpoint.

How someone could decide whether he must use DID Path or _relative-ref_ parameter when he creates a relative url into a service endpoint? This has to be clarified here and in Section 3.2.3

peacekeeper: Good point. Based on previous discussions on the topic, the sentence in 3.2.3 should be removed, since the `relative-ref` parameter will be used for this. There have even been proposals to remove DID Path altogether, but then the consensus was that DID Path may have other uses that have nothing to do with service endpoints. I will update this PR accordingly.

msporny:
Multiple reviews, changes requested and made, no objections, normative, merging.

PR #362, opened 2020-07-28
Update context requirements for JSON-LD producers/consumers.

dlongley:
While fixing the language around contexts, I noticed what I believe to be an erroneous comment regarding the value of `id` that states it *must* be a DID. However, this is *only* true when referring to the DID subject. The value of `id` of a service endpoint or verification method, for example, is not a DID.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/362.html" title="Last updated on Sep 1, 2020, 7:48 PM UTC (bf5a714)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/362/ce2f9a9...bf5a714.html" title="Last updated on Sep 1, 2020, 7:48 PM UTC (bf5a714)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2481,7 +2481,8 @@ this specification as idiomatic JSON.
         <li>
 Even though JSON-LD allows any IRI as node identifiers, <a>DID documents</a>
 are explicitly restricted to only describe DIDs. This means that the value of
-<code><a>id</a></code> MUST be a valid <a>DID</a> and not any other kind of IRI.
+<code><a>id</a></code> that refers to the <a>DID subject</a> MUST be a valid
+<a>DID</a> and not any other kind of IRI.
         </li>
         <li>
 Data types, such as integers, dates, units of measure, and URLs, are
@@ -2507,8 +2508,11 @@ property.
 The value of the <code>@context</code> property MUST be one or more <a>URIs</a>,
 where the value of the first <a>URI</a> is
 <code>https://www.w3.org/ns/did/v1</code>. All members of the
-<code>@context</code> property MUST exist be in the DID properties extension
-registry.
+<code>@context</code> property SHOULD exist in the DID specification
+registries in order to achieve interoperability across different
+representations. If a member does not exist in the DID specification
+registries, then the DID Document will not be interoperable across
+representations.
           </dd>
         </dl>
 
@@ -2543,7 +2547,11 @@ where the value of the first <a>URI</a> is
 <code>https://www.w3.org/ns/did/v1</code>. If more than one <a>URI</a> is
 provided, the <a>URIs</a> MUST be interpreted as an ordered set. It is
 RECOMMENDED that dereferencing each <a>URI</a> results in a document containing
-machine-readable information about the context.
+machine-readable information about the context. To enable interoperability
+across representations, <a>URIs</a> that exist in the DID specification
+registries SHOULD be assumed to refer to the information in the registries. In
+this case, if dereferencing is required during processing, the information from
+the registries SHOULD be used.
           </dd>
         </dl>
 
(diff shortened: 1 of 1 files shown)


dlongley:
@jricher, let me know what you think about the language here.

index.html
```diff
@@ -2573,8 +2574,11 @@ <h3>
 The value of the <code>@context</code> property MUST be one or more <a>URIs</a>,
 where the value of the first <a>URI</a> is
```
jricher, line 2575:
```suggestion
where the value of the first <a>URI</a> MUST be
```

index.html
```diff
@@ -2573,8 +2574,11 @@ <h3>
...
 <code>https://www.w3.org/ns/did/v1</code>. All members of the
-<code>@context</code> property MUST exist be in the DID properties extension
-registry.
+<code>@context</code> property SHOULD exist in the DID specification
+registries in order to achieve interoperability across different
+representations. If a member does not exist in the DID specification
+registries, then the DID Document will not be interoperable across
```
jricher, line 2580:
```suggestion
registries, then the DID Document might not be interoperable across
```

dlongley:
@jricher,

Thanks for your suggestions! I've applied them both.

> I'm a little confused as to why the JSON-LD consumer section doesn't need to mention the context property and its contents, but is that just because it's a clear and obvious property of JSON-LD that we shouldn't specify again here?

That is my take, yes. If others disagree, please speak up and suggest concrete changes.

peacekeeper:
I agree the requirement that `id` MUST be a DID is only true for the DID subject. I also agree with the clarifications around the use of `@context`.

index.html
```diff
@@ -2609,7 +2613,11 @@ <h3>
 <code>https://www.w3.org/ns/did/v1</code>. If more than one <a>URI</a> is
 provided, the <a>URIs</a> MUST be interpreted as an ordered set. It is
 RECOMMENDED that dereferencing each <a>URI</a> results in a document containing
-machine-readable information about the context.
+machine-readable information about the context. To enable interoperability
+across representations, <a>URIs</a> that exist in the DID specification
+registries SHOULD be assumed to refer to the information in the registries. In
+this case, if dereferencing is required during processing, the information from
+the registries SHOULD be used.
```
msporny, line 2620: I don't know what this is saying. It's vague. What does "refer to the information in the registries" mean? Definitions? Links to specifications? What does it mean to use information from the registry?

OR13: is this a normative pointer to a non normative registry (wg note) ?... I agree this could be clearer... prefer permanent links.

dlongley: URIs here refer to contexts. There might be additional metadata about those contexts (content hashes, etc.) in the registry that should be used to ensure there is no ambiguity about which context is being referred to. Can you propose some concrete text that would help?

msporny: I'm still at a loss as to what problem is attempting to be addressed with the text? Are you saying "The DID Spec Registries might contain information in them that are associated with URIs, such as cryptographic hashes, that further enhance the security of the system. Implementers are encouraged to use any extra information to ensure that producers and consumers are secure." Or something to that effect? I don't really like that language either because it's vague. You could also delete the sentence. :)

msporny: Still waiting on response to this to merge.

dlongley:
I will look into resolving the conflicts and looking at the comments here.

index.html
msporny: Dave and I talked about it... trying to propose some text.

index.html
```diff
@@ -2543,7 +2547,11 @@ <h3>
...
```
msporny, line 2554:
```suggestion
machine-readable information about the context. To enable interoperability
with other representations, <a>URLs</a> registered in the 
DID Specification Registries [[DID-SPEC-REGISTRIES]] referring to 
JSON-LD Contexts SHOULD be associated with a cryptographic hash of 
the content of the JSON-LD Context. This ensures that the interpretation of the
information by JSON-LD consumers will be the same as interpretations 
over other representations by other consumers that rely on the
DID Specification Registries [[DID-SPEC-REGISTRIES]].
```

msporny:

PR #365, opened 2020-07-30
Move "created" and "updated" from Core Properties to DID document metadata

peacekeeper:
Addresses #203 and #174.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/365.html" title="Last updated on Aug 10, 2020, 1:59 PM UTC (6663b38)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/365/3cc2641...6663b38.html" title="Last updated on Aug 10, 2020, 1:59 PM UTC (6663b38)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -3497,8 +3497,8 @@ describe timestamps associated with the DID document.
 
         <pre class="example" title="JSON-encoded DID document metadata example">
 {
-  "created": "2002-10-10T17:00:00Z",
-  "updated": "2016-10-17T02:41:00Z"
+  "created": "2019-03-23T06:35:22Z",
+  "updated": "2023-08-10T13:40:06Z"
 }
         </pre>
 
@@ -3508,8 +3508,8 @@ This example corresponds to a metadata structure of the following format:
 
         <pre class="example" title="DID document metadata example">
 «[
-  "created" → "2002-10-10T17:00:00Z",
-  "updated" → "2016-10-17T02:41:00Z"
+  "created" → "2019-03-23T06:35:22Z",
+  "updated" → "2023-08-10T13:40:06Z"
 ]»
         </pre>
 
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -3518,6 +3492,30 @@ <h2>
 ]»
         </pre>
 
+        <p>
+The next example demonstrates a JSON-encoded metadata structure that might be
+used as <a href="#did-document-metadata-properties">DID document metadata</a> to
+describe timestamps of the DID document.
```
msporny, line 3498:
```suggestion
describe timestamps associated with the DID document.
```

index.html
```diff
@@ -3518,6 +3492,30 @@ <h2>
...
+        </p>
+
+        <pre class="example" title="JSON-encoded DID document metadata example">
+{
+  "created": "2002-10-10T17:00:00Z",
```
msporny, line 3503: I'm wondering if we should use a date that is within the range of when the DID spec existed. So, perhaps using created dates after 2018 and update dates after 2020 would be more realistic? :)

msporny:
Merge conflicts need to be resolved.

index.html
peacekeeper: Updated the example dates in https://github.com/w3c/did-core/pull/365/commits/260972366213973017c1b8bdba2f49d250de7ed5.

peacekeeper:
@msporny thanks, merge conflicts should be fixed now.

peacekeeper:
> However, this PR does not address issue #174 - as it does not specify the semantics of these values.

Thanks @selfissued, the PR says the following:

> DID document metadata SHOULD include an updated property **to indicate the timestamp of the last Update operation.**

I think **"timestamp of the last Update operation"** should be sufficient as a definition. My idea is that if additional explanations are needed, then those should be added to the section about the [Update operation](https://w3c.github.io/did-core/#update), not to the metadata property.

msporny:
Mostly editorial, multiple positive reviews, changes requested and made, merging.

PR #371, opened 2020-08-11
Update note about controller value 

rhiaro:
following feedback from @dlongley and @jandrieu in #122

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/371.html" title="Last updated on Aug 31, 2020, 6:46 PM UTC (330a828)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/371/ce2f9a9...330a828.html" title="Last updated on Aug 31, 2020, 6:46 PM UTC (330a828)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1389,12 +1389,12 @@ the root of authority for information expressed in the <a>DID Document.</a> If
 the <code><a>controller</a></code> property is not present, the value can be
 inferred to be the same as that of the <code><a>id</a></code> property of the
 <a>DID document</a>. The way to determine how the <a>DID Document</a> can be
-changed is by consulting the relevant <a>DID method</a>. It is possible that
-the <a>DID method</a> will indicate that the <a>DID document</a> can no longer
-be changed beyond whatever its present value is, or that there are means for
-updating the <a>DID document</a> that are unrelated to the
-<code><a>controller</a></code> property. If no <a>verification methods</a> are
-present, it is unnecessary to express a <code><a>controller</a></code> value.
+changed is by consulting the relevant <a>DID method</a>. If no <a>verification
+methods</a> are present in the <a>DID Document</a>, then a
+<code><a>controller</a></code> property MUST be present, otherwise the
+the <a>DID document</a> can no longer be changed beyond whatever its present
+value is.
+```suggestion
       </p>
     </section>
 
--- a/index.html
+++ b/index.html
@@ -1384,16 +1384,17 @@ malicious activity by an attacker. See Section
       </pre>
 
       <p class="note" title="Determining the DID controller">
-If the <code><a>controller</a></code> property is not present, the
-<a>DID controller</a> can be determined by consulting the respective
-<a>DID method</a> specification. It cannot be assumed that the
-<a>DID subject</a> is also the controller of its own <a>DID</a>, nor
-can it be assumed that <em>only</em> entities identified by <a>DIDs</a>
-present in the value of <code><a>controller</a></code> are capable of exerting
-control over the <a>DID</a> in question. Similarly, <a>DID methods</a> are not
-obliged to respect the value of <code><a>controller</a></code> and can specify
-means for updating the <a>DID document</a> that are unrelated to the
-<code><a>controller</a></code> property.
+If the <code><a>controller</a></code> property is present, its value expresses
+the root of authority for information expressed in the <a>DID Document.</a> If
+the <code><a>controller</a></code> property is not present, the value can be
+inferred to be the same as that of the <code><a>id</a></code> property of the
+<a>DID document</a>. The way to determine how the <a>DID Document</a> can be
+changed is by consulting the relevant <a>DID method</a>. It is possible that
+the <a>DID method</a> will indicate that the <a>DID document</a> can no longer
+be changed beyond whatever its present value is, or that there are means for
+updating the <a>DID document</a> that are unrelated to the
+<code><a>controller</a></code> property. If no <a>verification methods</a> are
+present, it is unnecessary to express a <code><a>controller</a></code> value.
       </p>
     </section>
 
--- a/index.html
+++ b/index.html
@@ -1359,6 +1359,15 @@ certain verification methods for specific purposes.
           </dd>
       </dl>
 
+      <p>
+When a <code><a>controller</a></code> property is present in a
+<a>DID Document</a>, its value expresses one or more <a>DIDs</a>. Any
+<a>verification methods</a> contained in the <a>DID Documents</a> for those
+<a>DIDs</a> SHOULD be accepted as authoritative, such that proofs that satisfy
+those verification methods are to be considered equivalent to proofs provided
+by the <a>DID Subject</a>.
+      </p>
+
       <p class="note" title="Authorization vs authentication">
 Note that Authorization is separate from <a href="#authentication"></a>.
 This is particularly important for key recovery in the case of key loss,
@@ -1382,20 +1391,6 @@ malicious activity by an attacker. See Section
   }]
 }
       </pre>
-
-      <p class="note" title="Determining the DID controller">
-If the <code><a>controller</a></code> property is present, its value expresses
-the root of authority for information expressed in the <a>DID Document.</a> If
-the <code><a>controller</a></code> property is not present, the value can be
-inferred to be the same as that of the <code><a>id</a></code> property of the
-<a>DID document</a>. The way to determine how the <a>DID Document</a> can be
-changed is by consulting the relevant <a>DID method</a>. If no <a>verification
-methods</a> are present in the <a>DID Document</a>, then a
-<code><a>controller</a></code> property MUST be present, otherwise the
-the <a>DID document</a> can no longer be changed beyond whatever its present
-value is.
-```suggestion
-      </p>
     </section>
 
     <section>


index.html
```diff
@@ -1388,16 +1388,17 @@ <h2>
...
+the root of authority for information expressed in the <a>DID Document.</a> If
+the <code><a>controller</a></code> property is not present, the value can be
+inferred to be the same as that of the <code><a>id</a></code> property of the
+<a>DID document</a>. The way to determine how the <a>DID Document</a> can be
+changed is by consulting the relevant <a>DID method</a>. It is possible that
+the <a>DID method</a> will indicate that the <a>DID document</a> can no longer
+be changed beyond whatever its present value is, or that there are means for
+updating the <a>DID document</a> that are unrelated to the
+<code><a>controller</a></code> property. If no <a>verification methods</a> are
+present, it is unnecessary to express a <code><a>controller</a></code> value.
```
dlongley, line 1401:
```suggestion
<a>DID document</a>. The way to determine how the <a>DID Document</a> can be
changed is by consulting the relevant <a>DID method</a>. If no <a>verification
methods</a> are present in the <a>DID Document</a>, then a
<code><a>controller</a></code> property MUST be present, otherwise the
the <a>DID document</a> can no longer be changed beyond whatever its present
value is.
```suggestion

File: index.html [hunk 2]
```diff
@@ -1388,16 +1388,17 @@ <h2>
...
-<code><a>controller</a></code> property.
+If the <code><a>controller</a></code> property is present, its value expresses
+the root of authority for information expressed in the <a>DID Document.</a> If
+the <code><a>controller</a></code> property is not present, the value can be
+inferred to be the same as that of the <code><a>id</a></code> property of the
+<a>DID document</a>. The way to determine how the <a>DID Document</a> can be
+changed is by consulting the relevant <a>DID method</a>. If no <a>verification
+methods</a> are present in the <a>DID Document</a>, then a
+<code><a>controller</a></code> property MUST be present, otherwise the
+the <a>DID document</a> can no longer be changed beyond whatever its present
```
jandrieu, line 1399: This doesn't seem accurate, in that the DID Method *can* provide a means to update the did document without specifying a controller property. In particular, I believe BTCR defers all change properties to BTC, without regard to what is in the DID Document.

index.html
```diff
@@ -1388,16 +1388,17 @@ <h2>
...
+value is.
```
dlongley, line 1400: @jandrieu -- would this address your concern?

```suggestion
<code><a>controller</a></code> property MUST be present, otherwise it
is possible that the <a>DID document</a> can no longer be changed beyond
whatever its present value is. This can be determined by consulting the
applicable <a>DID method</a> specification.
```

jandrieu: It seems that the "MUST" is misplaced. If "MUST" is used, that's a normative requirement, which means the following phrase of "otherwise it is possible" is not actually possible with a conformant implementation.

Are you intending to say that, as a requirement,
* DID Methods MUST return DID Documents that either have a verification method or a controller property. 
* DID Methods MUST respect the controller property

That's what this language suggests to me. 

However, I don't believe BTCR does either of this today (I could be wrong), which makes me wonder if this particular use is actually in sync with a minimal set of capability for DIDs as envisioned to date.

brentzundel, line 1400: I'm not comfortable with the proposed text, would this re-wording be acceptable?

```suggestion
If the <code><a>controller</a></code> property is present, its value expresses
the root of authority for information expressed in the <a>DID Document.</a> If
the <code><a>controller</a></code> property is not present, the root of
authority for information expressed in the <a>DID Document</a> is determined
by the applicable <a>DID Method</a> specification. 
```

rhiaro: I thought this while I was drafting it and then apparently completely forgot: we can't use normative language in a note block, so this MUST needs to come out anyway.

dlongley: @brentzundel -- How does one write DID-method agnostic verification software if we can't assume that the controller is the subject when `controller` is not present? It's one thing to say that it might not be clear who can further makes changes to the document (or not) when there's no controller and no verification methods present, but that's a matter left to DID methods. The problem, I believe, with your suggestion is that it affects DID Document consumers like DID-method agnostic verification software.

dlongley: I'm fine if this is reworded to remove the "MUST" since it's in a note. The language I'm looking for here is an indication to DID-method agnostic consumers (such as verification software) such that it can understand the root of authority over what is expressed in the DID Document. This should either be the `controller` (if present) or the `subject`, IMO. Whether or not DID methods have implementation specific details for updating DID Documents is a separate matter to me. I want consuming software to be able to be structured such that it only need to decide whether or not it trusts a given DID method/resolver for that method and then from there, once it has a DID Document, it can assume the root authority is as stated above (thereby allowing consistent verification code).

agropper:
Is my comment
https://github.com/w3c/did-core/issues/373#issuecomment-672427360 relevant
to this issue (#371) ?

index.html
brentzundel: If there is no `controller` property and no verification methods, what should DID-method agnostic software assume about the DID Document?

dlongley: @brentzundel, that the root of authority is the DID subject. Unless the software is for updating the DID Document, it doesn't care about whether or not the DID Document can be further changed, it just cares about what is presently in the DID Document.

brentzundel: That makes sense, I have adjusted my suggestion accordingly.
I think it is cleaner if this note doesn't also attempt to address  whether or not a DID Document can be further changed.

dlongley: @brentzundel thanks, I'm ok with this change. @jandrieu what do you think?

jandrieu: But the root of authority is the DID Controller, which is not the property _controller_, but rather the party(ies) that actually control the DID Document, which is up to the DID Method to specify.

This latest language would require DID Methods to specify a controller field that actually controls the DID Document whenever the actual Controller is not the Subject. However, the DID Method can't actually know whether or not the Controller is the Subject. (And currently methods don't have to rely on the controller property; they are all free to design & implement their own approach to decentralized root of authority).

I would go the other way:
> If the <code>controller</code> property is not present, the root of
authority for information expressed in the <code>DID Document</code> is any entity capable of 
updating the DID Document as specified by the <code>DID Method</code>.

jandrieu: Or...
> If the controller property is not present, then root of
authority for information expressed in the DID Document 
cannot be determined for the purpose of verification methods.
Any DID Method that wishes to support verification methods MUST 
provide a controller value. However, DID Methods 
are *not* required to support verification methods, nor are they 
required to rely on verification methods for DID Document updates.

dlongley, line 1400: @jandrieu and @brentzundel -- what about this? I'm trying to cover all of the concerns here at once: ensuring we're speaking clearly to what `controller` enables verifiers to *consider* in a DID-method-agnostic way and indicating that DID methods may say whatever they want about what a verifier actually needs to accept for a DID Document update.

```suggestion
If the <code><a>controller</a></code> property is present, its value expresses
the root of authority for information expressed in the <a>DID Document</a>.
Expressing a <code><a>controller</a></code> property in a <a>DID Document</a>
enables <a>verifiers</a> to consider proofs generated by a
<a>DID controller</a> on behalf of a <a>DID subject</a>, even the same
<a>DID Document</a> expresses no <a>verification methods</a>. Note that
whether any proofs will ultimately be accepted for some purpose is within the
purview of the <a>verifier</a>. DID methods, for example, may require
<a>verifiers</a> to use a mechanism totally unrelated to <a>verification
methods</a> for validating DID Document updates.
```

dlongley: @jandrieu -- please see my tweak to your above below. Let me know what you think.

agropper:
I'm not following all of the subtleties of this discussion, but I hope we
end up with clarity, ideally an example, around a non-repudiable signature.

In the prescription use-case, the DID subject is able to sign a
prescription for a controlled substance under DEA regs. What does this mean:
- the DID method will need to be "certified" by a federal agency or their
delegate?
- the DID document will need to be co-controlled by a federal agent?
- changes to the DID document will need to be logged and audited?
- the use of the DID by the doctor will need to be logged and audited?
- all of the above?

- Adrian

jandrieu:
I think it would make sense to concisely enumerate the options. The current text reads as if the controller property is required to verify proofs. I realize it doesn't say that, but opening with the statement that the controller allows XX, even when YY doesn't clearly illuminate the fact that one may use either controller or verification methods (or both) to verify proofs and that for DID Documents with neither, a Requesting Party cannot rely upon the DID Document for proof verification.

> 
To enable verification methods, a DID Document must include either a "controller" property, one or more verification methods, or both. If one or more verification methods are present, then Requesting Parties may verify proofs by checking the proof mechanisms whose verification purpose matches the proof's proof purpose.  If a "controller" property is present in an initial DID Document, and the value of that property resolves to additional verification methods, such as from a secondary DID Document, then Requesting Parties MUST treat those additional verification methods as if they were present in the initial DID Document. If there are no verification methods specified: neither in the initial DID Document nor in any secondary sources referred to by a "controller" property, then Requesting Parties cannot rely on the DID Document to verify proofs.

We should probably also clarify whether or not Requesting Parties MUST also resolve all controller properties of secondary DID Documents. Namely, I'm fairly sure the "controller" property requires supporting arbitrary depth trees and should have language defining how to handle circular references.

dlongley:
@jandrieu -- I approve of the direction, but things are a little inverted in the actual suggested language with respect to how proofs/verification methods are handled. I'll discuss with you offline a bit and see if we can resolve it.

dlongley:
I'm looking, perhaps, for language that is more like:

"Systems or protocols that make use of verification methods (which may or may not include DID methods) can use the "controller" property in DID Documents to consider proofs generated by a controller on behalf of a DID subject as valid."

index.html
brentzundel: This text is fine for me, except the last sentence. I keep trying to parse it and failing, so I'm not sure sure you're trying to say there.

msporny:
We're still waiting for this conversation thread to converge on something. There remain a number of outstanding questions, IIUC. What do we need to move this PR forward?

index.html
brentzundel: changed my suggestion to reflect @jandrieu 's input.

jandrieu: Brent. I'm not seeing your suggestion. Is it 
> If no <a>verification
methods</a> are present in the <a>DID Document</a>, then a
<code><a>controller</a></code> property MUST be present, otherwise the
the <a>DID document</a> can no longer be changed beyond whatever its present

If I'm tracking the "current" proposal, then this is equivalent to adding a requirement that the DID Method actually respect the "controller" value AND must put EITHER verification methods or a controller property.

The fact is, today, a DID Method 
1. *can* ignore the value of a controller property
2. *can* leave controller property empty even when there are no verification methods
3. *can* provide a DID Document update operation that uses neither verification methods nor the controller value

What I'm hearing is that #3 is something that would like to be changed: that DID Methods MUST provide a DID Update mechanism that uses either Verification Methods or the controller property.  This would be a new requirement.

brentzundel: the text I've recommended is here: https://github.com/w3c/did-core/pull/371#discussion_r469486922

I don't think #3 should be changed, or if it is, it should explicitly allow a DID Method to include neither a controller nor a verification property.

jandrieu: Hmmm... I need to talk more offline with Dave, I think. The current language (your latest per that comment) suffers from a lack of clarity of what "root authority" means.

I was focused on the fact that the ultimate authority for the content of a DID Document is the DID Controller as supported by the DID Method. This Controller is the party that ACTUALLY can change the document.  The "controller" property is NOT about the ultimate authority as it may or may not indicate the Controller per the DID Method.

So when the "controller" property is called the "root authority", my initial read is that is meant to clarify who has the lowest level control over a DID Document: who can update or create the document.

But what Dave means is about verification methods and the authority for using such methods.

Conflating the two notions of Controller and "controller" AND the two notions of "root" and "ultimate" authority is what has this text wrapped around the axle.

I think we should change the property from "controller" to something else. The namespace collision with "Controller" is a nightmare that is only going to continue as more developers start working with DIDs.

Secondly, I think we should define and use different terms for the "root authority" for verification methods and the ultimate authority to update the document.

brentzundel: @jandrieu @dlongley [The suggestion above](https://github.com/w3c/did-core/pull/371/files#r469486922) contains the text we all agreed to during our call today.

@rhiaro please let us know if [this text](https://github.com/w3c/did-core/pull/371/files#r469486922) is also acceptable to you.

dlongley: +1 to the above text. Thanks!

rhiaro: Thanks for working through this all! My only quibble is the normative language, as this is a note. I could just lowercase the SHOULD when I update the PR, or we could alter it to not say may and should at all - any preference?

brentzundel: Could we change it from a note to a regular paragraph?

jandrieu: +1 to the above text AND to change it from a not to a regular paragraph.

dlongley: +1 to the text and to change it from a note to a regular paragraph.

msporny:
Still waiting for this discussion to converge. I believe everyone is waiting on @rhiaro to note if she's ok with @brentzundel's text (but I may have my PRs confused).

rhiaro:
I have updated the section to use @brentzundel's [wording here](https://github.com/w3c/did-core/pull/371#discussion_r469486922) ... however (sorry!) I'm now not satisfied this answers my original question, which is basically what is a DID doc consumer expected to think/do/understand if there is no `controller` property? 

(The answer I think was concluded to be "consult the DID method" but we seem to have lost that wording.)

It also duplicates to some extent, I think, text that is already in the definition of the `controller` property:

> The value of the `controller` property MUST be a string or an ordered set of strings that conform to the rules in Section <a
href="#did-syntax"></a>. The corresponding DID document(s) SHOULD contain verification relationships that explicitly permit the use of certain verification methods for specific purposes.

dlongley:
@rhiaro,

> I have updated the section to use @brentzundel's wording here ... however (sorry!) I'm now not satisfied this answers my original question, which is basically what is a DID doc consumer expected to think/do/understand if there is no controller property?

Well, we intentionally described what the presence of the `controller` property *enables* as opposed to talking about what happens when it isn't there. When it's not there, you just don't get the feature you get when it *is* there. I don't think there's anything more to it. So there's nothing special to think/do/understand for a DID doc consumer.

rhiaro:
Okay, I'm fine to go with the most recent changes and close the issue.

msporny (approved):

msporny:

Issue #368, opened 2020-08-10
Ability for DID in a Method to be known as more than one ID, with a canonical selection


csuwildcat:
Is there an existing, standard mechanism that allows a DID Method to express that two DIDs in its system are provably, deterministically the same? For example: Say you have one representation of the DID that is 1000 characters long, and another that is the hash of that 1000 characters. These are provably the same in the target system, so is there a way to disclose the hashed version and the 1000 character version are the same DID? If not, could we add a property like `canonical` so a DID Method can stipulate this in the returned DID Document?

OR13:
@talltree @dmitrizagidulin @msporny I suppose this a variation on the question of `sameSubject` vs `alsoKnownAs`....

the short answer for me is that `sameAs` was created to do exactly this....  

https://github.com/w3c/did-core/issues/33
https://github.com/w3c-ccg/security-vocab/pull/45
https://github.com/w3c/did-spec-registries/issues/80

Knowing your use case.... I don't see any reason why anyone would object to:

`shortFormDID sameAs longFormDID`

but clearly @msporny and others are concerned about the security implications of RDF statements.... see the issues linked.

peacekeeper:
I agree with @OR13 that this would probably be done with the new "same" property (whatever its name will be in the end).

Some DID methods will allow the controller to put these "same" properties into the DID document however they want. Other DID methods such as the one proposed here will have this property "built-into" their DID documents.

csuwildcat:
@peacekeeper I am having a hard time with the 'same' definition, because what I am really seeking is a way to express "If a canonical property/value is not present, use the version of the identifier you used to resolve this document, but if the canonical property/value is present, you SHOULD use that identifier instead from here forward." Same seems to indicate that they are interchangeable, but what I want is unidirectionality and for people to cease use of the longer form of the ID if there is a shorter canonical one present in the DID Document. Can you help me understand how this would be achieved in a normative way across all DID Methods using 'same'? My goal is for there to be no ambiguity about the directionality, and have it be a generally understood paradigm.

OR13:
@csuwildcat sounds like you are looking for https://www.w3.org/OWL/#:~:text=The%20W3C%20Web%20Ontology%20Language,things%2C%20and%20relations%20between%20things.

Want to describe how things are related using an international standard.... and not reinvent the wheel?

`sameAs` is one OWL relation.... there are others.

csuwildcat:
[...]

PR #376, opened 2020-08-20
Clarify id prop language to allow equivalent DID values

csuwildcat:
***
<a href="https://pr-preview.s3.amazonaws.com/csuwildcat/did-core/pull/376.html" title="Last updated on Aug 26, 2020, 2:16 PM UTC (2559e6a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/376/4dedfe5...csuwildcat:2559e6a.html" title="Last updated on Aug 26, 2020, 2:16 PM UTC (2559e6a)">Diff</a>


index.html
```diff
@@ -1330,8 +1330,10 @@ <h2>DID Subject</h2>
 such as when a <a>DID resolver</a> is performing <a>DID resolution</a>.
 However, the fully resolved <a>DID document</a> always contains a valid
 <code><a>id</a></code> property. The value of <code><a>id</a></code> in the
-resolved <a>DID document</a> is expected to match the <a>DID</a> that was
-resolved.
+resolved <a>DID document</a> SHOULD match the <a>DID</a> that was
```
kdenhartog, line 1333: Wait this isn't a MUST? Seems odd to me that I'd map did:example:123 => { "id": "did:example:456", ... }

Can we provide some further detail on why this is best left as a should?

csuwildcat:
There was no normative, explicit directive at all before this, so I put in SHOULD simply because there was no feedback yet. Personally, I couldn't think of anything else it would/should ever be either, so I'm happy to go with MUST.

csuwildcat:
@kdenhartog to be clear: if it was

csuwildcat:
Sorry, GitHub's UI freaked out and bizarrely closed the issue, so pay no attention to the Closed entry 😜

index.html
```diff
@@ -1330,8 +1330,10 @@ <h2>DID Subject</h2>
 such as when a <a>DID resolver</a> is performing <a>DID resolution</a>.
 However, the fully resolved <a>DID document</a> always contains a valid
 <code><a>id</a></code> property. The value of <code><a>id</a></code> in the
-resolved <a>DID document</a> is expected to match the <a>DID</a> that was
-resolved.
+resolved <a>DID document</a> MUST match the <a>DID</a> that was
+resolved, or be populated by the <a>DID method</a> with another <a>DID</a> it
```
kdenhartog, line 1334: I'm good with this. The "or..." statement seems like it could cause some tricky issues with how de-referencing would work on identifying the did document, such as sub-objects (verificationMethod block, service block, etc) being identified with a different base did than the top level id. However, since the point of this statement is to point out that the method authors must state how it's canonically mapped, I think it should work out fine as long as the method is written properly.

kdenhartog (approved):

index.html
```diff
@@ -1330,8 +1330,10 @@ <h2>DID Subject</h2>
 such as when a <a>DID resolver</a> is performing <a>DID resolution</a>.
 However, the fully resolved <a>DID document</a> always contains a valid
 <code><a>id</a></code> property. The value of <code><a>id</a></code> in the
-resolved <a>DID document</a> is expected to match the <a>DID</a> that was
-resolved.
+resolved <a>DID document</a> MUST match the <a>DID</a> that was
+resolved, or be populated by the <a>DID method</a> with another <a>DID</a> it 
+determines to be the equivalent canonical <a>DID</a> for the <a>DID</a> 
+that was resolved.
```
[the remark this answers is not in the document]
msporny: Waiting on you to resolve this @csuwildcat...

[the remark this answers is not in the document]
csuwildcat: Will get to it today!

csuwildcat:
@msporny I added in @dlongley's change, so I think we're good, right?

index.html
```diff
@@ -1330,8 +1330,10 @@ <h2>DID Subject</h2>
 such as when a <a>DID resolver</a> is performing <a>DID resolution</a>.
 However, the fully resolved <a>DID document</a> always contains a valid
 <code><a>id</a></code> property. The value of <code><a>id</a></code> in the
-resolved <a>DID document</a> is expected to match the <a>DID</a> that was
-resolved.
+resolved <a>DID document</a> MUST match the <a>DID</a> that was
+resolved, or be populated according to the <a>DID method</a> with another
+<a>DID</a> it specifies is the equivalent canonical <a>DID</a> for the <a>DID</a> 
+that was resolved.
```
brentzundel, line 1336:
```suggestion
resolved <a>DID document</a> MUST match the <a>DID</a> that was
resolved, or be populated with an equivalent, canonical <a>DID</> 
as specified by the related <a>DID method</a>.
```

brentzundel: I understood what you were saying and like the change, but the wording felt kind of clumsy, so this is my attempt to wordsmith.
Just a suggestion.

csuwildcat: Two things here I want to note:

1. Where it says `THE equivalent canonical DID`, was intentional to make sure that it was a singular relationship, where as a change to `an equivalent canonical DID` seems to suggest that it is selecting from many. I think it would be best to leave this `the`.
2. While I am fine with the general change of the wording to `as specified by the related DID method`, do you think we should use a different word that `related` here? I ask because this is about relationships between things, and maybe we should be more explicit that its not just the `related` method, its the Method of the DID being resolved.

brentzundel: good points, suggestion has been updated.

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
brentzundel, line 1: looks like you might have inadvertently added some personal IDE settings to the PR.

OR13: I wish we all used vs code and agreed to committed settings... like auto format or not.... every w3c repo i work it, i have a bunch of settings disabling my default behavior, like format on save... but yea, probably a mistake  to commit this file.

peacekeeper:
I cannot say exactly why, but I feel slightly uncomfortable with allowing the `id` property to contain anything other than the DID that was resolved.

Another solution to this use case (besides the idea of a `canonical` link relation discussed in https://github.com/w3c/did-core/issues/368) could be to return a null DID document, plus resolution metadata which contains the canonical DID that should be resolved instead. This would be comparable to HTTP's redirect status codes.

E.g. if you resolve `did:example:123`, the result of the `resolve()` function would be:

1. did-resolution-metadata = `«["redirect" → "did:example:456"]»`
2. did-document = `null`
3. did-document-metadata = `null`

A resolver could be configured to automatically follow such redirects, to make this transparent for the client.

csuwildcat:
I really think making some sort of canonical redirect from one DID to another based on having to generically follow data/directives in the resolution metadata to another DID/DID Document, then making the same inference in the end you would have this way, is going to require a ton more spec text across multiple specs (some outside of the DID Core spec) and will be prone to devs footgunning themselves with the complexity and ergonomic issues. The unidirectional nature of the canonical value language is the most objective/empirical of all the equivalence expressions we have discussed, so I really hope we can at least get this in. If not, I would hope the fallback is a specific property like `canonical` being added, because going through redirect hoops seems like a lot of complexity for very little benefit.

.vscode/settings.json
csuwildcat: Didn't realize I had, I can delete.

msporny:
I'll note that the last SHOULD in this statement is not testable.

Issue #364, opened 2020-07-30
Fill or remove section on DID URL Dereferencing


peacekeeper:
A section on [DID Resolution](https://w3c.github.io/did-core/#did-resolution) has been added (https://github.com/w3c/did-core/pull/331). A section on [Metadata Structure](https://w3c.github.io/did-core/#metadata-structure) has also been added (https://github.com/w3c/did-core/pull/347).

We also have an empty placeholder section on [DID URL Dereferencing](https://w3c.github.io/did-core/#did-url-dereferencing), which we should either fill, or remove. Let's discuss in this new issue.

If there is support for having this section, we should be able to re-use content from existing PRs such as https://github.com/w3c/did-core/pull/253/.

For example, an abstract DID URL dereferencing function could look like this (analogous to the `resolveStream` function):

```
dereferenceStream ( did-url, did-url-dereferencing-input-metadata ) ->
    ( did-url-dereferencing-metadata, content-stream, did-document-metadata )
```

wyc:
@peacekeeper how does your work found in the CCG relate to this section?
https://w3c-ccg.github.io/did-resolution/

I have been reviewing https://github.com/w3c/did-core/pull/253

peacekeeper:
@wyc the idea is that DID Core will only contain definitions of abstract functions with their inputs and outputs.  The `resolve()` function has already been added in https://w3c.github.io/did-core/#did-resolution. This issue is now about adding the `dereference()` function in a similar way.

DID Core however would leave out implementation details such as resolver architectures, protocols, metadata data formats, algorithms, etc.. The DID Resolution spec in the CCG is meant to be the place for such content that goes beyond the abstract function definition.

peacekeeper:
This was addressed by https://github.com/w3c/did-core/pull/381. Closing.

Issue #380, opened 2020-08-26
Why DID Document streaming (`resolveStream`)?


wyc:
Hi, I've been working on the DID URL Dereferencing section as per https://github.com/w3c/did-core/issues/364 and was also reviewing the [DID Resolution](https://w3c.github.io/did-core/#did-resolution) section for consistency. I noticed that there were two functions, `resolve` and `resolveStream`. To me, streaming is closer to an implementation detail, and I'm not sure that it's within our purview to require both modes. At this time I have to consider if DID URL Dereferencing needs streaming or not within the spec for my work. It seems to have been added here: https://github.com/w3c/did-core/commit/2da9671a4508a50b06e0e915833bd7470f5afb2d

Unfortunately I couldn't find any discussion, so I'm opening an issue.

1. Are there notes from the discussion preceding this decision?
2. Why did we want to require a stream function at all?
3. Can someone explain how it was decided that an implementer MUST implement both `resolve` and `resolveStream`? 

Here are some reasons I came up with that could make sense as to why things are they way they are:

- **Why have streaming at all in the spec?** DID Documents could get huge, and this could be future-proofing
- **Let's just assume we want streaming. Why are they two different functions?** We want conforming implementations to always be able to produce a DID document conforming to the data models listed, so therefore we cannot "bunch together" streaming and non-streaming responses, because this would allow terrible things (tm) like a DID method author returning some absurd binary packed message and telling a resolver to "go figure, I'm conforming to the spec because you can get a DID document out of this hot mess if you cast the right spells and incantations but I won't actually return one conformant to one of the data models described (JSON-LD/CBOR-LD/etc.) because I don't have to."

I feel that if we changed the definition of `did-document` into one that also accepted byte streams, then we could eliminate the latter abstract function along with `did-document-stream`. Here's some possible language to make that happen:

### A: Incorporate Streaming
> **did-document**
> If the resolution is successful, and if the resolve function was called, this MUST be a DID document conforming to the abstract data model or a byte stream which can be parsed by the caller of the resolve function into a DID document conforming to the abstract data model. Implementers MAY use a combination of `did-resolution-metadata` fields and "Read" operation of the DID method to indicate values which are byte streams. If the resolution is unsuccessful, this value MUST be empty.

This is scary for the reason above, that an implementation never actually needs to return a conforming data blob.

### B: Eliminate Streaming
> **did-document**
[...]

PR #381, opened 2020-08-26
Fill DID URL Dereferencing Section

wyc:
As per: https://github.com/w3c/did-core/issues/364

- Fill the DID URL Dereferencing section with abstract function, function
  parameters, return values, and their definitions, using language and format
  consistent to (and drawn from) the DID Resolution section.
- Delegate the handling of DID URLs which are also DIDs to the DID Resolution
  process, thereby ensuring that DID URL Dereferencing is a superset of DID
  Resolution, just as DID URLs are a superset of DIDs.

***
<a href="https://pr-preview.s3.amazonaws.com/wyc/did-core/pull/381.html" title="Last updated on Sep 8, 2020, 11:38 PM UTC (3ce1d0e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/381/8c47db8...wyc:3ce1d0e.html" title="Last updated on Sep 8, 2020, 11:38 PM UTC (3ce1d0e)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -3475,9 +3475,8 @@ content-stream
             </dt>
             <dd>
 If the <code>dereferencing</code> function was called and successful, this MUST
-contain a resource corresponding to the <a>DID URL</a>. 
-Especially in the case that the <a>DID URL</a> is a <a>DID</a>, the
-<code>content-stream</code> MAY be a <a>DID document</a> in one of the
+contain a resource corresponding to the <a>DID URL</a>.
+The <code>content-stream</code> MAY be a <a>DID document</a> in one of the
 conformant <a href="#representations">representations</a> obtained through
 the resolution process.
 
--- a/index.html
+++ b/index.html
@@ -3410,6 +3410,191 @@ property.
         <h2>
 DID URL Dereferencing
         </h2>
+        <p>
+The DID URL dereferencing function dereferences a <a>DID URL</a> into
+a resource with contents depending on the <a>DID URL</a>'s components,
+including the <a>DID method</a>, method-specific identifier, query parameters,
+and fragment. When a <a>DID URL</a> is also a <a>DID</a>, this process
+ performs <a>DID Resolution</a>. The details of how this
+process is accomplished are outside the scope of this specification, but all
+conformant implementations MUST implement two functions which have the
+following abstract forms:
+        </p>
+
+        <p><code>
+dereference ( did-url, did-url-dereferencing-input-metadata ) <br>
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;-&gt; ( did-url-dereferencing-metadata, content, content-metadata )
+        </code></p>
+
+        <p>
+The input variables of these functions MUST be as follows:
+        </p>
+
+        <dl>
+            <dt>
+did-url
+            </dt>
+            <dd>
+A conformant <a>DID URL</a> as a single string. This is the <a>DID URL</a> to dereference.
+This input is REQUIRED.
+            </dd>
+            <dt>
+did-url-dereferencing-input-metadata
+            </dt>
+            <dd>
+A <a href="metadata-structure">metadata structure</a> consisting of input
+options to the <code>dereference</code> function in addition to the <code>did-url</code>
+itself.
+Properties defined by this specification are in <a href="#did-url-dereferencing-input-metadata-properties"></a>.
+This input is REQUIRED, but the structure MAY be empty.
+            </dd>
+        </dl>
+
+        <p>
+The output variables of these functions MUST be as follows:
+        </p>
+
+        <dl>
+            <dt>
+did-url-dereferencing-metadata
+            </dt>
+            <dd>
+A <a href="metadata-structure">metadata structure</a> consisting of values
+relating to the results of the <a>DID URL Dereferencing</a> process. This
+structure is REQUIRED and MUST NOT be empty.
+Properties defined by this specification are in 
+<a href="#did-url-dereferencing-metadata-properties"></a>.
+If the dereferencing is not successful, this structure MUST contain an
+<code>error</code> property describing the error.
+            </dd>
+            <dt>
+content
+            </dt>
+            <dd>
+If the <code>dereferencing</code> function was called and successful, this MUST
+contain a resource corresponding to the <a>DID URL</a>.
+Additionally, in the case that the <a>DID URL</a> is a <a>DID</a>, then this
+MUST be a <code>did-document</code> as described in <a>DID Resolution</a>,
+obtained through the resolution process.
+
+If the dereferencing is unsuccessful, this value MUST be empty.
+            </dd>
+            <dt>
+content-metadata
+            </dt>
+            <dd>
+If the dereferencing is successful, this MUST be a <a href="metadata-structure">
+metadata structure</a>. This structure contains
+metadata about the <code>content</code>.  
+Additionally, in the case that the <a>DID URL</a> is a <a>DID</a>, then this
+MUST be a <code>did-document-metadata</code> as described in <a>DID Resolution</a>,
+obtained through the resolution process.
+
+If the dereferencing is unsuccessful, this output MUST be an empty <a href="metadata-structure">metadata structure</a>.
+            </dd>
+        </dl>
+
+        <p>
+<a>DID URL Dereferencing</a> implementations MUST NOT alter the signature of
+these functions in any way. <a>DID URL Dereferencing</a> implementations MAY
+map the <code>dereference</code> function to a method-specific internal
+function to perform the actual <a>DID URL Dereferencing</a> process. <a>DID
+URL Dereferencing</a> implementations MAY implement and expose additional
+functions with different signatures in addition to the <code>dereference</code>
+function specified here.
+        </p>
+
+        <section>
+            <h3>
+DID URL Dereferencing Input Metadata Properties
+            </h3>
+
+            <p>
+The possible properties within this structure and their possible values are defined by [[DID-SPEC-REGISTRIES]].
+This specification defines the following common properties.
+            </p>
+
+            <dl>
+                <dt>
+accept
+                </dt>
+                <dd>
+The MIME type the caller prefers for <code>content</code>. The <a>DID URL
+Dereferencing</a> implementation SHOULD use this value to determine the
+representation contained in the returned value if such a representation is
+supported and available. This property is OPTIONAL.
+                </dd>
+            </dl>
+        </section>
+
+        <section>
+            <h3>
+DID URL Dereferencing Metadata Properties
+            </h3>
+
+            <p>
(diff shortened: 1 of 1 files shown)


wyc:
Note that I have not added a separate abstract function to handle "streaming" dereferencing in the same fashion as DID Resolution's `resolveStream` because specification of the proper content type should also include support for byte streams. See @peacekeeper's compendium of discussions leading up to this in https://github.com/w3c/did-core/issues/380 for an explanation of why this is the case for DID Resolution. 

Also, note that `content-type` is required in the output metadata to maintain symmetry with `resolveStream`, especially in the case of delegation to the DID Resolution process if the DID URL is a DID. I personally think we should consider following HTTP's precedent, in which Content-Type is [not mandated](https://www.w3.org/Protocols/rfc2616/rfc2616-sec7.html#sec7.2.1) and has a recommended assumed type if absent:
> Any HTTP/1.1 message containing an entity-body SHOULD include a Content-Type header field defining the media type of that body. If and only if the media type is not given by a Content-Type field, the recipient MAY attempt to guess the media type via inspection of its content and/or the name extension(s) of the URI used to identify the resource. If the media type remains unknown, the recipient SHOULD treat it as type "application/octet-stream".

This would simplify many minimal viable implementations of our standard, if metadata sections upon success could be empty.

wyc:
Furthermore, should we resolve this PR, I would propose in a new issue that we change the ordering of DID Resolution abstract function outputs from `(did-resolution-metadata, did-document, did-document-metadata)` to `(did-document, did-document-metadata, did-resolution-metadata)`.

For now, I have followed suit in this PR to maintain consistency with DID Resolution, but this proposal would then change DID URL Dereferencing abstract function outputs from `(did-url-dereferencing-metadata, content, content-metadata)` to `(content, content-metadata, did-url-dereferencing-metadata)`.

I propose this simply because I believe a user expects to receive function outputs in a descending order of importance, and I think we can agree that in general for DID Resolution the most important output is the DID document, and for DID URL Dereferencing, it is the content.

index.html
```diff
@@ -3410,6 +3410,191 @@ <h3>
         <h2>
 DID URL Dereferencing
         </h2>
+        <p>
+The DID URL dereferencing function dereferences a <a>DID URL</a> into
+a resource with contents depending on the <a>DID URL</a>'s components,
+including the <a>DID method</a>, method-specific identifier, query parameters,
```
peacekeeper, line 3416:
```suggestion
including the <a>DID method</a>, method-specific identifier, path, query,
```
Propose to 1. include path, since that is also part of the DID URL, and 2. not assume here that the query always contains key=value parameters, even though that's what we mostly expect (see our discussion in https://github.com/w3c/did-core/issues/385).

index.html
```diff
@@ -3410,6 +3410,191 @@ <h3>
...
+and fragment. When a <a>DID URL</a> is also a <a>DID</a>, this process
+ performs <a>DID Resolution</a>. The details of how this
```
peacekeeper, line 3418:
```suggestion
and fragment. This process depends on <a>DID resolution</a>
of the <a>DID</a> contained in the <a>DID URL</a>.
The details of how this
```

Event though technically I agree with this, it's a bit too concrete for my taste here, remember we wanted to only do abstract function definitions here, not describe the concrete algorithm behind them.

index.html
```diff
@@ -3410,6 +3410,191 @@ <h3>
...
+                </dt>
+                <dd>
+The error code from the dereferencing process.
+This property is REQUIRED when there is an error in the dereferencing process.
+The value of this property is a single keyword string.
+The possible property values of this field are defined by [[DID-SPEC-REGISTRIES]].
+This specification defines the following error values:
+                    <dl>
+                        <dt>
+invalid-url
```
peacekeeper, line 3559:
```suggestion
invalid-did-url
```

index.html
```diff
@@ -3410,6 +3410,191 @@ <h3>
...
+A <a href="metadata-structure">metadata structure</a> consisting of values
+relating to the results of the <a>DID URL Dereferencing</a> process. This
+structure is REQUIRED and MUST NOT be empty.
+Properties defined by this specification are in 
+<a href="#did-url-dereferencing-metadata-properties"></a>.
+If the dereferencing is not successful, this structure MUST contain an
+<code>error</code> property describing the error.
+            </dd>
+            <dt>
+content
```
peacekeeper, line 3471: I propose to rename this to `content-stream`. In my opinion, `dereference()` is more like `resolveStream()` than `resolve()`, since dereferencing returns a concrete representation of a resource. Would you agree or do you see that differently?

If you agree, then you could also add a sentence to the "did-url-dereferencing-metadata" item that says that it MUST contain a `content-type` property (similar to how we say that in the "did-resolution-metadata" item).

index.html
```diff
@@ -3410,6 +3410,191 @@ <h3>
...
+            </dt>
+            <dd>
+If the <code>dereferencing</code> function was called and successful, this MUST
+contain a resource corresponding to the <a>DID URL</a>.
+Additionally, in the case that the <a>DID URL</a> is a <a>DID</a>, then this
+MUST be a <code>did-document</code> as described in <a>DID Resolution</a>,
```
peacekeeper, line 3477:
```suggestion
The content MAY be a <a>DID document</a> in one of the conformant <a href="#core-representations">.
```
Similar to one of my earlier comments, I would prefer slightly more abstract language rather than a concrete statement about what the result will be. For example, there are DID URLs which contain more than just DIDs and still return a DID document. I think we don't want so many details on what exactly the function returns in what situations (this should go into the DID Resolution spec, see [this algorithm](https://w3c-ccg.github.io/did-resolution/#dereferencing-algorithm)).

peacekeeper (changes requested):
Hey @wyc I think this is great work! I only have a few smaller suggestions, see inline. And maybe one "bigger" comment, which is that I think `dereference()` returns a `content-stream,` i.e. a byte stream with a `content-type`, similar to the `resolveStream()` function. Let me know what you think.

peacekeeper:
> I would propose in a new issue that we change the ordering of DID Resolution abstract function outputs

I don't feel too strongly about this, but I think @jricher and others would argue that you typically want the `did-resolution-metadata` or `did-url-dereferencing-metadata` first, because you first need the `content-type` before you can do anything with the `did-document-stream` or `content-stream`....

index.html
wyc: +1

index.html
wyc: Do you recommend we use this PR to move the MUST to SHOULD in `resolveStream`, `resolve`, and `dereference`? I will start on this path unless objections arise.

peacekeeper: I think I'd propose that we make it a SHOULD here for `dereference` (following your suggestion), and then create a separate issue later to discuss whether or not we should do the same for `resolveStream`. I think I remember that back then there was already some discussion about this, but I can't find any references now.

wyc: Makes sense, and I think it's a good idea. I'll incorporate these changes.

msporny:
Could you please do a re-review @peacekeeper, looks like @wyc did a number of changes. It would be good for @jricher to take a look as well to make sure it matches his expectations for this section.

peacekeeper:
I'm mostly fine with this PR, except for the minor topics we still had in the open thread above https://github.com/w3c/did-core/pull/381#discussion_r482941868.

@wyc do you want to fix this?

wyc:
I will be making those improvements this week.

wyc:
@peacekeeper I have made a series of changes to address our discussion in the thread above, ready for final review. Upon close of this PR, I will create new issues to:

- Discuss reordering of return values from our abstract function calls to put the content/document first
- Change the MUST to SHOULD wrt content-type in DID Resolution

index.html
```diff
@@ -3462,20 +3463,23 @@ <h2>
...
             <dd>
 If the <code>dereferencing</code> function was called and successful, this MUST
-contain a resource corresponding to the <a>DID URL</a>.
-The content MAY be a <a>DID document</a> in one of the conformant <a href="#representations">representations</a>.
-obtained through the resolution process.
+contain a resource corresponding to the <a>DID URL</a>. 
+Especially in the case that the <a>DID URL</a> is a <a>DID</a>, the
+<code>content-stream</code> MAY be a <a>DID document</a> in one of the
+conformant <a href="#representations">representations</a> obtained through
+the resolution process.
```
peacekeeper, line 3482:
```suggestion
contain a resource corresponding to the <a>DID URL</a>.
The <code>content-stream</code> MAY be a <a>DID document</a> in one of the
conformant <a href="#representations">representations</a> obtained through
the resolution process.
```

As discussed before, even though I agree with your text, I think it's better if we avoid concrete statements about what exactly the DID URL dereferences to in which cases.

index.html
```diff
@@ -3484,11 +3488,11 @@ <h2>
...
             <dd>
 If the dereferencing is successful, this MUST be a <a href="metadata-structure">
-metadata structure</a>. This structure contains
-metadata about the <code>content</code>.  
-Additionally, in the case that the <a>DID URL</a> is a <a>DID</a>, then this
-MUST be a <code>did-document-metadata</code> as described in <a>DID Resolution</a>,
-obtained through the resolution process.
+metadata structure</a>, but the structure MAY be empty. This structure contains
+metadata about the <code>content-stream</code>.  
+Especially in the case that the <a>DID URL</a> is a <a>DID</a>, this MAY be a
```
peacekeeper, line 3493:
```suggestion
If the <code>content-stream</code> is a <a>DID document</a>, this MUST be a
```

I think this is a more reliable way of describing the same idea..

peacekeeper:
Thanks for the update, I only have two more suggestions (see above), then from my perspective we're good to go..

index.html
wyc: Thanks, agreed!

wyc:
@peacekeeper I agreed with both your suggestions and found them to be reasonable, merged.

peacekeeper (approved):
Looks good to me now, thanks for all your work!

Let's give others some time to review as well.

kdenhartog (approved):
Great work @wyc thanks for adding this part to the spec!

index.html
```diff
@@ -3410,6 +3410,194 @@ <h3>
...
+
+            <dl>
+                <dt>
+accept
+                </dt>
+                <dd>
+The MIME type the caller prefers for <code>content-stream</code>. The <a>DID URL
+Dereferencing</a> implementation SHOULD use this value to determine the
+representation contained in the returned value if such a representation is
+supported and available. This property is OPTIONAL.
```
msporny, line 3528: wow... I don't think I've ever come across a spec with the OPTIONAL RFC keyword in it... we usually do "If provided..." -- keeping this in there for now... OPTIONAL doesn't result in test suite language.

msporny (approved):
Normative, multiple positive reviews, requested changes made, no objections, merging.

index.html
peacekeeper: It's also in the (parallel) section of DID Resolution section: https://w3c.github.io/did-core/#did-resolution-input-metadata-properties

PR #388, opened 2020-09-03
Address Unknown Properties in JSON-LD

OR13:
### allow `@context` to include `@base` 

If we are allowed to use JSON-LD, might as well actually use it. (allow JSON-LD developers to use JSON-LD features)

### Add SHOULD language to JSON-LD Producer section

Allow JSON-LD Producers to generate JSON-LD that contains unknown properties. (99% of did methods currently do this)

### Add SHOULD language to JSON-LD Consumer section

Allow JSON-LD Consumers to drop unknown properties. (universal resolver currently does this)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/388.html" title="Last updated on Sep 17, 2020, 12:50 PM UTC (66dc66d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/388/1547e0a...66dc66d.html" title="Last updated on Sep 17, 2020, 12:50 PM UTC (66dc66d)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2526,7 +2526,7 @@ The value of the <code>@context</code> property MUST be exactly one of the follo
   </li>
   <li>An array of one or more elements,
     where the value of the first element is
-    <code>https://www.w3.org/ns/did/v1</code>.
+    a string with the value of <code>https://www.w3.org/ns/did/v1</code>.
 
     <pre class="example">
       {
--- a/index.html
+++ b/index.html
@@ -2379,6 +2379,11 @@ the root object. Properties MAY define additional data sub structures subject
 to the value representation rules in the list above.
         </p>
 
+        <p>
+The member name <code>@context</code> MUST NOT be used as this property is
+reserved for JSON-LD producers.
+        </p>
+
 
       </section>
 
--- a/index.html
+++ b/index.html
@@ -2525,7 +2525,7 @@ The value of the <code>@context</code> property MUST be exactly one of the follo
     </pre>
   </li>
   <li>An array of one or more elements,
-    where the value of the first <a>URI</a> is
+    where the value of the first element is
     <code>https://www.w3.org/ns/did/v1</code>.
 
     <pre class="example">
--- a/index.html
+++ b/index.html
@@ -2379,10 +2379,6 @@ the root object. Properties MAY define additional data sub structures subject
 to the value representation rules in the list above.
         </p>
 
-        <p>
-The member name <code>@context</code> MUST NOT be used as this property is
-reserved for JSON-LD producers.
-        </p>
 
       </section>
 
@@ -2517,25 +2513,66 @@ property.
         <dl>
           <dt>@context</dt>
           <dd>
-The value of the <code>@context</code> property MUST be one or more <a>URIs</a>,
-where the value of the first <a>URI</a> is
-<code>https://www.w3.org/ns/did/v1</code>. All members of the
+The value of the <code>@context</code> property MUST be exactly one of the following:
+
+<ul>
+  <li>The string <code>https://www.w3.org/ns/did/v1</code>. 
+    <pre class="example">
+      {
+        "@context": "https://www.w3.org/ns/did/v1",
+        "id": "did:example:123"
+      }      
+    </pre>
+  </li>
+  <li>An array of one or more <a>URIs</a>,
+    where the value of the first <a>URI</a> is
+    <code>https://www.w3.org/ns/did/v1</code>.
+
+    <pre class="example">
+      {
+        "@context": [
+          "https://www.w3.org/ns/did/v1",
+          "https://example.com/blockchain-identity/v1"
+        ],
+        "id": "did:example:123"
+      }      
+    </pre>
+  </li>
+  <li>An array where the first value is
+    <code>https://www.w3.org/ns/did/v1</code>, and the final value is an object.
+
+    <pre class="example">
+      {
+        "@context": [
+          "https://www.w3.org/ns/did/v1",
+          "https://example.com/blockchain-identity/v1",
+          {
+            "@base": "did:example:123"
+          }
+        ],
+        "id": "did:example:123"
+      }      
+    </pre>
+  </li>
+</ul>
+
+All members of the
 <code>@context</code> property SHOULD exist in the DID specification
 registries in order to achieve interoperability across different
-representations. If a member does not exist in the DID specification
+representations. 
+
+If a member does not exist in the DID specification
 registries, then the DID Document will not be interoperable across
 representations.
+
+<p>
+  Producers SHOULD not produce DID Documents that 
+  contain properties not defined by the <code>@context</code>. 
+  Such properties MAY be dropped by Consumers.
+</p>
+
           </dd>
         </dl>
-
-        <p class="issue" data-number="202">
-This specification defines globally interoperable documents, and the requirement
-that the context value be in the verifiable data registry means that different JSON-LD
-processors can consume the document without having to dereference anything in
-the context field. However, a pair of producers and consumers can have local
-extension agreements. This needs to be clarified in a section on extensibility
-and linked here when available.
-        </p>
 
       </section>
 
@@ -2554,9 +2591,10 @@ interpreted using JSON-LD processing under the rules of the defined
         <dl>
           <dt>@context</dt>
           <dd>
-The value of the <code>@context</code> property MUST be one or more <a>URIs</a>,
-where the value of the first <a>URI</a> is
-<code>https://www.w3.org/ns/did/v1</code>. If more than one <a>URI</a> is
+The value of the <code>@context</code> property MUST conform 
+to the <a href="#production-0">JSON-LD Production Rules</a>. 
+
+If more than one <a>URI</a> is
 provided, the <a>URIs</a> MUST be interpreted as an ordered set. It is
 RECOMMENDED that dereferencing each <a>URI</a> results in a document containing
 machine-readable information about the context. To enable interoperability
@@ -2580,6 +2618,11 @@ unknown data members on consumption, and this section needs to be updated with
 that decision.
         </p>
 
+        <p>
+          Consumers SHOULD drop all properties from a DID Documents that 
+          are not defined by the <code>@context</code>. 
+        </p>
+ 
       </section>
 
     </section>
--- a/index.html
+++ b/index.html
@@ -2524,7 +2524,7 @@ The value of the <code>@context</code> property MUST be exactly one of the follo
       }      
     </pre>
   </li>
-  <li>An array of one or more <a>URIs</a>,
+  <li>An array of one or more elements,
     where the value of the first <a>URI</a> is
     <code>https://www.w3.org/ns/did/v1</code>.
 


index.html
```diff
@@ -2517,26 +2513,67 @@ <h3>
...
+      {
+        "@context": [
+          "https://www.w3.org/ns/did/v1",
+          "https://example.com/blockchain-identity/v1"
+        ],
+        "id": "did:example:123"
+      }      
+    </pre>
+  </li>
+  <li>An array where the first value is
```
tplooker, line 2541: I'm probably missing something but can you elaborate as to why this explicit variation of the `@context` value has to be noted in the spec? For instance why it is not sufficient to purely state that the did document `@context` value is either a string where the value is `https://www.w3.org/ns/did/v1` or an array where the first member of the array MUST be a string with the value `https://www.w3.org/ns/did/v1` and impose no further constrains around what the other elements in the array must be.

OR13: I suppose it could say, an array with first value equal to `https://www.w3.org/ns/did/v1`... I tend to favor explicit over implicit definitions, hence explicit support for the cases I have seen in the wild.

index.html
```diff
@@ -2580,6 +2618,11 @@ <h3>
 that decision.
         </p>
 
+        <p>
+          Consumers SHOULD drop all properties from a DID Documents that 
+          are not defined by the <code>@context</code>.
```
selfissued, line 2623: Could we modify this to "Consumers SHOULD drop all properties from DID Documents that are not defined by `@context` or that they do not understand."?

index.html
```diff
@@ -2517,26 +2513,67 @@ <h3>
...
-representations. If a member does not exist in the DID specification
+representations. 
+
+If a member does not exist in the DID specification
 registries, then the DID Document will not be interoperable across
 representations.
+
+<p>
+  Producers SHOULD not produce DID Documents that 
+  contain properties not defined by the <code>@context</code>.
```
selfissued, line 2570: Could we modify this to "Producers SHOULD not produce DID Documents that contain properties not defined by the `@context` unless they have an expectation that some consumers will understand them"?

selfissued (changes requested):
I mostly really like this, as I think it brings us closer to a unified model.  I made two wording suggestions that I think will even better align the JSON-LD binding with the JSON binding.

index.html
OR13: I cannot test for "do not understand" I can test for absence of properties in the context.

I assume that changing this to a MUST, would draw an even large distinction between JSON and JSON-LD Consumption. Given that a consumer can request content as `application/did+json` if they want nothing dropped, I'd be comfortable with a MUST here as well.

I would also be comfortable saying no properties are dropped and `@context` is required, and erasing the distinction between JSON-LD and JSON-LD with unknown properties :)

index.html
OR13: Same comment as above, I cannot test this statement.

index.html
OR13: @tplooker I tried rewording this, and it always ended up dangerously vague. Feel free to suggest changes you think will provide better clarity. 

I would prefer to resolve this conversation and keep my suggested changes to the requirements associated with `@context`, can I take that action?

index.html
```diff
@@ -2517,26 +2513,67 @@ <h3>
...
+<ul>
+  <li>The string <code>https://www.w3.org/ns/did/v1</code>. 
+    <pre class="example">
+      {
+        "@context": "https://www.w3.org/ns/did/v1",
+        "id": "did:example:123"
+      }      
+    </pre>
+  </li>
+  <li>An array of one or more <a>URIs</a>,
```
tplooker, line 2527:
```suggestion
  <li>An array of one or more elements,
```

index.html
```diff
@@ -2517,26 +2513,67 @@ <h3>
...
+    where the value of the first <a>URI</a> is
```
tplooker, line 2528:
```suggestion
    where the value of the first element is
```

index.html
```diff
@@ -2517,26 +2513,67 @@ <h3>
...
+    <code>https://www.w3.org/ns/did/v1</code>.
```
tplooker, line 2529:
```suggestion
    a string with the value of <code>https://www.w3.org/ns/did/v1</code>.
```

index.html
tplooker: @OR13 just made a suggestion above, let me know what you think but if that works then I think this section can be removed?

tplooker (commented):
Made a concrete inline suggestion w.r.t my comment, let me know if doesn't make sense

index.html
```diff
@@ -2517,26 +2513,67 @@ <h3>
...
+<ul>
+  <li>The string <code>https://www.w3.org/ns/did/v1</code>. 
+    <pre class="example">
+      {
+        "@context": "https://www.w3.org/ns/did/v1",
+        "id": "did:example:123"
+      }      
+    </pre>
+  </li>
+  <li>An array of one or more elements,
```
tplooker, line 2527: Another idea, how about something along the lines of following, need to make sure we refer to the JSON types correctly though

```
Further to this how about the following

```
An array of one or more elements, where the value types of the array members MUST either be a string or an object and the first element of the array is the string <code>https://www.w3.org/ns/did/v1</code>
```
```

index.html
```diff
@@ -2517,26 +2513,67 @@ <h3>
...
+    where the value of the first element is
+    a string with the value of <code>https://www.w3.org/ns/did/v1</code>.
```
tplooker, line 2529:
```suggestion
  <li>An array of one or more elements,
    where the value types of the members of the array
    MUST be either a string or an object and the first element is
    the string <code>https://www.w3.org/ns/did/v1</code>.
```

tplooker (commented):
Another minor suggestion of language around the array case

index.html
```diff
@@ -2517,26 +2513,68 @@ <h3>
...
+  <li>An array where the first value is
+    <code>https://www.w3.org/ns/did/v1</code>, and the final value is an object.
+
+    <pre class="example">
+      {
+        "@context": [
+          "https://www.w3.org/ns/did/v1",
+          "https://example.com/blockchain-identity/v1",
+          {
+            "@base": "did:example:123"
```
msporny, line 2551: Hmm... this is a problem due to this section: https://w3c.github.io/did-core/#relative-did-urls

We may have to note that different representations may impose their own Base URI resolution mechanisms (like JSON-LD) and so those run after the generalized Base URI algorithm is run. I think that's the behavior we want. 

Otherwise, this statement above is going to be ignored... and honestly, I'd argue that the use of `@base` is something we do not want to suggest that people do. If we're going to have examples like this in here, we need to say that reprepsentation-specific base processing happens after the Base URI establishment in Section 3.2.5 today.

msporny (changes requested):
Overall, LGTM. The `@base` example raises an issue we need to address in general. I'm fine with it being addressed in another PR... but at present, `@base` in JSON-LD will get ignored per the spec.

peacekeeper:
So if we allow the JSON representation (`application/did+json`) to include a `@context`, which will be ignored, will this not lead to dangerous situations where a difference in the context will be interpreted correctly if `content-type` is `application/did+ld+json`, but the difference will be missed if the `content-type` is `application/did+json`?

E.g. Document 1 (includes an extra `@context` which defines `ethereumAddress`):

```
{
  "@context":[
    "https://identity.foundation/EcdsaSecp256k1RecoverySignature2020/lds-ecdsa-secp256k1-recovery2020-0.0.jsonld"
  ],
  "id":"did:example:123",
  "publicKey":[{
    "id": "did:example:123#vm-3",
    "type": "EcdsaSecp256k1RecoveryMethod2020",
    "ethereumAddress": "0xF3beAC30C498D9E26865F34fCAa57dBB935b0D74"
  }]
}
```

E.g. Document 2 (only the `@context` is different):

```
{
  "@context":[
    "https://someothersource.com/other-definitions.jsonld"
  ],
  "id":"did:example:123",
  "publicKey":[{
    "id": "did:example:123#vm-3",
    "type": "EcdsaSecp256k1RecoveryMethod2020",
    "ethereumAddress": "0xF3beAC30C498D9E26865F34fCAa57dBB935b0D74"
  }]
}
```

A JSON consumer will think both documents are the same. A JSON-LD consumer will understand the difference. This can't really be what we want?

I think last time the group proposed that JSON-LD DID documents can be processed with JSON consumers as well, we assumed that at a minimum the JSON consumers would have to check that the URIs in @context match what they expect (e.g. see https://github.com/w3c/did-core/pull/142#issuecomment-565077587). Now we want to allow JSON consumers to simply drop `@context`?

peacekeeper:
My feeling is, now that we went through a long process that led to the abstract data model, multiple representations, and the registries, we should not blur the lines between representations. We should not say too much about how to process JSON-LD with JSON consumers, or the other way round.

A DID document in `application/did+ld+json` MUST have a `@context`.

A DID document in `application/did+json` MUST NOT have a `@context`. Why change this rule?

index.html
OR13: I don't think there is anything wrong with using `@base`, I have written tests that show that it works fine for DID Document's and VCs, and per the JSON-LD spec, it is allowed... https://json-ld.org/spec/latest/json-ld/#using-the-document-base-as-the-default-vocabulary

did:ion,  did:elem, and our implementation did:key all use `@base`... in my opinion, it makes JSON-LD DID Document representations much more aesthetically pleasing, and it produces identical N-Quads to the much more repetitive fully qualified URI representations.

`@base` does not get ignored which you compact a JSON-LD DID Document to n-quads using off the shelf tooling.

Here is an example which you can explore using the JSON-LD Playground https://json-ld.org/playground/:

```
{
  "@context": [
    {
      "@base": "did:key:z6Mkf5rGMoatrSj1f4CyvuHBeXJELe9RPdzo2PKGNCKVtZxP"
    }
  ],
  "id": "did:key:z6Mkf5rGMoatrSj1f4CyvuHBeXJELe9RPdzo2PKGNCKVtZxP",
  "publicKey": [
    {
      "id": "#z6Mkf5rGMoatrSj1f4CyvuHBeXJELe9RPdzo2PKGNCKVtZxP",
      "controller": "did:key:z6Mkf5rGMoatrSj1f4CyvuHBeXJELe9RPdzo2PKGNCKVtZxP",
      "publicKeyBase58": "dbDmZLTWuEYYZNHFLKLoRkEX4sZykkSLNQLXvMUyMB1"
    }
  ],
  "authentication": ["#z6Mkf5rGMoatrSj1f4CyvuHBeXJELe9RPdzo2PKGNCKVtZxP"],
  "assertionMethod": ["#z6Mkf5rGMoatrSj1f4CyvuHBeXJELe9RPdzo2PKGNCKVtZxP"],
  "capabilityDelegation": ["#z6Mkf5rGMoatrSj1f4CyvuHBeXJELe9RPdzo2PKGNCKVtZxP"],
  "capabilityInvocation": ["#z6Mkf5rGMoatrSj1f4CyvuHBeXJELe9RPdzo2PKGNCKVtZxP"],
  "keyAgreement": [
    {
      "id": "#z6LScqmY9kirLuY22G6CuqBjuMpoqtgWk7bahWjuxFw5xH6G",
      "type": "X25519KeyAgreementKey2019",
      "controller": "did:key:z6Mkf5rGMoatrSj1f4CyvuHBeXJELe9RPdzo2PKGNCKVtZxP",
      "publicKeyBase58": "2AbNdSuzFSpGvsiSPBfnamcKzk9Q3WRRpY2EToHZEuKW"
    }
  ]
}
```

OR13:
@peacekeeper because in practice JSON.parse will succeed on DID Documents which contain an `@context` and developers will, with high probability, produce such documents and return them when `application/did+json` is requested.... 99% of DID Methods I have tested produce JSON with an `@context` that contains unknown properties... I am a bayesian, I believe the future looks like the past... I have low expectations that DID Method authors are going to change this behavior, I tend to plan for what I think will happen, not what I want to happen.

### Option 1

If we want JSON and JSON-LD representations to be close, allowing `@context` and not dropping properties is one path to keeping them close.

### Option 2

If we want JSON and JSON-LD representation to not be close, requiring that they never be the same and that unknown properties get dropped from JSON-LD seems like a path forward... 

Regardless of which path the DID WG picks...I bet JSON developers still return DID Documents with an `@context`... especially when the spec says it MUST be ignored by JSON consumers...

The language we use for representations should be consistent... if we say something is JSON and allows arbitrary unknown properties to be present and ignored... but then forbid a property which is present in JSON-LD... thats terrible developer experience... 

I'm a JSON developer, if I want to return JSON with an `@context` and I know its ok to do so, because the spec says so... it says `@context` MUST be ignored by JSON consumers.... I plan to return `@context` in all my JSON DID Documents... I plan for all my DID Documents to be both valid JSON and valid JSON-LD... I am capable of producing both, regardless of what DID Core says... and currently... despite what did core says....

Why is the DID Core spec forcing developers to delete a property that consumers are REQUIRED to ignore? and more importantly, what happens if producers ignore what the did core spec says today?... nothing, except that a producer can now produce a document which is both valid JSON and valid JSON-LD AND you can tell which it is, because we have `content-type`.

Lets let JSON be JSON, and JSON-LD be JSON-LD.

But this isn't a hill I need to die on, especially since I know that no matter what we put in the did core spec, JSON.parse will continue to work on strings which contain an `@context` :)

 As a did method author I can produce documents that are valid JSON and JSON-LD, and simply note that I chose not to be did core spec conformant with that particular normative statement.... paraphrased.. "JSON DID Documents MUST NOT be the equal to JSON-LD DID Documents"... I don't agree that this is a good thing for us to say, and I would be happy for all my DID Methods to fail the test associated with that statement.

index.html
```diff
@@ -2379,10 +2379,6 @@ <h3>Production</h3>
 to the value representation rules in the list above.
         </p>
 
-        <p>
-The member name <code>@context</code> MUST NOT be used as this property is
```
jricher, line 2383: Removing this is problematic. The prohibition is to prevent determining the document type by deep inspection after parsing. In other words, send JSON LD as if it were just JSON and expect it to function. We should not allow plain JSON to have this member.

jricher (changes requested):
This goes too far in making the JSON and JSON LD indistinguishable from each other. We need to take steps to prohibit sloppy production and assumptions of lazy consumption of the data formats if we want any hope of interoperability.

index.html
OR13: this prohibition _enables_ content type sniffing... it enables a `application/did+json` to be sniffed to be guessed to be also JSON-LD... if we don't want folks to guess content type by sniffing JSON, we should make it not possible to do that... with normative statements... removing this statement prevents developers from sniffing content to determine content type....

Developers should be using `content-type` and ignoring unknown properties when requesting `application/did+json`... the consumer logic relies on  this... the producer logic should behave the same.

@jricher would you say you are generally of the opinion we should take `option 1` or `option 2`?

OR13: btw I removed this in an attempt to harmonize unknown properties, by making JSON and JSON-LD documents potentially identical, and relying on `content-type` to distinguish them... I am not married to that approach, just trying to get unknown properties to be treated consistently across representations... seems like that door is potentially closing.

jricher: I'm in favor of harmonization but this prohibition is not the same as an ignored property. If the property shows up in a `application/json` document then that's an error, it's not allowed to get dropped or passed through.

OR13: The consumption section assumes this property WILL show up AND that it will be present instead of an error... 

https://w3c.github.io/did-core/#consumption

Screen Shot 2020-09-10 at 7 48 51 AM

It's not `application/json` its `application/did+json`.... and I've tested this....  JSON.stringify / JSON.parse will work with the property, and no "error" is thrown.

If a JSON Producer produces JSON that contains 2 properties `@context` and `shamefullUnknownProperty`... what is the exact error that that is thrown by JSON.stringify / JSON.parse?... if you are asserting that somehow DID Core should define an error for "JSON that contains an `@context`" and that error should be thrown instead of producing a did document when `application/did+json` is requested, thats the same as not supporting that representation.... do we want to start encouraging folks to throw errors when a representation they don't support is requested?

Given the Producer is responsible for producing... I would expect the only place you can throw an error would be on the consumer side... 

Are you saying the JSON Consumer section has normative statements about a format that is not possible to produce?

@jricher can you answer my question regarding `option 1` / `option 2`?

jricher: @OR13 this is part of why I said we need to define "ignore" more precisely here. If "ignore" means "pass through to the model" then it shouldn't be ignore in the consumer, it should be "prohibit". If we're saying "ignore" means "drop from the model" then ignore is functionally the same in many cases, but I still think this should be an error case. 

I don't like either option1 or option2 but agree that the group needs to pick :some: path forward. I don't think that JSON and JSONLD need to look exactly like each other, or even should. I'll repeat what I've said before:

1. An "unknown" property can be carried through to the model as a raw data structure with no indication of what it is or validation of its formats. It can be queried as a raw value but does not appear in the DID document model as a known named property.
2. An "unknown" property can be dropped from the resulting model.
3. An "unknown" property can cause an error.

Specific named properties can have all of those behaviors for them as well.

OR13: @jricher sounds like you prefer JSON and JSON-LD to be prohibited from being equivalent.... I don't agree :)

As a JSON and JSON-LD developer, I would find it useful to ask for `application/did+json` when I wanted 0 properties to be dropped, and to ask for `application/did+ld+json` when I wanted properties not defined in the context to be dropped.  

I agree that we need to define what an "unknown" property is, here goes.

1. In JSON DID Documents there is no such thing as an "unknown" property.
2. In JSON-LD DID Documents, an unknown property is any property not defined in the `@context`.
3. Because the abstract data model MUST support all representations, it follows that there is no such thing as an unknown property in the abstract data model.... hence the abstract data model is essentially equivalent to JSON / Infra.

JSON developers should be able to add arbitrary JSON to their DID Document, and it should be passed through producers / consumers.... for a JSON developer `@context` is just another arbitrary property.

Now consider the case there where JSON -> JSON-LD -> JSON.

Depending on the rules associated with Producing / Consuming... properties might get dropped.

What this PR says is that JSON MAY contain an `@context`... if it does, and it covers 100% of the properties (if its done correctly), the JSON in is the same as the JSON out. (allow skilled developers to make JSON and JSON-LD representations that are valid and equivalent, and trivially bi-directionally convertible).

Now consider the case where it's done poorly (misses some properties)....

The JSON output would lack those properties if we agree that JSON-LD should drop them.... this PR says that MAY happen.

Now consider the case we have today where JSON does not have an `@context`... any properties not defined in the registry are dropped (because JSON-LD, not because of JSON logic)... and in practice (without substantial engineering) all the JSON properties would be treated as "unknown".... if the did core context in injected (questionable from a security perspective), then did core properties are preserved... if a developer writes the code to find all properties defined in the registries, and the JSON developer including properties both in and outside the registries, then only the properties defined outside of the registries would be dropped... but this now assuming a LOT of software has been written to match terms in JSON to registry extensions, and it actually assumes that those terms can only be defined in the registries.... which we know is not true for both JSON and JSON-LD, from an extensibility perspective.

This is the problem I am attempting to fix with this PR.

When a Producer makes JSON, it's something which can be `JSON.parse`ed.
When a Consumer consumes JSON, they call `JSON.parse`.

The extensibility mechanism of JSON is that is allows "unknown properties"... thats a feature.

When a Producer makes JSON-LD, it's something which can be `JSON.parse`ed.... 99% of JSON-LD Producers include terms not defined in the `@context`... probably because they are used to JSON, in other words. JSON-LD Producers produce JSON with terms not defined in the `@context`... this is a fact. We can't write spec text to change it.

When a Consumer consumes JSON-LD, they call `JSON.parse` and then one of 2 things happens.

1. They drop unknown properties
2. They forward the JSON as is (no JSON-LD processing yet).

Dropping "unknown properties" is considered a security feature, in much the same way that you probably wouldn't sign a blank check if I asked you too.

The extensibility mechanism for JSON-LD is adding to the `@context`.

As a developer who uses JSON and JSON-LD regularly to create DID Methods and work with JOSE and Linked Data based Verifiable Credentials, I try my best to make valid JSON-LD... but I will be the first to admit, its easier to make valid JSON.

I would love if when trying to integrate with a DID Method I could actually use JSON to work around the failings of a developer who had not yet gotten JSON-LD correct.... and if they never did... I would never ask for JSON-LD from them.... For example, this `did:sov` in the universal resolver today:

```
{
  "@context": "https://www.w3.org/2019/did/v1",
  "id": "did:sov:builder:VbPQNHsvoLZdaNU7fTBeFx",
  "service": [],
  "authentication": [
    {
      "type": "Ed25519SignatureAuthentication2018",
      "publicKey": [
        "did:sov:builder:VbPQNHsvoLZdaNU7fTBeFx#key-1"
      ]
    }
  ],
  "publicKey": [
    {
      "id": "did:sov:builder:VbPQNHsvoLZdaNU7fTBeFx#key-1",
      "publicKeyBase58": "GasrpJqMo3W9KApN4Dh62cGhYG24xhwNnCgqEKEiTKe6"
    }
  ]
}
```

https://www.w3.org/2019/did/v1 needs to be updated to  https://www.w3.org/ns/did/v1... this is currently not valid JSON-LD... but it is valid JSON.

What do we want to have happen here for when I request this with `application/did+json` ? should the intermediate code delete the `@context` to conform to did core, hiding this failure from JSON developers but throw an error when they ask for `application/did+ld+json` ?

***When a developer asks for JSON, they should get JSON that has not been altered in any way from what the did method code produces*** .... that means if the did producer added `@context` erroneously or intentionally it SHOULD not be deleted by intermediate producers / consumers, `@context` is either a known property in the registry, which should be preserved, or its an unknown property which should be preserved.

Accepting these changes would make the JSON representation more logically consistent and useful for developers like myself.

msporny:
This PR is currently blocked due to lack of consensus. The specification is also contradictory at this point in time wrt. the DID Spec Registries document. It's also not internally consistent wrt. JSON vs. JSON-LD (and even that's arguable because you could consider `@context` to NOT be an unknown property and we're calling it out and throwing an error on purpose -- I believe this is @jricher's position). 

In any case, we can't merge this yet as we haven't reached consensus yet. 

At this point, we should ask implementers what they plan to do with `@context` when they see it in a `document-stream` that has been marked with a did+json media type?

index.html
jricher: JSON and JSONLD aren't the same, and lazy producers shouldn't treat them as the same. That's the nature of the prohibition.

The transformation between formats should always go through the abstract data model. "context" is not a property in the same way that other properties are -- so much as it is semantic decoration (or a processing directive) on the model to make LD processing function properly, and it would be injected by a JSON LD producer when serializing from the model, just like it would do if it were serializing a model invented out of whole cloth. If we have interoperability at the data model level, this should work fine in all of the situations you describe above -- or at least, work in a consistent and predictable manner.

We still need to decide what it means to "ignore" a property. If it's "parse and preserve" or "drop" then it should be the same across all parsers and serializers, regardless of format. If JSON-LD separated its processing directives like "context" out of the way, then this wouldn't be a discussion we'd have to have at all. But since they're folded into the JSON document, then we need to prohibit the production of a non-LD document to have a "context" field lest a JSON-LD producer take in that field and erroneously pass it through instead of generating an appropriate context for the abstract data model that it is presenting.

OR13: > But since they're folded into the JSON document, then we need to prohibit the production of a non-LD document to have a "context" field lest a JSON-LD producer take in that field and erroneously pass it through instead of generating an appropriate context for the abstract data model that it is presenting.

Marking `@context` prohibited in JSON will just result in massive spec conformance tests failure, and it's logically inconsistent wrt the rest of JSON processing.

JSON Production should be defined as a process that yields a JSON (string) with an `id` property, all other properties are not required / should be ignored.

JSON Consumption should be defined as a process of parsing a JSON string.... and then checking for the existence of the `id` property... all properties by `id` should be ignored.

1. We don't need to prohibit JSON producers from producing JSON-LD.  (spec currently prohibits this)

2. We don't need to prohibit JSON consumers from consuming JSON-LD. (spec currently explicitly allows this)

3. We don't need to prohibit JSON-LD producers from producing JSON.  (spec currently prohibits this)

4. We don't need to prohibit JSON-LD consumers from consuming JSON. (spec currently prohibits this)

We need the JSON representation to be internally consistent. 

I think we can all agree that JSON is produced and consumed with JSON.stringify and JSON.parse (or their equivalents in other languages).... yet the normative statements in the spec today are not consistent with this.

We don't appear to have consensus regarding JSON-LD production / consumption, should properties be dropped?

Since the answer to this question depends on the presence of `@context` member, and steps such as `compact` and `expand` that happen ***after*** JSON.parse.... I suggest we address the JSON representation first. 

I propose that no properties are dropped from the JSON string returned when `application/did+json` is requested... The DID Method author should be able to produce JSON, and know that nobody is going to mutilate it, (causing it to hash to a different value)... ANY mutation to the result of `resolve` / `resolveStream` should be considered a very bad idea from a security perspective...

If JSON-LD folks want to do that, on the consumption side... thats their call.... lets not break JSON because of it.

This PR supports my proposal, and if accepted would make the following statements true:

1. JSON producers MAY produce JSON-LD.  (spec currently prohibits this)

2. JSON consumers MAY consuming JSON-LD. 

3. JSON-LD producers MAY produce JSON.  (spec currently prohibits this)

4. JSON-LD consumers MAY consume JSON. (spec currently prohibits this)

jricher: JSON-LD producers MUST produce JSON-LD (not non-LD JSON). JSON producers MUST produce JSON (not JSON-LD). The two should not mix. If you're going to transform them from one to the other then you need to process things post-parse as an abstract data model. If data model interoperability is really the goal then this shouldn't be a problem.

OR13: JSON-LD Producers, produce stuff that can be `JSON.stringified`... spec text won't change that.

JSON Producers, produce stuff that can be `JSON.stringified`.... spec text should be say more.... security issues arise if spec text says more.

JSON-LD Consumers, consume stuff that can be `JSON.parsed`.... spec text agrees, unclear what happens next.

JSON Consumers, consume stuff that can be `JSON.parsed`... spec text says they MUST consume JSON-LD.

@jricher are you proposing we call the JSON representation "Non-LD JSON" ? I would be in favor of that, especially given the normative prohibitions currently in the spec... :)

jricher: If changing the label is what makes this work for you, then that's fine with me. It's very clear to me that the proponents of the "plain JSON" or "pure JSON" are not interested in JSON-with-other-stuff-in-it, like JSON-LD. 

Also I think the spec needs to differentiate between a property of the DID document and a processing directive of the representation format. "context" is the latter, and so it having different rules from a property of the document is not only ok, it's expected. Reserving it so that other formats don't step on it and start treating it like a document property is the right move. It's important to think outside of the individual ecosystems when looking at this level of interoperability. You can't assume that everything starts out in JSON-LD and the rest of the world just has to deal with it. It's easy to imagine the happy path, where json.parse does the "right thing" well enough, but that's the same kind of thinking that got people used to calling "eval()" on JSON documents back in the day -- the happy path works but the other cases go haywire fast. I am arguing for a level of production/consumption that's at one step past the parser, since it's specifically about creating an object model with specific semantics -- the DID document.

Think of this degenerate case: A JSON producer creates a DID document and includes the `@context` field as a property in it. It's not meaningful to the JSON producer at all, but maybe they copied it from an example without knowing what they were doing, or they have their own internal meaning for the field and it's got some distinctively non-LD value in there, like `27`. The JSON producer drops that into the JSON stream. Someone else parses it properly, as JSON, and they get this extra `@context` property in there, since they "ignore" it and pass it through to the document model. But now someone's asking for this DID document in JSON-LD and so this party is now a producer and has to create a JSON-LD document from this abstract model. But... the abstract model already has a `@context` property. What does this second producer do with it? Do they make an array with the DID context and the `27`? Do they replace this weird value with their own? 

I argue that this is an error case that we can avoid both at the time the JSON document is produced by the first party and when it is consumed by the second party. In the first case, this is no longer a compliant document (this is in the text that's being tagged for removal here). In the second case, we throw an error (which I stated above should change in the JSON consumer section). This step happens after JSON.parse is called on the string, as the document is being transformed into the DID document abstract model using the rest of the rules in the section. 

Remember: JSON consumption isn't as simple as a parse call anyway. You are required to go through all of the properties that you know about and make sure they're the right data type and, in some cases, internal format. Throwing out an input with a prohibited property isn't any more difficult than what's already required.

OR13: > Also I think the spec needs to differentiate between a property of the DID document and a processing directive of the representation format.

strong +1

> You can't assume that everything starts out in JSON-LD and the rest of the world just has to deal with it.

I am not... I am assuming that things start as JSON :)


> But now someone's asking for this DID document in JSON-LD and so this party is now a producer and has to create a JSON-LD document from this abstract model. But... the abstract model already has a `@context` property. What does this second producer do with it? Do they make an array with the DID context and the `27`? Do they replace this weird value with their own?

Nobody should be "replacing" or "deleting" things that a DID Method Produces. That is a serious security mistake.

I can see the argument for throwing an error when a JSON producer returns data with an `@context`... but not if that same document can contain arbitrary properties like `homeAddress` or `ssn`.... this has nothing to do with JSON-LD... it has to do with data sanitization, immutability and security of JSON... the producer is responsible for doing this.... at the point at which they produce something, that artifact should now be treated as immutable... if anyone is going to fix a problem with it, it should be the original producer....

JSON Producers MAY include `@context` is the same as saying...

JSON Producers MAY include arbitrary unknown properties....

^ I have heard that this is a desired feature of JSON proponents... that I can add whatever I want to my did document, and it will not be dropped.... if thats the case, I should be able to add JSON Schema, JSON-LD, ... any other JSON... to my did document... once that JSON is returned by a Producer... nobody should be dropping those properties if they keep using vanilla JSON.... thats is what I am trying to achieve.

You are arguing that even though a JSON Producer added a value to a did document (that they knew)... other JSON producers know better and will remove it.... I think thats a security issue I can't live with as a JSON developer.

If you are arguing that JSON Producers don't know the properties they are adding to a did document, I can't help (same is true for JSON-LD btw).... I am assuming that DID Method authors / controllers know the properties they are adding to did documents... and that they are adding them for a reason.

We have to assume the following:

1. Producers take security seriously and only add JSON properties that they intend to use.
2. Producers are responsible for the original shape of the data, it should be possible to use integrity checking on what they produce, and know if some party down stream has altered it (regardless of the reason).

Skilled developers should be capable of producing JSON that is both JSON Schema and JSON LD and that hashes to the same value and that is a valid JSON DID Document.

Producers should be the ones who decide what JSON is returned, and nobody should be mutating did documents without clearly communicating that the document has been "sanitized" or "transformed".

JSON Producers should be able to do this:

Producer 1 -> Consumer 1 -> Producer 2 -> Consumer 2 -> Producer 3 -> Consumer 3

The JSON at the beginning should be the same as the JSON at the end or the JSON representation is not useful / extensible in the ways a JSON developer would expect (unknown properties are preserved, no schema technology is required AND no schema technology is prohibited).

Currently the did spec normatively prohibits this.

peacekeeper:
> At this point, we should ask implementers what they plan to do with `@context` when they see it in a `document-stream` that has been marked with a did+json media type?

@dlongley or others, is there any guidance on this question in the broader JSON-LD world? It doesn't feel like this question should be answered specifically for DID documents.

index.html
jricher: The argument of "JSON-LD is just JSON" doesn't fly. It very explicitly is not equivalent.

> JSON Producers MAY include @context is the same as saying...
> JSON Producers MAY include arbitrary unknown properties...

I completely disagree that these statements are equivalent, and that's at the root of the problem. This is also why a processing directive like context needs to be treated specially and not like a generic property. Yes it's the responsibility of the producer to produce a valid document, which is why we should continue to forbid them producing potentially bad documents. And it's the responsibility of the consumer to know how to handle bad documents, and this would be defined as one way things can be "bad". 

If a document gets translated through a bunch of steps, I would expect it to be semantically equivalent but not the same document at the end. So if Producer 2 spits out JSON-LD but producers 1 and 3 spit out JSON, what are we supposed to do with the context field that Producer 2 would have to add? 

The real chain is more like this, and the intermediate steps are very important:

DID-Doc1 (abstract) -> Producer 1 -> JSON -> Consumer 1 -> DID Doc2 (abstract) -> Producer 2 -> JSON-LD -> Consumer 2 -> DID Doc 3 (abstract) -> Producer3 -> JSON -> Consumer 3 -> DID-Doc4 (Abstract)

If we allow producer 1 to drop in "context" then this falls apart. But if we do these then it works even when you swap out :

1. Prohibit known processing directives (like context) where they don't belong -- we can make representations declare them upon registration instead of just listing "context" specially
2. Pass through unknown properties in the general data model

OR13: We are certainly making progress here.

> 2. Pass through unknown properties in the general data model

I agree with this.

> 1. Prohibit known processing directives (like context) where they don't belong 

presence of a JSON value in a JSON object is not a processing directive.... `content-type : application/did+json` is.

I think this is the heart of the issue.

https://w3c.github.io/did-core/#core-representations

> Consumers MUST NOT determine the representation of a document through its content alone.

I don't look in an HTTP body to see how I should process it.... and if its says its JSON... I don't delete a property from it as the first step after parsing it...

The JSON representation cannot prohibit JSON membership that might be useful for extensions / used by subsequent processing... that would break existing software that might be used on JSON DID Documents, and it breaks the extensibility model of pure JSON.... 

https://w3c.github.io/did-core/#extensibility

>Representations MAY define other extensibility mechanisms including methods for decentralized extensions. Such extension mechanisms MUST support lossless conversion into any other conformant representation.

for example:

https://json-schema.org/draft/2019-09/json-schema-core.html

Should JSON DID Documents be incapable of containing JSON Schema?... No.

Should a Producer use the `content-type` `application/did+schema+json` to indicate that the did document is both JSON and JSON Schema (processing directive). Yes.

If someone wants to make a stricter version of a did document that is ALWAYS valid WRT a JSON Schema, ALWAYS valid JSON-LD, and has some terrible mime type like `application/did+ld+schema+json`... 

They should be able to do so... Requiring the base level JSON DID Document representation to be incompatible with higher level JSON representations like `application/did+vnd.mycompany.myapp.v2+json`...  or `application/did+schema+json`...   or `application/did+ld+json`... is not necessary, counter intuitive and harmfully limits the value of vanilla JSON as representation.

JSON should be the least restrictive.... if people want a more restrictive implementation, they should create a representation that enforces those restrictions... and call it something other than `application/did+json`... like `application/did+vnd.mycompany.myapp.v2+json`... 

If a company wants to produce JSON which never contains an `@context` or a `$schema`... let them create another `content-type`... and require them to define exactly what that representation is... so that interoperability with it can be better understood.

jricher: I think here we have the disconnect:

My stance is that the JSON version should not be the most loose or most inclusive of all possible formats, but ironically my motivation for that is exactly the motivations you state as being in favor of JSON being the least restrictive. You should only be looking at the document as JSON, and so it shouldn't contain JSON-LD. You should be able to understand the entire thing purely as JSON and not need any JSON-LD processing to make sense of it. 

Additionally, `@context` and `$schema` are absolutely processing directives because they indicate to software what to do with the JSON object, and they define how to process it. The plain JSON format should be free of these things, since someone including one of these in the output document has every reason to assume it will be obeyed by the consumer. 

What's being proposed in this PR is great for JSON-LD, but detrimental for interoperability and any other formats coming down the road.

peacekeeper: After reading this thread several times, I have to say @jricher 's comments resonate with me more.. Yes JSON.parse will succeed even if there is a `@context`, but it doesn't feel right to me to say that a plain JSON consumer can/must/should just ignore it. This has been discussed in-depth in the 105 (!) comments in this PR last year: https://github.com/w3c/did-core/pull/142

Perhaps we should change the JSON consumption rules to something like this:

_If the consumer encounters a `@context` object member, it MUST either switch to JSON-LD consumption rules, or - if it does not support the JSON-LD representation - return an error._

jricher: I don't think we want the consumer to switch based on content. I would argue that the rule should be:

> If the consumer encounters a reserved object member, such as `@context`, it MUST return an error.

OR13:
Seems like one path forward would be do the following:

### Option 1
1. Create a representation called "JSON without Processing Directives".
2. Document the list of processing directives to include `@context` , `$schema`, `~thread`, `@type`
2 Let vanilla JSON defined as JSON that does not have properties removed or added as part of the production / consumption process.

### Option 2
1. Create a representation called "JSON with Processing Directives".
2. Let JSON with Processing Directives be defined as JSON that does not have properties removed or added as part of the production / consumption process.

:) I think most developers will recognize "JSON with Processing Directives" as the JSON they know and love : )

index.html
jricher: I've proposed a changed text at #394

OR13:
@jricher @selfissued @peacekeeper @msporny I have reverted the deletion of the normative statement, lets see if Justin can get the language correct, and leave this PR as guidance for JSON-LD consumers / producers.

OR13:
Since the controversial change has been removed from this PR and move to https://github.com/w3c/did-core/pull/396

I am closing this PR, and will open a new one with just changes for JSON-LD.

Issue #405, opened 2020-09-20
Is this a misspelling? 32 vs. 33 bit for publicKeyBase58


iherman:
(This may reflect my ignorance in crypto...)

The [Table on key types](https://www.w3.org/TR/did-core/#key-types-and-formats) includes a reference to the `publicKeyBase58` key. The text always says:

> be encoded as the raw 32-byte public key value encoded in Base58 Bitcoin format [BASE58] using the publicKeyBase58 property

_except_ for `secp256k1-koblitz` where it refers to 33-byte instead of 32. Is that intentional?

Also, while I am at it: the example below the table includes an example referring to the `JwsVerificationKey2020` class. This class is not defined in the table. Either define it as an official core class, or remove it from the example.

OR13:
`JwsVerificationKey2020` is a VM type... it is in the did spec registries as `JsonWebKey2020` I will correct the type.

IMO, almost all of the key representation stuff in did core should be removed, did method authors can represent keys however they want.... we created the did spec registries to show them how to have representations that are interoperable.

OR13:
33 is not a typo here is an example:

```
 {
          "id": "#zQ3shptjE6JwdkeKN4fcpnYQY3m9Cet3NiHdAfpvSUZBFoKBj",
          "type": "EcdsaSecp256k1VerificationKey2019",
          "controller": "did:key:zQ3shptjE6JwdkeKN4fcpnYQY3m9Cet3NiHdAfpvSUZBFoKBj",
          "publicKeyBase58": "24waDFAUAS16UpZwQQTXVEAmm17rQRjadjuAeBDW8aqL1"
        }
```

iherman:
Ok, so 33 bits is o.k. Thanks for checking

iherman:
> IMO, almost all of the key representation stuff in did core should be removed, did method authors can represent keys however they want.... we created the did spec registries to show them how to have representations that are interoperable.

That is an altogether different question and is one of the questions I asked in https://github.com/w3c/did-core/issues/403 on §5.3.1. To avoid confusion, and based on https://github.com/w3c/did-core/issues/405#issuecomment-695802241 I propose to close this issue.

brentzundel:
No comments since marked pending close, closing.

zamicol:
[...]

PR #406, opened 2020-09-20
Add JsonWebKey2020  (JWK) key representation to did core.

OR13:
addresses https://github.com/w3c/did-core/issues/405

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/406.html" title="Last updated on Sep 20, 2020, 3:50 PM UTC (602e2f0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/406/fbff723...602e2f0.html" title="Last updated on Sep 20, 2020, 3:50 PM UTC (602e2f0)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1683,6 +1683,14 @@ a JWK [[RFC7517]] or be encoded as the raw 32-byte public key value in Base58
 Bitcoin format [[BASE58]] using the <code>publicKeyBase58</code> property.
               </td>
             </tr>
+            <tr>
+              <td>
+JWK<br/>(<code>JsonWebKey2020</code>)
+              </td>
+              <td>
+Key types listed in <a href="https://www.iana.org/assignments/jose/jose.xhtml">JOSE</a>, represented using [[RFC7517]] using the <code>publicKeyJwk</code> property.
+              </td>
+            </tr>
           </tbody>
         </table>
 
@@ -1697,7 +1705,7 @@ Example:
   <span class="comment">...</span>
   "verificationMethod": [{
     "id": "did:example:123#_Qq0UL2Fq651Q0Fjd6TvnYE-faHiOpRlPVQcY_-tA4A",
-    "type": "JwsVerificationKey2020",
+    "type": "JsonWebKey2020",
     "controller": "did:example:123",
     "publicKeyJwk": {
       "crv": "Ed25519",


kdenhartog (approved):

msporny (approved):

index.html
```diff
@@ -1683,6 +1683,14 @@ <h3>Key types and formats</h3>
 Bitcoin format [[BASE58]] using the <code>publicKeyBase58</code> property.
               </td>
             </tr>
+            <tr>
+              <td>
+JWK<br/>(<code>JsonWebKey2020</code>)
```
msporny, line 1688: To be clear, I still object to key types where you cannot tell if the key was meant to have public key material or private key material. We have special rules in DID Core that help you understand that, but take this key type of this environment and it turns into a foot gun where processors can't determine whether it was meant to hold private key material or public key material.

OR13: It's not that JOSE developers don't understand your argument. It's that they find being able to work with existing software more valuable than absolute conformance to your aesthetic preferences.

Web Crypto happily understands the difference between JWKs.

Web Crypto has no idea what `Ed25519VerificationKey2018` and `publicKeyBase58` are.

Blocking support for legacy (in your mind, perhaps misguided) key representations is not the same as designing / innovating newer safer representations.

msporny: > Web Crypto has no idea what Ed25519VerificationKey2018 and publicKeyBase58 are.

Web Crypto has no idea what `JsonWebKey2020` is either... :)

You're inventing a new type that's broken in a way that has long been identified. You will understand the issue when people use the JsonWebKey2020 foot gun outside of DID Documents to accidentally publish private key material. 

> Web Crypto happily understands the difference between JWKs.

Can you explain how Web Crypto knows the intent behind a key expression? How does it know if the developer intended to express a public key vs. a private key?

OR13: https://mdn.github.io/dom-examples/web-crypto/export-key/index.html

``
{
 "crv": "P-384",
 "d": "JbPKurXh-rPefsLGZJ3c4wQNQ9nKKtNTvwQFfQBpnW1-oq4OTzzAKGXG0rAKbf7o",
 "ext": true,
 "key_ops": [
  "sign"
 ],
 "kty": "EC",
 "x": "rGpcD0j9TxQ90J7KnIehMq9G-rhmMjkbk1g3CI_wy_jkxCfHPtyU5PQ_15Uaz5IH",
 "y": "z3QT9Rzf4kFS2TZhxkGskThAN_wKhdaYwg2TfxNDhrbHLAnAEFBPMOZj3Qz-oMYl"
}
```

Web Crypto is substantially more documented than linked data crypto. 

https://developer.mozilla.org/en-US/docs/Web/API/Web_Crypto_API

In general, no software knows what a developer meant to do when they accidentally publish a `publicKeyJwk` with `d` or an `Ed25519VerificationKey2018` with `privateKeyBase58`.... there is no solution for human error, but clear documentation can help mitigate it.

msporny: > In general, no software knows what a developer meant to do when they accidentally publish a publicKeyJwk with d 

Why not? You have all the information you need, don't you? It's meant to be a public key -- it says it right there. Why are you not throwing an error when you see `d` in a public key!? :)

Note that the important information came from *outside* the JOSE structure... actually looking at the structure tells you nothing... you just assume it was always meant to be a private key... but it's both a public and a private key -- which is the whole point (and one of the problems with JOSE).

Also note that if the external structure said `privateKeyJwk`, you still might not know if that is what was meant... which is why `JsonWebKey2020` needs to be more specific.

OR13: > Why not? You have all the information you need, don't you? It's meant to be a public key -- it says it right there. Why are you not throwing an error when you see `d` in a public key!? :)

because https://github.com/w3c/did-spec-registries/blob/master/contexts/did-v1.jsonld#L63

I am relying on JSON development conventions... I'm not against defining all of it again in did core :)

But, it seems a terrible waste of time to copy all the terms and definitions from:

- https://tools.ietf.org/html/rfc7517
- https://www.iana.org/assignments/jose/jose.xhtml#web-key-parameters

I was thinking that it would easier to keep key representation details in the Linked Data Signature Suites, which define the verification method type, and keep did core to general support for widely used cryptography. 

I assume we don't want to / cannot link directly to the w3c ccg work items / linked data signature suites, but I do clearly explain this to readers: https://w3c-ccg.github.io/lds-jws2020/#json-web-key-2020

I'm hearing a contradiction in the key representations topics on my open PRs, perhaps we need another special topic call on key representations.

I'm hearing "define this more in did core" but also "we don't want to define linked data verification methods in did core"... given that, Im not sure how I can help progress this much more than I am now with the PRs I have open.

msporny:
Editorial, multiple reviews, changes requested, some made, no objections, merging.

index.html
msporny: > I was thinking that it would be easier to keep key representation details in the Linked Data Signature Suites, which define the verification method type, and keep did core to general support for widely used cryptography.

Agreed.

> I assume we don't want to / cannot link directly to the w3c ccg work items / linked data signature suites, but I do clearly explain this to readers

We do and can, see the NOTE at the very bottom of this section:

https://www.w3.org/TR/vc-data-model/#proofs-signatures

That section is also an example of what I hope we end up with in DID Core -- we talk about the properties and what they can contain, we provide examples, but don't go into details of each cryptographic suite.

PR #407, opened 2020-09-20
Cleanup key representations

OR13:
addresses https://github.com/w3c/did-core/issues/405

- [x] Remove undocumented and unused Schnorr hand waving
- [x] Remove publicKeyPem from RSA (JWK only)
- [x] Fix Json Web Key name
- [x] clarify how `publicKeyJwk` is used instead of `publicKeyBase58`

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/407.html" title="Last updated on Oct 9, 2020, 5:19 PM UTC (5a68b39)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/407/fbff723...5a68b39.html" title="Last updated on Oct 9, 2020, 5:19 PM UTC (5a68b39)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1658,7 +1658,7 @@ encoded as the raw 32-byte public key value in Base58 Bitcoin format
 secp256k1
               </td>
               <td>
-Secp256k1 Koblitz public key values MUST either be encoded as a JWK [[RFC7517]] 
+Secp256k1 public key values MUST either be encoded as a JWK [[RFC7517]] 
 using the <code>publicKeyJwk</code> or be
 encoded as the raw 33-byte public key value in Base58
 Bitcoin format [[BASE58]] using the <code>publicKeyBase58</code> property.
--- a/index.html
+++ b/index.html
@@ -1619,7 +1619,7 @@ expressed and encoded.
         <table class="simple" id="public-key-support">
           <caption>
 This table defines the support for public key expression in a DID document.
-For each public key type, a maximum of two encoding formats are supported.
+For each public key type, a maximum of two encoding formats is supported.
           </caption>
           <thead>
             <tr>
@@ -1638,9 +1638,8 @@ Support
 RSA<br/>(<code>RsaVerificationKey2018</code>)
               </td>
               <td>
-RSA public key values MUST either be encoded as a JWK [[RFC7517]] or be
-encoded in Privacy Enhanced Mail (PEM) format using the
-<code>publicKeyPem</code> property.
+RSA public key values MUST be encoded as a JWK [[RFC7517]] using the
+<code>publicKeyJwk</code> property.
               </td>
             </tr>
             <tr>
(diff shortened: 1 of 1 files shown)


brentzundel (approved):

index.html
```diff
@@ -1638,48 +1638,40 @@ <h3>Key types and formats</h3>
...
-[[RFC7517]] or be encoded as the raw 33-byte public key value in Base58
+Secp256k1 public key values MUST either be encoded as a JWK [[RFC7517]] 
+using the <code>publicKeyJwk</code> or be
+encoded as the raw 33-byte public key value in Base58
 Bitcoin format [[BASE58]] using the <code>publicKeyBase58</code> property.
               </td>
             </tr>
             <tr>
               <td>
-secp256r1<br/>(<code>SchnorrSecp256k1VerificationKey2019</code>)
```
msporny, line 1668: This PR removes secp256r1 support entirely, which is problematic as that is a widely deployed key type for HSMs -- we are going to have to provide this key format before CR. Rather than removing it, we should make sure it's defined in the security context.

OR13: `SchnorrSecp256k1VerificationKey2019` <- this is not defined anywhere.

https://crypto.stackexchange.com/questions/33052/is-there-any-difference-between-nist-and-secp-curves-in-terms-of-their-algorithm

`JsonWebKey2020` <- supports P-256.

msporny: Ahhh, right. So, we should clearly state that P-* is supported via JsonWebKey2020. Not mentioning it might lead people to believe the wrong thing, as I did.

OR13: I'm hesitant to add that clarity to this PR, but will happily be assigned an issue to follow up on it.

We do have this example https://w3c.github.io/did-core/#example-31-did-document-with-many-different-verification-methods

which shows a number of key types which are not mentioned anywhere in did core... 

I'd rather not try and define all the key types a suite supports again in did core... thats the job of the suite, not did core. 

I'm hoping we can work towards such language in this section, and instead just note that `verificationMethod.type` is responsible for defining the associated cryptography.

msporny: > I'm hesitant to add that clarity to this PR, but will happily be assigned an issue to follow up on it.

Given that this PR seems like it's removing P-* support, and given some of the direction given by Chairs on "things that have been in the spec for a while can't be removed without consensus by the group", I'm reluctant to approve this for removal. I'd rather we clearly express how support for secp256r1 is supported by DID Core in this PR.

We can ask the Chairs for clarity on the DID WG Editors call today.

OR13: @msporny based on your feedback on the other PR, i believe this would be a solution as well: https://github.com/w3c/did-core/issues/423

kdenhartog: I'd prefer to explicitly state them rather than take the approach of #423 because it gives detail and examples directly in the did-core spec rather than leaning on non-normative external documents for people to know what they can support. While it is legitimate to support any additional LD-proof key type, my expectation is that proper understanding and usage will fall into the "Tribal knowledge, advanced user feature" for the people who understand what LD-Proofs are and the rest will pick between the options made available in the spec.

m just now noticing this error `secp256r1 !== SchnorrSecp256k1VerificationKey2019`

msporny: @kdenhartog I'm reading your comment as: "Don't merge this, I want the secp256r1 section updated"? Am I reading your comment correctly?

msporny: @OR13 one way to deal with this is state that the group is still discussing examples of secp256r1 in an issue marker. That would allow us to merge this.

OR13: @msporny I added a comment here: https://github.com/w3c/did-core/issues/423#issuecomment-706305138 and added a note to the spec linking to that issue.

OR13: also, fyi... there is already an issue related to Schnorr: 

> The Working Group is still debating whether secp256k1 Schnorr public key values
> will be elaborated upon in this specification and if so, how they will be
> expressed and encoded.

but it does not make sense, since a "secp256k1 schnorr public key" is a "secp256k1" public key, which is supported in JWK or base58 encoding... hence the issue is IMO, not needed... unless you don't agree that schnorr signatures can be verified unless the verification method has a specific string value, that is not `EcdsaSecp256k1VerificationKey2019` or `JsonWebKey2020`... you can see why the former would be confusing, but the latter is not.

msporny (approved):

msporny:

drumenbassik (approved):

drumenbassik (approved):

drumenbassik (approved):

PR #408, opened 2020-09-20
Add JSON-LD Producer / Consumer details

OR13:
- [x] JSON-LD Producers MAY use `string` or `array of string or object` values for `@context`

- [x] JSON-LD Producers SHOULD NOT produce DID Documents with terms not defined in `@context`.
- [x] JSON-LD Consumers SHOULD drop properties from DID Documents not defined in `@context`.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/408.html" title="Last updated on Oct 12, 2020, 1:56 PM UTC (de525cd)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/408/1547e0a...de525cd.html" title="Last updated on Oct 12, 2020, 1:56 PM UTC (de525cd)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -15,6 +15,9 @@
     src='https://www.w3.org/Tools/respec/respec-w3c-common'>
   </script><!--script src='./respec-w3c-common.js' class='remove'></script-->
 
+  <script src="https://unpkg.com/reqlist/lib/reqlist.js" class="remove"></script>
+  <link href="https://unpkg.com/reqlist/lib/reqlist.css" class="removeOnSave" rel="stylesheet" type="text/css" />
+
   <script class='remove' src="./common.js">
   </script>
   <script class="remove" type="text/javascript">
@@ -55,6 +58,15 @@
             branch: "master"
     },
     includePermalinks: false,
+
+    // Uncomment these to use the respec extension that generates a list of
+    //   normative statements:
+    // preProcess: [
+    //   prepare_reqlist
+    // ],
+    // postProcess: [
+    //   add_reqlist_button
+    // ],
 
 
     // editors, add as many as you like
--- a/index.html
+++ b/index.html
@@ -2526,7 +2526,7 @@ The value of the <code>@context</code> property MUST be exactly one of the follo
   </li>
   <li>An array of one or more elements,
     where the value of the first element is
-    <code>https://www.w3.org/ns/did/v1</code>.
+    a string with the value of <code>https://www.w3.org/ns/did/v1</code>.
 
     <pre class="example">
       {
--- a/index.html
+++ b/index.html
@@ -2379,6 +2379,11 @@ the root object. Properties MAY define additional data sub structures subject
 to the value representation rules in the list above.
         </p>
 
+        <p>
+The member name <code>@context</code> MUST NOT be used as this property is
+reserved for JSON-LD producers.
+        </p>
+
 
       </section>
 
--- a/index.html
+++ b/index.html
@@ -2525,7 +2525,7 @@ The value of the <code>@context</code> property MUST be exactly one of the follo
     </pre>
   </li>
   <li>An array of one or more elements,
-    where the value of the first <a>URI</a> is
+    where the value of the first element is
     <code>https://www.w3.org/ns/did/v1</code>.
 
     <pre class="example">
--- a/index.html
+++ b/index.html
@@ -2626,7 +2626,7 @@ that decision.
 
         <p>
           Consumers SHOULD drop all properties from a DID Documents that 
-          are not defined by the <code>@context</code>. 
+          are not defined via the <code>@context</code>. 
         </p>
  
       </section>
--- a/index.html
+++ b/index.html
@@ -2572,7 +2572,7 @@ registries, then the DID Document will not be interoperable across
 representations.
 
 <p>
-  Producers SHOULD not produce DID Documents that 
+  Producers SHOULD NOT produce DID Documents that 
   contain properties not defined by the <code>@context</code>. 
   Such properties MAY be dropped by Consumers.
 </p>
--- a/index.html
+++ b/index.html
@@ -2557,6 +2557,7 @@ The value of <code>@context</code> MUST be exactly one of these values.
         ...
       }
     </pre>
+    <p class="issue"  data-number="432">The working group is still discussing restrictions on <code>@base</code>, beyond what JSON-LD allows.</p>
   </li>
 </ul>
 
--- a/index.html
+++ b/index.html
@@ -2549,10 +2549,7 @@ The value of <code>@context</code> MUST be exactly one of these values.
       {
         "@context": [
           "https://www.w3.org/ns/did/v1",
-          "https://example.com/blockchain-identity/v1",
-          {
-            "@base": "did:example:123"
-          }
+          "https://example.com/blockchain-identity/v1"
         ],
         ...
       }
--- a/index.html
+++ b/index.html
@@ -2510,43 +2510,41 @@ Production
         </h3>
 
         <p>
-The <a>DID document</a> is serialized following the rules in the JSON processor,
-with one addition: <a>DID documents</a> MUST include the <code>@context</code>
-property.
+The <a>DID document</a> MUST be serialized as a JSON document,
+with one additional requirement: <br/> 
+The <a>DID document</a> MUST include the <code>@context</code> property.
         </p>
 
         <dl>
           <dt>@context</dt>
           <dd>
-The value of the <code>@context</code> property MUST be exactly one of the following:
+
+<p>
+The <a href="https://www.w3.org/TR/json-ld11/#the-context">JSON-LD specification </a> 
+defines values that are valid for this property.
+</p>
+
+<p>
+The value of <code>@context</code> MUST be exactly one of these values.
+</p>
 
 <ul>
-  <li>The string <code>https://www.w3.org/ns/did/v1</code>. 
+  <li>
+    The <a href="https://infra.spec.whatwg.org/#strings">INFRA string</a>
+    <code>https://www.w3.org/ns/did/v1</code>.
     <pre class="example">
       {
         "@context": "https://www.w3.org/ns/did/v1",
-        "id": "did:example:123"
-      }      
+        ...
+      }
     </pre>
   </li>
-  <li>An array of one or more elements,
-    where the value types of the members of the array
-    MUST be either a string or an object and the first element is
-    the string <code>https://www.w3.org/ns/did/v1</code>.
-
-    <pre class="example">
-      {
-        "@context": [
-          "https://www.w3.org/ns/did/v1",
-          "https://example.com/blockchain-identity/v1"
-        ],
-        "id": "did:example:123"
-      }      
-    </pre>
-  </li>
-  <li>An array where the first value is
-    <code>https://www.w3.org/ns/did/v1</code>, and the final value is an object.
-
+  <li>
+    An <a href="https://infra.spec.whatwg.org/#lists">INFRA list</a>, 
+    with first item the <a href="https://infra.spec.whatwg.org/#strings">INFRA string</a>
+    <code>https://www.w3.org/ns/did/v1</code>, and subsequent items of type 
+    <a href="https://infra.spec.whatwg.org/#strings">INFRA string</a> or 
+    <a href="https://infra.spec.whatwg.org/#maps">INFRA map</a>.
     <pre class="example">
       {
         "@context": [
@@ -2556,11 +2554,12 @@ The value of the <code>@context</code> property MUST be exactly one of the follo
             "@base": "did:example:123"
           }
         ],
-        "id": "did:example:123"
-      }      
+        ...
+      }
     </pre>
   </li>
 </ul>
+
 
 All members of the
 <code>@context</code> property SHOULD exist in the DID specification
@@ -2568,12 +2567,12 @@ registries in order to achieve interoperability across different
 representations. 
 
 If a member does not exist in the DID specification
-registries, then the DID Document will not be interoperable across
+registries, then the DID Document might not be interoperable across
 representations.
 
 <p>
   Producers SHOULD NOT produce DID Documents that 
-  contain properties not defined by the <code>@context</code>. 
+  contain properties not defined via the <code>@context</code>. 
   Properties that are not defined via the <code>@context</code> MAY be dropped by Consumers.
 </p>
 
--- a/index.html
+++ b/index.html
@@ -2379,10 +2379,6 @@ the root object. Properties MAY define additional data sub structures subject
 to the value representation rules in the list above.
         </p>
 
-        <p>
-The member name <code>@context</code> MUST NOT be used as this property is
-reserved for JSON-LD producers.
-        </p>
 
       </section>
 
@@ -2517,25 +2513,66 @@ property.
         <dl>
           <dt>@context</dt>
           <dd>
-The value of the <code>@context</code> property MUST be one or more <a>URIs</a>,
-where the value of the first <a>URI</a> is
-<code>https://www.w3.org/ns/did/v1</code>. All members of the
+The value of the <code>@context</code> property MUST be exactly one of the following:
+
+<ul>
+  <li>The string <code>https://www.w3.org/ns/did/v1</code>. 
+    <pre class="example">
+      {
+        "@context": "https://www.w3.org/ns/did/v1",
+        "id": "did:example:123"
+      }      
+    </pre>
+  </li>
+  <li>An array of one or more <a>URIs</a>,
+    where the value of the first <a>URI</a> is
+    <code>https://www.w3.org/ns/did/v1</code>.
+
+    <pre class="example">
+      {
+        "@context": [
+          "https://www.w3.org/ns/did/v1",
+          "https://example.com/blockchain-identity/v1"
+        ],
+        "id": "did:example:123"
+      }      
+    </pre>
+  </li>
+  <li>An array where the first value is
+    <code>https://www.w3.org/ns/did/v1</code>, and the final value is an object.
+
+    <pre class="example">
+      {
+        "@context": [
+          "https://www.w3.org/ns/did/v1",
+          "https://example.com/blockchain-identity/v1",
+          {
+            "@base": "did:example:123"
+          }
+        ],
+        "id": "did:example:123"
+      }      
+    </pre>
+  </li>
+</ul>
+
+All members of the
 <code>@context</code> property SHOULD exist in the DID specification
 registries in order to achieve interoperability across different
-representations. If a member does not exist in the DID specification
+representations. 
+
+If a member does not exist in the DID specification
 registries, then the DID Document will not be interoperable across
 representations.
+
+<p>
+  Producers SHOULD not produce DID Documents that 
+  contain properties not defined by the <code>@context</code>. 
+  Such properties MAY be dropped by Consumers.
+</p>
+
           </dd>
         </dl>
-
-        <p class="issue" data-number="202">
-This specification defines globally interoperable documents, and the requirement
-that the context value be in the verifiable data registry means that different JSON-LD
-processors can consume the document without having to dereference anything in
-the context field. However, a pair of producers and consumers can have local
-extension agreements. This needs to be clarified in a section on extensibility
-and linked here when available.
-        </p>
 
       </section>
 
@@ -2554,9 +2591,10 @@ interpreted using JSON-LD processing under the rules of the defined
         <dl>
           <dt>@context</dt>
           <dd>
-The value of the <code>@context</code> property MUST be one or more <a>URIs</a>,
-where the value of the first <a>URI</a> is
-<code>https://www.w3.org/ns/did/v1</code>. If more than one <a>URI</a> is
+The value of the <code>@context</code> property MUST conform 
+to the <a href="#production-0">JSON-LD Production Rules</a>. 
+
+If more than one <a>URI</a> is
 provided, the <a>URIs</a> MUST be interpreted as an ordered set. It is
 RECOMMENDED that dereferencing each <a>URI</a> results in a document containing
 machine-readable information about the context. To enable interoperability
@@ -2580,6 +2618,11 @@ unknown data members on consumption, and this section needs to be updated with
 that decision.
         </p>
 
+        <p>
+          Consumers SHOULD drop all properties from a DID Documents that 
+          are not defined by the <code>@context</code>. 
+        </p>
+ 
       </section>
 
     </section>
--- a/index.html
+++ b/index.html
@@ -2574,7 +2574,7 @@ representations.
 <p>
   Producers SHOULD NOT produce DID Documents that 
   contain properties not defined by the <code>@context</code>. 
-  Such properties MAY be dropped by Consumers.
+  Properties that are not defined via the <code>@context</code> MAY be dropped by Consumers.
 </p>
 
           </dd>
--- a/index.html
+++ b/index.html
@@ -2524,7 +2524,7 @@ The value of the <code>@context</code> property MUST be exactly one of the follo
       }      
     </pre>
   </li>
-  <li>An array of one or more <a>URIs</a>,
+  <li>An array of one or more elements,
     where the value of the first <a>URI</a> is
     <code>https://www.w3.org/ns/did/v1</code>.
 


index.html
```diff
@@ -2517,26 +2518,68 @@ <h3>
...
+      {
+        "@context": "https://www.w3.org/ns/did/v1",
+        "id": "did:example:123"
+      }      
+    </pre>
+  </li>
+  <li>An array of one or more elements,
+    where the value types of the members of the array
+    MUST be either a string or an object and the first element is
+    the string <code>https://www.w3.org/ns/did/v1</code>.
```
msporny, line 2535: Try to use Infra types for "array", "string", and "object".

index.html
```diff
@@ -2517,26 +2518,68 @@ <h3>
...
+    <code>https://www.w3.org/ns/did/v1</code>, and the final value is an object.
+
+    <pre class="example">
+      {
+        "@context": [
+          "https://www.w3.org/ns/did/v1",
+          "https://example.com/blockchain-identity/v1",
+          {
+            "@base": "did:example:123"
+          }
```
msporny, line 2557: Feeling stronger about this now... let's not invite trouble... can we just say something to the effect of "anything that is valid for JSON-LD is valid here" and then give a simple example that defines a one-type use property (even that I'm feeling uncomfortable stating).

index.html
```diff
@@ -2517,26 +2518,68 @@ <h3>
...
 registries in order to achieve interoperability across different
-representations. If a member does not exist in the DID specification
+representations. 
+
+If a member does not exist in the DID specification
 registries, then the DID Document will not be interoperable across
 representations.
+
+<p>
+  Producers SHOULD not produce DID Documents that
```
msporny, line 2575:
```suggestion
  Producers SHOULD NOT produce DID Documents that 
```

Might want to link to "producer" and "consumer" definitions in conformance section?

index.html
```diff
@@ -2517,26 +2518,68 @@ <h3>
...
+representations. 
+
+If a member does not exist in the DID specification
 registries, then the DID Document will not be interoperable across
 representations.
+
+<p>
+  Producers SHOULD not produce DID Documents that 
+  contain properties not defined by the <code>@context</code>. 
+  Such properties MAY be dropped by Consumers.
```
msporny, line 2577:
```suggestion
  Properties that are not defined via the <code>@context</code> MAY be dropped by Consumers.
```

... just to be extra clear. Ideally conformance statements are single sentences that stand on their own. Imagine extracting the sentence out of the specification and reading it w/o any of the surrounding context in order to help craft conformance statements that are crystal clear.

index.html
```diff
@@ -2580,6 +2624,11 @@ <h3>
 that decision.
         </p>
 
+        <p>
+          Consumers SHOULD drop all properties from a DID Documents that 
+          are not defined by the <code>@context</code>.
```
msporny, line 2629:
```suggestion
          are not defined via the <code>@context</code>. 
```

`by` indicates that `@context` defines it directly... `via` indicates that it is defined "by way of" the `@context` ... that is, linking elsewhere is just as good as defining it inline. Might be too pedantic, but we should be consistent throughout the document.

msporny (changes requested):
Minor nits... the `@base` example is problematic since you can't do that in JSON-only and we have a Base URL algorithm in the spec. We may even want to go as far as outlawing the use of `@base` in DID Documents... `id` establishes base... if we allow JSON-LD to override that here, I think we might harm interop and potentially open ourselves up to security attacks for DID Methods that use JSON-LD and do relative references within the document.

Other than the `@base` thing and some minor editorial nits, LGTM.

index.html
OR13: @msporny "anything that is valid for JSON-LD is valid here" this is exactly what I am trying to say! :)

Can you propose concrete language here? We don't need to even define `@context` if we say that we support `100%` of JSON-LD, and just link to that defintion.... I would prefer to just say that, and provide an example that uses JSON-LD relative refs, in the JSON-LD representation section.

OR13:
> Minor nits... the `@base` example is problematic since you can't do that in JSON-only and we have a Base URL algorithm in the spec. We may even want to go as far as outlawing the use of `@base` in DID Documents... `id` establishes base... if we allow JSON-LD to override that here, I think we might harm interop and potentially open ourselves up to security attacks for DID Methods that use JSON-LD and do relative references within the document.
> 
> Other than the `@base` thing and some minor editorial nits, LGTM.

@msporny Sidetree, Microsoft, Transmute and others use `@base`.... it was chosen specifically because of the desire to have JSON-LD that looks like compact JSON (`@context is ignored`) and is treated the same way by JSON-LD and JSON...

https://identity.foundation/sidetree/spec/v0.1.0/ (you have to search for `@base`.... spec-up is terrible).

> If the DID Document is being rendered in the JSON-LD representation variant, the implementer SHOULD add an @base entry to the document’s @context, and set the @base value to the id of the resolved DID. This ensures relative path values in the output DID Document are correctly projected into id-related strings by JSON-LD parsers.

I would object to prohibiting `@base` in DID Core, it would cripple JSON / JSON-LD interop in `did:ion`, `did:elem` and our representation format for `did:key`, which proves that `@base` is safe to use... so we are not going to stop using it....  However I would not object to a sentence in JSON-LD Producer section that stated:

"If `@base` is present in `@context` it SHOULD/MUST match the `id` of the DID Document".

I'm not sure if that closes the door to fancy use cases / features I am not aware of in JSON-LD... Sidetree doesn't say `@base` MUST match the `id`.... 

This is another one of those "maybe the did method author should make this call scenarios"... my goal is to allow them to make decisions as long as they don't have interoperability and support for off the shelf tooling.

index.html
OR13: @msporny I attempted to clarify this in https://github.com/w3c/did-core/pull/408/commits/f9f1ee80144cc434533b530c23c66c83c33bcda9

index.html
OR13: done in https://github.com/w3c/did-core/pull/408/commits/f9f1ee80144cc434533b530c23c66c83c33bcda9

msporny:
> If the DID Document is being rendered in the JSON-LD representation variant, the implementer SHOULD add an `@base` entry to the document’s `@context`, and set the `@base` value to the id of the resolved DID. This ensures relative path values in the output DID Document are correctly projected into id-related strings by JSON-LD parsers.

The right way to do that is to set the BASE for the JSON-LD processor, not to hack it with the addition of `@base`. We may want to instead put a normative requirement in JSON-LD consumers to set the @base to the value of `id` before processing. That way, it would just work in JSON-LD w/o having to hack `@base` in there.

What we have here is probably not good enough to do that right now:

https://w3c.github.io/did-core/#relative-did-urls

OR13:
I am strongly against asking anyone reading the did core spec to "configure JSON-LD Processors"...

I would even be in favor of language which specifically said, "NO JSON-LD Processing happens as a result of JSON-LD Consumption".

msporny:
> I am strongly against asking anyone reading the did core spec to "configure JSON-LD Processors"...

I went back and read the JSON-LD 1.1 spec... and I think we're miscommunicating... the base IRI is set automatically via the Document Loader interface:

https://www.w3.org/TR/json-ld11-api/#the-jsonldprocessor-interface

> The documentUrl in the returned RemoteDocument is used as base IRI and the contextUrl is used instead.

I believe that ^^^ happens automatically, so doing it again via `@base` is unnecessary when doing JSON-LD processing.

> I would even be in favor of language which specifically said, "NO JSON-LD Processing happens as a result of JSON-LD Consumption".

Hmm, this is where the definition of "consumption" becomes tricky... I assume you mean: "Don't do JSON-LD processing when you consume a JSON-LD document into the abstract data model"... leaving developers to choose when they invoke the JSON-LD processing algorithms, which I'd be happy with putting into the spec.

I want to merge this, but we need more opinions on `@base` before we do so.

/cc @dlongley @gkellogg

OR13:
@gkellogg @msporny 

The reason I propose the use of `@base` is that I have tests using DIDs and VCs that I know work, and I have compared the associated N-QUADS, to ensure that what I have is equivalent.

consider: 

```
{
    "@context": [
        "https://w3id.org/did/v0.11"
    ],
    "id": "did:key:z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd",
    "publicKey": [
        {
            "id": "did:key:z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd#z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd",
            "controller": "did:key:z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd",
            "publicKeyBase58": "5yKdnU7ToTjAoRNDzfuzVTfWBH38qyhE1b9xh4v8JaWF"
        }
    ],
    "authentication": [
        "did:key:z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd#z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd"
    ],
    "assertionMethod": [
        "did:key:z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd#z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd"
    ],
    "capabilityDelegation": [
        "did:key:z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd#z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd"
    ],
    "capabilityInvocation": [
        "did:key:z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd#z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd"
    ],
    "keyAgreement": [
        {
            "id": "did:key:z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd#zC68u6qt8EEiaCuuZKWLQ3hAkrQ6K1BMKuWSerjLCvyCKc",
            "type": "X25519KeyAgreementKey2019",
            "controller": "did:key:z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd",
            "publicKeyBase58": "6eBPUhK2ryHmoras6qq5Y15Z9pW3ceiQcZMptFQXrxDQ"
        }
    ]
}
```

yields: 

```
<did:key:z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd#z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <https://w3id.org/security#Ed25519VerificationKey2018> .
<did:key:z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd#z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd> <https://w3id.org/security#controller> <did:key:z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd> .
<did:key:z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd#z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd> <https://w3id.org/security#publicKeyBase58> "5yKdnU7ToTjAoRNDzfuzVTfWBH38qyhE1b9xh4v8JaWF" .
<did:key:z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd#zC68u6qt8EEiaCuuZKWLQ3hAkrQ6K1BMKuWSerjLCvyCKc> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <https://json-ld.org/playground/X25519KeyAgreementKey2019> .
<did:key:z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd#zC68u6qt8EEiaCuuZKWLQ3hAkrQ6K1BMKuWSerjLCvyCKc> <https://w3id.org/security#controller> <did:key:z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd> .
<did:key:z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd#zC68u6qt8EEiaCuuZKWLQ3hAkrQ6K1BMKuWSerjLCvyCKc> <https://w3id.org/security#publicKeyBase58> "6eBPUhK2ryHmoras6qq5Y15Z9pW3ceiQcZMptFQXrxDQ" .
<did:key:z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd> <https://w3id.org/security#assertionMethod> <did:key:z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd#z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd> .
<did:key:z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd> <https://w3id.org/security#authenticationMethod> <did:key:z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd#z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd> .
<did:key:z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd> <https://w3id.org/security#capabilityDelegationMethod> <did:key:z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd#z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd> .
<did:key:z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd> <https://w3id.org/security#capabilityInvocationMethod> <did:key:z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd#z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd> .
<did:key:z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd> <https://w3id.org/security#keyAgreementMethod> <did:key:z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd#zC68u6qt8EEiaCuuZKWLQ3hAkrQ6K1BMKuWSerjLCvyCKc> .
<did:key:z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd> <https://w3id.org/security#publicKey> <did:key:z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd#z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd> .
```

now consider: 

```
{
    "@context": [
        "https://w3id.org/did/v0.11",
      {"@base": "did:key:z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd"}
    ],
    "id": "did:key:z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd",
    "publicKey": [
        {
            "id": "#z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd",
            "controller": "did:key:z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd",
            "publicKeyBase58": "5yKdnU7ToTjAoRNDzfuzVTfWBH38qyhE1b9xh4v8JaWF"
        }
    ],
    "authentication": [
        "#z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd"
    ],
    "assertionMethod": [
        "#z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd"
    ],
    "capabilityDelegation": [
        "#z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd"
    ],
    "capabilityInvocation": [
        "#z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd"
    ],
    "keyAgreement": [
        {
            "id": "#zC68u6qt8EEiaCuuZKWLQ3hAkrQ6K1BMKuWSerjLCvyCKc",
            "type": "X25519KeyAgreementKey2019",
            "controller": "did:key:z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd",
            "publicKeyBase58": "6eBPUhK2ryHmoras6qq5Y15Z9pW3ceiQcZMptFQXrxDQ"
        }
    ]
}
```

Yields:

```
<did:key:z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd#z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <https://w3id.org/security#Ed25519VerificationKey2018> .
<did:key:z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd#z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd> <https://w3id.org/security#controller> <did:key:z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd> .
<did:key:z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd#z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd> <https://w3id.org/security#publicKeyBase58> "5yKdnU7ToTjAoRNDzfuzVTfWBH38qyhE1b9xh4v8JaWF" .
<did:key:z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd#zC68u6qt8EEiaCuuZKWLQ3hAkrQ6K1BMKuWSerjLCvyCKc> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <did:X25519KeyAgreementKey2019> .
<did:key:z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd#zC68u6qt8EEiaCuuZKWLQ3hAkrQ6K1BMKuWSerjLCvyCKc> <https://w3id.org/security#controller> <did:key:z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd> .
<did:key:z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd#zC68u6qt8EEiaCuuZKWLQ3hAkrQ6K1BMKuWSerjLCvyCKc> <https://w3id.org/security#publicKeyBase58> "6eBPUhK2ryHmoras6qq5Y15Z9pW3ceiQcZMptFQXrxDQ" .
<did:key:z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd> <https://w3id.org/security#assertionMethod> <did:key:z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd#z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd> .
<did:key:z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd> <https://w3id.org/security#authenticationMethod> <did:key:z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd#z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd> .
<did:key:z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd> <https://w3id.org/security#capabilityDelegationMethod> <did:key:z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd#z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd> .
<did:key:z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd> <https://w3id.org/security#capabilityInvocationMethod> <did:key:z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd#z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd> .
<did:key:z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd> <https://w3id.org/security#keyAgreementMethod> <did:key:z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd#zC68u6qt8EEiaCuuZKWLQ3hAkrQ6K1BMKuWSerjLCvyCKc> .
<did:key:z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd> <https://w3id.org/security#publicKey> <did:key:z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd#z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd> .
```

msporny:
> The reason I propose the use of @base is that I have tests using DIDs and VCs that I know work, and I have compared the associated N-QUADS, to ensure that what I have is equivalent.

Not saying what you have doesn't work. I'm sure it does. I'm saying that it should work w/o specifying `@base`, and if it doesn't, your Document Loader is probably misconfigured by not setting `base IRI` correctly.

OR13:
> > The reason I propose the use of @base is that I have tests using DIDs and VCs that I know work, and I have compared the associated N-QUADS, to ensure that what I have is equivalent.
> 
> Not saying what you have doesn't work. I'm sure it does. I'm saying that it should work w/o specifying `@base`, and if it doesn't, your Document Loader is probably misconfigured by not setting `base IRI` correctly.

JSON-LD Playground is broken then, this is what I get without `@base`.

```
<did:key:z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd> <https://w3id.org/security#assertionMethod> <https://json-ld.org/playground/#z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd> .
<did:key:z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd> <https://w3id.org/security#authenticationMethod> <https://json-ld.org/playground/#z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd> .
<did:key:z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd> <https://w3id.org/security#capabilityDelegationMethod> <https://json-ld.org/playground/#z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd> .
<did:key:z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd> <https://w3id.org/security#capabilityInvocationMethod> <https://json-ld.org/playground/#z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd> .
<did:key:z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd> <https://w3id.org/security#keyAgreementMethod> <https://json-ld.org/playground/#zC68u6qt8EEiaCuuZKWLQ3hAkrQ6K1BMKuWSerjLCvyCKc> .
<did:key:z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd> <https://w3id.org/security#publicKey> <https://json-ld.org/playground/#z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd> .
<https://json-ld.org/playground/#z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <https://w3id.org/security#Ed25519VerificationKey2018> .
<https://json-ld.org/playground/#z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd> <https://w3id.org/security#controller> <did:key:z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd> .
<https://json-ld.org/playground/#z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd> <https://w3id.org/security#publicKeyBase58> "5yKdnU7ToTjAoRNDzfuzVTfWBH38qyhE1b9xh4v8JaWF" .
<https://json-ld.org/playground/#zC68u6qt8EEiaCuuZKWLQ3hAkrQ6K1BMKuWSerjLCvyCKc> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <https://json-ld.org/playground/X25519KeyAgreementKey2019> .
<https://json-ld.org/playground/#zC68u6qt8EEiaCuuZKWLQ3hAkrQ6K1BMKuWSerjLCvyCKc> <https://w3id.org/security#controller> <did:key:z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd> .
<https://json-ld.org/playground/#zC68u6qt8EEiaCuuZKWLQ3hAkrQ6K1BMKuWSerjLCvyCKc> <https://w3id.org/security#publicKeyBase58> "6eBPUhK2ryHmoras6qq5Y15Z9pW3ceiQcZMptFQXrxDQ" .
```

> In its absence, it falls back to the encapsulating document (e.g., the base of an HTML document containing JSON-LD in a script element) or 5.1.3 URI used to retrieve the entity.

^ this is problematic to rely on... consider the cases where the URI has been lost, but you have the document.... and of course the HTML stuff will not work.

msporny:
> JSON-LD Playground is broken then, this is what I get without @base.

It's not broken... the JSON-LD Playground sets the `Base IRI` to the URL of the playground if it's not already set. You could suggest another base URL, but that community has been around this discussion multiple times and picked the URL it did because that's where consensus was at the time. We could add an option that let's you change the `Base IRI` in the Playground, but again, that would require people to write the code.

> ^ this is problematic to rely on... consider the cases where the URI has been lost, but you have the document.... and of course the HTML stuff will not work.

If you lose the URI, your system is broken wrt. relative URLs... just like when you lose the base URI for HTML pages your relative references in your HTML page becomes broken. This is why I was arguing against relative URIs in the DID Document, but I lost that argument long ago. I'm glad you're starting to recognize the problems created by relative IRIs in DID Documents. :)

That said, the problem isn't insurmountable... if you are missing the `Base IRI`, we tell you exactly what to set it to in DID Core:

https://w3c.github.io/did-core/#relative-did-urls

The issue isn't that JSON-LD is broken, or the DID Core spec is broken. The issue is that using `@base` is solving a problem that is already solved and so the suggestion that we should provide guidance on using `@base` when the `Base IRI` is missing is... misguided. If you follow DID Core when processing DID Documents, which you should be doing, then `Base IRI` can never be not set to the `id` in the DID Document.

OR13:
> If you lose the URI, your system is broken wrt. relative URLs...

this is not true.

https://www.w3schools.com/tags/att_base_href.asp

^ this is functionally equivalent to what I am proposing.

There is a difference between:

- The URI that was requested, and that I must keep hold of to have relative refs that work.
- The base URI in the document, which I will have as long as I have the document.

> In its absence, it falls back to the encapsulating document (e.g., the base of an HTML document containing JSON-LD in a script element) or 5.1.3 URI used to retrieve the entity.

"the base of an HTML document containing JSON-LD in a script element"

^ this won't work.

"URI used to retrieve the entity" 

^ this information is not a reliable source for the base especially when all you have is the document.

Configuration of JSON-LD Processors should be not be a requirement for relative refs to work. 

The Document by itself should support base URIs.

If the default behavior of JSON-LD processors was to set `@base` to `didDocument.id`... this would not be necessary, but because thats not the default behavior, this is necessary... consider reading the following JSON from disk and applying the frame api directly to it.... what happens?

```
{
    "@context": [
        "https://w3id.org/did/v0.11"
    ],
    "id": "did:key:z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd",
    "publicKey": [
        {
            "id": "#z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd",
            "controller": "did:key:z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd",
            "publicKeyBase58": "5yKdnU7ToTjAoRNDzfuzVTfWBH38qyhE1b9xh4v8JaWF"
        }
    ],
    "authentication": [
        "#z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd"
    ],
    "assertionMethod": [
        "#z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd"
    ],
    "capabilityDelegation": [
        "#z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd"
    ],
    "capabilityInvocation": [
        "#z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd"
    ],
    "keyAgreement": [
        {
            "id": "#zC68u6qt8EEiaCuuZKWLQ3hAkrQ6K1BMKuWSerjLCvyCKc",
            "type": "X25519KeyAgreementKey2019",
            "controller": "did:key:z6MkjRagNiMu91DduvCvgEsqLZDVzrJzFrwahc4tXLt9DoHd",
            "publicKeyBase58": "6eBPUhK2ryHmoras6qq5Y15Z9pW3ceiQcZMptFQXrxDQ"
        }
    ]
}
```

The tests I have run suggest it ALWAYS fails, without "doing something which as a moderately experience JSON-lD developer, I still don't know how to do"....

Much safer to just allow `@base` because JSON-LD allows it, leave it to people who understand advanced JSON-LD features to use or not use the approach I am suggesting.

OR13:
To be clear, I am not objecting to "there is a better way to do this that is more complicated than including `@base`, and requires you to configure the JSON-LD Processor manually before applying api's like `frame`, `compact` or `expand`."

I am objecting to "lets prohibit a solution that works with off the shelf tooling and requires no software configuration".

I suggest documenting, the alternative solution in the implementation guide, since it requires configuration of software, which is not specific to the data model itself.

dlongley:
> Much safer to just allow @base because JSON-LD allows it, leave it to people who understand advanced JSON-LD features to use or not use the approach I am suggesting.

You can already do this. You can download a DID Document and add `@base` to it -- or you can store DID Documents in a VDR using `@base`. But when a DID Document is returned from `resolve`, regular consumers shouldn't have to deal with it. So I don't think we should allow it there, because it's forcing additional processing on those consumers. I certainly don't want to have to process `@base` all the time in my software -- but you're welcome to take the output of `resolve` and add `@base` in your software all day long, it won't affect me and we don't have to say anything to enable you to do that.

msporny:
>  "there is a better way to do this that is more complicated than including @base, and requires you to configure the JSON-LD Processor manually before applying api's like frame, compact or expand."

If you think that is the proposal, then you're not understanding the proposal. :)

The proposal is: A `conforming consumer` that understands JSON-LD is responsible for setting the `Base IRI` on the JSON-LD processor. A `conforming consumer` that understands JSON-only is responsible for modifying all relative URLs in the DID Document such that the resulting absolute URLs are resolved relative to `id`.

What you are proposing only works for JSON-LD... JSON-only processors will fail.

Does that make sense?

OR13:
> The proposal is: A `conforming consumer` that understands JSON-LD is responsible for setting the `Base IRI` on the JSON-LD processor before doing any JSON-LD processing. A `conforming consumer` that understands JSON-only is responsible for modifying all relative URLs in the DID Document such that the resulting absolute URLs are resolved relative to `id`.
> 
> What you are proposing only works for JSON-LD... JSON-only processors will fail.
> 
> Does that make sense?

-1 to your proposal :)

I have JSON-only + JSON-LD interop tests for DID Key, which takes this approach.... it works and nobody needs to "mutate" the result of resolvers... I will continue to object to any "mutation" of the raw bytes returned by a did method, doing so creates security / usability issues, and its not necessary.

Here is my proposal:

A JSON-LD DID Document Producer returns JSON, with an `@context` with first value equal to `https://www.w3.org/ns/did/v1`, and subsequent values conforming to what is allowed by https://w3c.github.io/json-ld-syntax/#dfn-context.

`@base` need not be specifically mentioned in DID Core, but prohibiting it / altering the expectations around how `@context` is defined in the JSON-LD spec.... that I will continue to object to.

To be clear, I have test proving that what @msporny and @dlongley are proposing does not actually work.

Here they are: https://github.com/OR13/jsonld-did-production-with-base/blob/main/produce-without-base.test.js#L23

Note that this only applies to JSON-only interop with `fragment` only identifiers... something I expect you may not use regularly.

dlongley:
@OR13,

> To be clear, I have tests proving that what @msporny and @dlongley are proposing does not actually work.
> 
> Here they are: https://github.com/OR13/jsonld-did-production-with-base/blob/main/produce-without-base.test.js#L23

I don't think those tests do what @msporny is proposing, which is passing the `base` option to the JSON-LD processor. You can see here:

https://github.com/OR13/jsonld-did-production-with-base/blob/main/test-helpers.js#L18
```js
const expanded = await jsonld.expand(didDocument, { documentLoader });
```

That it is not passed. You need to do:

```js
const expanded = await jsonld.expand(didDocument, { documentLoader, base: didDocument.id });
```

Also, what I'm proposing is to add `@base` if you want after resolution or use `@base` in your VDR before resolution -- or anywhere else as you see fit, just don't return it as the result of `resolve`. The result of `resolve` shouldn't require people to process it because it keeps things simple. These approaches are captured in your other positive tests and they pass.

So, the test that fails is the one that doesn't pass in `base` to the processor -- which isn't what @msporny or I proposed.

dlongley:
@OR13,

> @base need not be specifically mentioned in DID Core, but prohibiting it / altering the expectations around how @context is defined in the JSON-LD spec.... that I will continue to object to.

I would also object to altering the expectations around how `@context` is defined. I'm not suggesting we do that -- I'm suggesting that, of the JSON-LD legal values for `@context` we choose a limited set for DID Documents that are returned from `resolve` (limit "producers"). This is just defining a particular "profile" for `@context` for DID Documents returned from `resolve`, it does not alter the definition of a JSON-LD context nor change behavior or how processors work, it just limits what people produce so as to make things simpler for consumers.

In fact, this approach enables one to treat the JSON-LD version just like the JSON version with only an additional check on expected strings in the `@context` array. If the `@context` field is permitted in the JSON version, then simple consumption software can treat both the same (this has been my preference from the beginning).

Again, these limitations would only apply to what is returned from `resolve` -- only to "producers" as it is narrowly defined (we still need better terminology here).

index.html
```diff
@@ -2509,34 +2510,75 @@ <h3>
         </h3>
 
         <p>
-The <a>DID document</a> is serialized following the rules in the JSON processor,
-with one addition: <a>DID documents</a> MUST include the <code>@context</code>
-property.
+The <a>DID document</a> MUST be serialized as a JSON document,
+with one additional requirement: <br/> 
+The <a>DID document</a> MUST include the <code>@context</code> property.
```
peacekeeper, line 2515: Since `@context` isn't a "property" of the DID subject, can we call it something else?

```suggestion
The JSON document MUST include the <code>@context</code> member.
```

index.html
```diff
@@ -2509,34 +2510,75 @@ <h3>
...
         <dl>
           <dt>@context</dt>
           <dd>
-The value of the <code>@context</code> property MUST be one or more <a>URIs</a>,
-where the value of the first <a>URI</a> is
-<code>https://www.w3.org/ns/did/v1</code>. All members of the
+
+<p>
+The <a href="https://www.w3.org/TR/json-ld11/#the-context">JSON-LD specification </a> 
+defines values that are valid for this property.
```
peacekeeper, line 2524:
```suggestion
defines values that are valid for this member.
```

index.html
OR13: -1 to using the term "member".

both `id ` and `@context` are "properties of a did document"... only `id` is required.

http://json-schema.org/understanding-json-schema/reference/object.html#:~:text=%2C%20%22object%22%5D-,Properties,used%20to%20validate%20that%20property.

OR13:
I am against suggesting developers do this:

```
const expanded = await jsonld.expand(didDocument, { documentLoader, base: didDocument.id });
```

or limiting the values of `@context` in a way that developers would have to do this, to have relative URIs in JSON-LD DID Documents.

this spec is about a data model, but this change you ask for developers to understand JSON-LD in a way that I don't believe is helpful.

I am not against a DID Method, making the choice to set `@base` in a processor, and saying so in their did method spec... thats fine.

I am against preventing a DID Method from choosing to handle this without configuring a processor.

I am also not sure of the language support for this proposal in Go, Rust, Python, .NET, or Java.... which is another reason, not to require this approach...

Even if someone wrote tests to prove it worked in all those languages, I have already shown a simply test, that highlights that what I have works without the more complicated configuration.

index.html
msporny: Agree with @OR13 -- this whole contemplation of different classes of properties in a DID Document is new and does not have consensus.

@peacekeeper -- are you objecting to this PR if your suggested change is not made?

index.html
msporny: Not my style, but it's not harmful, and it seems like people want the extra verbosity. I'm standing aside.

msporny (approved):

index.html
msporny: I still think `@base` examples in DID Core is a bad idea. I'm actively blocking this PR -- we need to get broader consensus from the group on this.

msporny (changes requested):
Blocking this PR until the `@base` example is removed. I'm suggesting this is an anti-pattern and adds nothing wrt. DID Core as it is. It's not that doing `@base` is illegal in JSON-LD... it's just that it doesn't have a mapping in JSON-only... and the JSON-only algorithm is to use `didDocument.id` as the base URI.

One thing that could get this PR is in adding an issue to the `@base` example noting that the group is discussing whether it's usage is appropriate in DID Core.

OR13:
@msporny opened an issue and linked to this PR.

csuwildcat:
If one wants to cleanly support relative pathing while using JSON-LD, I don't see how `@base` isn't the best option, given it automatically performs the desired action. This seems like an option that should certainly be noted somewhere in the spec (at the moment, it's already added to ION DID Docs when requested as JSON-LD).

index.html
peacekeeper: `id` is a [property of the DID subject](https://w3c.github.io/did-core/#dfn-id). `@context` is not.

OR13: `id` is also a `property` of a did document, in the sense that JSON objects have properties... perhaps the ADM does not have properties in your mind?

IMO the ADM also supports `@context` as a property, because otherwise it would not be possible to use the ADM to produce JSON-LD.

index.html
```diff
@@ -2509,34 +2510,76 @@ <h3>
...
+    <a href="https://infra.spec.whatwg.org/#strings">INFRA string</a> or 
+    <a href="https://infra.spec.whatwg.org/#maps">INFRA map</a>.
+    <pre class="example">
+      {
+        "@context": [
+          "https://www.w3.org/ns/did/v1",
+          "https://example.com/blockchain-identity/v1",
+          {
+            "@base": "did:example:123"
+          }
```
msporny, line 2555:
```suggestion
          "https://example.com/blockchain-identity/v1"
```

OR13: @msporny I am taking this change, acknowledging that we have not reach consensus on restrictions of `@context`, and I think its best we do so on issues instead of PRs. See https://github.com/w3c/did-core/issues/432

msporny: Agree that by taking this change we're not saying that anything around `@base` is resolved. The intent is to continue discussing how the group views `@base` and whether or not we should provide guidance in the spec on its usage.

msporny (approved):

msporny:

Issue #356, opened 2020-07-23
Restrict the number of key representation members


OR13:
`publicKeyJwk`, `publicKeyBase58`.... you should only be allowed to use 1 of these in a did core vm.

Someday it will be obvious which of these is the most common for all key types.

OR13:
cc @msporny as discussed on the call.

msporny:
Yes, agreed... the reasoning being that we don't want people to try to express the same key in two different formats and accidentally put in two different keys. By restricting to one used vs. the other, we eliminate a lot of error checking code.

OR13:
I believe a normative statement here, would address this issue:

https://w3c.github.io/did-core/#verification-methods

we should generalize this to account for non public key verification material.

OR13:
We need to add guidance that you cannot include multiple key representations in the same verification method. needs pr.

OR13:
Needs a PR saying "don't out private material" in did documents.

rhiaro:
Looks like this is done and can be closed?

OR13:
agreed, this should be closed.

msporny:
Issue submitter agrees that the issue has been addressed. Closing.

PR #415, opened 2020-09-23
Forbid multiple key formats in the same verification method

OR13:
addresses #356

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/415.html" title="Last updated on Sep 23, 2020, 2:08 PM UTC (900ea11)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/415/14b5eda...900ea11.html" title="Last updated on Sep 23, 2020, 2:08 PM UTC (900ea11)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1591,9 +1591,6 @@ A <a>verification method</a> MUST contain a single verification material propert
 (for example <code>publicKeyJwk</code> or <code>publicKeyBase58</code>, not both).
 </p>
 
-
-
-
         <p>
 This specification strives to limit the number of formats for expressing public
 key material in a <a>DID document</a> to the fewest possible, to increase the
--- a/index.html
+++ b/index.html
@@ -1586,6 +1586,14 @@ property of <em>result</em>, throw an error.
 A public key can be used as a <a>verification method</a>.
         </p>
 
+<p>
+A <a>verification method</a> MUST contain a single verification material property
+(for example <code>publicKeyJwk</code> or <code>publicKeyBase58</code>, not both).
+</p>
+
+
+
+
         <p>
 This specification strives to limit the number of formats for expressing public
 key material in a <a>DID document</a> to the fewest possible, to increase the
--- a/index.html
+++ b/index.html
@@ -1587,8 +1587,10 @@ A public key can be used as a <a>verification method</a>.
         </p>
 
 <p>
-A <a>verification method</a> MUST contain a single verification material property
-(for example <code>publicKeyJwk</code> or <code>publicKeyBase58</code>, not both).
+A <a>verification method</a> MUST NOT contain multiple verification material properties. 
+For example, expressing key material in a <a>verification method</a> using both 
+<code>publicKeyJwk</code> and <code>publicKeyBase58</code> at the same time 
+is prohibited.
 </p>
 
         <p>


msporny (approved):

index.html
```diff
@@ -1586,6 +1586,11 @@ <h3>Key types and formats</h3>
 A public key can be used as a <a>verification method</a>.
         </p>
 
+<p>
+A <a>verification method</a> MUST contain a single verification material property
+(for example <code>publicKeyJwk</code> or <code>publicKeyBase58</code>, not both).
```
msporny, line 1591:
```suggestion
A <a>verification method</a> MUST NOT contain multiple verification material properties. 
For example, expressing key material in a <a>verification method</a> using both 
<code>publicKeyJwk</code> and <code>publicKeyBase58</code> at the same time 
is prohibited.
```

Might want to specify this as a prohibition instead.

msporny (changes requested):
Perhaps state this in the negative or say "MUST only".

index.html
OR13: yes, i had it that way, but changed it... was not sure of the conventions for this.

msporny: I prefer what I suggested above. Let's wait to see if others have better suggestions than the two options put forward.

msporny: Oh wait, you already accepted it... bleh -- Github now doesn't show that the suggested change was accepted by you. :(

In any case, I'm happy with it as is... let's see if the new text confuses in a different way. :)

OR13: I had what you suggested before, I committed... I think it's clearer. I was just not sure what the standard practice for "allow-list" vs "deny-list" in specs is.... in security "allow-list" is generally better, but in this case, I agree its less readable.

msporny: Oh, I see what you're saying now.

From a spec text perspective, as a general rule, whatever is more clear and resistant to being misinterpreted is preferred. Consistency on how you say things throughout the spec is also a consideration, but less of one than the previous sentence.

In general, allow-lists are preferred over deny-lists... except when you really don't want the thing on the deny-list to happen because it will lead to catastrophic failures. That's more of an exception than a rule... and fundamentally, what goes in is whatever you get the group to come to consensus on... and unfortunately that's not always consistent throughout the spec. 

So, Editors are tasked with keeping all the language as consistent as possible throughout the spec while keeping at least all the things above in mind.

iherman:
I am all for a precise definition:-) The example of no JWK and base58 in the same verification method is clear. To avoid any kind of misunderstanding it is probably worth adding that, also, only one base58 (or JWK) property can exist on one verification method. Note that, at this point, the RSA VM uses a third term (`publicKeyPem`) which must be unique as well. Maybe worth emphasizing that, too.

(This is what I tried to express in the `JWKorBase58` shape in [SHACL](https://github.com/w3c/did-core/wiki/Draft-DID-Core-Vocabulary-Shape)).

OR13:
> Note that, at this point, the RSA VM uses a third term (`publicKeyPem`) which must be unique as well.

I have another open PR to remove that.

msporny (approved):

kdenhartog (approved):

peacekeeper (approved):

msporny:
Normative, multiple positive reviews, changes requested and made, merging.

Issue #361, opened 2020-07-28
Define serializations for Numbers, Dates, other properties


jricher:
Infra doesn't define serializations for things like numbers and dates, so we should do that for metadata properties within the spec. Infra gives us strings and we can define how the string is parsed. This can be a simple set of recommendations like "use a decimal integer for integers" and "use an ISO date in XXX format". 

These need to be recommendations that property definitions can refer to. Properties MUST define how they're parsed, though, so we need something to anchor on.

OR13:
Related to big numbers... https://tools.ietf.org/html/rfc8785

jricher:
JCS is probably not the best approach to handling this, but the points in the above are sound.

OR13:
Is there some better way of defining how the JSON representations we discuss in the spec handle dates and numbers?... clearly we are having trouble using Infra to do that.

I'd love to have a way to clearer about what we mean when we say JSON.... otherwise the did spec would seem to just be ignoring these issues with JSON.... and we use JSON a lot :) JSON-LD is JSON.

jricher:
This isn't about JSON, this is about strings within Infra. 

JSON serialization of values is its own other thing entirely.

msporny:
> the did spec would seem to just be ignoring these issues with JSON

Which version of JSON subset do you mean?

https://tools.ietf.org/html/rfc7159
https://tools.ietf.org/html/rfc7493 

:)

... and then there are issues wrt. converting to/from CBOR.

In any case, the thing Justin is asking for is hopefully a little simpler than solving all of the issues w/ converting from JSON-to-JSON across different architectures (53-bit vs 64-bit JSON environments, for example).

OR13:
Either way, DID Core should not use a term like "JSON" or "INFRA" without warning about these issues.

iherman:
This is also related to #379

jonnycrunch:
similar concern for ordered map that I brought up before.  https://infra.spec.whatwg.org/#maps defines ordered map, but doesn't specify HOW it is ordered to make it deterministic.  I, myself, like the option with json<-->dagCBOR that is in https://tools.ietf.org/html/rfc7049.

jonnycrunch:
copied over from issue #435,  

[...]

PR #282, opened 2020-05-12
Added CBOR section 

jonnycrunch:
added CBOR section.  Not totally complete but I could use some help validating the CDDL and tag 98 syntax.  Thanks!   @kdenhartog @selfissued @jricher

***
<a href="https://pr-preview.s3.amazonaws.com/jonnycrunch/did-core/pull/282.html" title="Last updated on Jun 9, 2020, 3:22 PM UTC (3b4a2b6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/282/48bc712...jonnycrunch:3b4a2b6.html" title="Last updated on Jun 9, 2020, 3:22 PM UTC (3b4a2b6)">Diff</a>

PR #420, opened 2020-09-29
updated cbor section

jonnycrunch:
updated CBOR to expand discussion regarding consumption and production.

***
<a href="https://pr-preview.s3.amazonaws.com/jonnycrunch/did-core/pull/420.html" title="Last updated on Dec 1, 2020, 3:58 PM UTC (8bec5aa)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/420/b86bed4...jonnycrunch:8bec5aa.html" title="Last updated on Dec 1, 2020, 3:58 PM UTC (8bec5aa)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -134,13 +134,6 @@
         company: "Danube Tech",
         companyURL: "https://danubetech.com/",
         w3cid: 46729
-      },
-      {
-        name: "Jonathan Holt, DO, MS",
-        url: "https://www.linkedin.com/in/jonathan-holt-do-ms",
-        company: "ConsenSys Health",
-        companyURL: "https://www.consensyshealth.com",
-        w3cid: 95341
       }
     ]
     };
@@ -2806,179 +2799,118 @@ Like Javascript Object Notation (JSON) [[RFC8259]], Concise Binary Object
 Representation (CBOR) [[RFC7049]] defines a set of formatting rules for the
 portable representation of structured data. CBOR is a more concise,
 machine-readable, language-independent data interchange format that is
-self-describing and has built-in semantics for interoperability.  With specific
-constraints, CBOR can support all JSON data types (including JSON-LD) for
+self-describing and has built-in semantics for interoperability.  CBOR prioritizes 
+the simplicity of the encoder and decoder over the size of the encoded data.
+With specific constraints, CBOR can support all JSON data types for
 translation between the DID document model (described in <a
-href="#data-model">Data Model</a> and <a>DID Documents</a>) and other core
-representations.
+href="#data-model">Data Model</a> and the <a href="#core-properties"> Core Properties section</a> of the DID Document</a>).
       </p>
 
-      <p>
-Concise Data Definition Language (CDDL) [[RFC8610]] is a notation used to
-express Concise Binary Object Representation (CBOR), and by extension JSON Data
-Structures. The following notation expresses the DID Document model in CBOR
-representation with specific constraints for deterministic mappings between
-other core representations.
-      </p>
-
-      <pre class="example nohighlight" title="DID Document data model for CBOR expressed in CDDL notation">
-DID-document = {
-  ? @context : uri
-  id : did
-  ? publicKey : [* publicKey ]
-  ? authentication :  [ *did // *publicKey // *tstr ]
-  ? service : [ + service  ]
-  ? controller : did / [ *did ]
-  ? created : time
-  ? updated : time
-  proof :  any
-}
-
-publicKey = {
-  id : did
-  type : text
-  controller : uri
-}
-
-did = tstr .pcre "^did\\:(?&lt;method-name>[a-z0-9]{2,})\\:(?&lt;method-specific-id>[A-Za-z0-9\\.\\-\\:\\_]+)"
-
-did-url =  tstr .pcre "^did\\:(?&lt;method-name>[a-z0-9]{2,})\\:(?&lt;method-specific-id>[A-Za-z0-9\\.\\-\\:\\_]+)\\;(?&lt;path>[A-Za-z0-9\\/)(?&lt;query>\\?[a-z0-9\\=\\&amp;])#(?&lt;fragment>.+)"
-
-service = {
-  id : did-url
-  type : text
-  serviceEndpoint : uri
-  ? description : text
-  * tstr => any
-}
-      </pre>
-
+<p>
+When producing and consuming DID documents that are encoded in CBOR (as indicated 
+by a content-type of application/did+cbor in the resolver metadata), the following rules MUST be followed.
+</p>
+      
       <section>
         <h3>Production</h3>
         <p>
-When producing DID Documents that are represented as CBOR, in addition to the
-suggestions in section 3.9 of the CBOR [[RFC7049]] specification for
-deterministic mappings, the following constraints of the DID Document model
-MUST be followed:
-        </p>
+A DID document encoded in CBOR must be represented as a CBOR Map (major type 5) conforming to [[RFC7049]]. 
+Similar to the JSON encoding of a DID document, all properties of the DID document MUST be represented by using the 
+property name as the name of the member of the CBOR map.  The values of properties of the data model described in <a href="#core-properties"> § 5. Core Properties</a>, including 
+all extensions, MUST be encoded in CBOR [[RFC7049]] by mapping INFRA property values to CBOR types as follows: </p>
 
-        <ul>
-          <li>
-Map keys MUST be strings.
-          </li>
-          <li>
-Integer encoding MUST be as short as possible.
-          </li>
-          <li>
-The expression of lengths in CBOR major types 2 through 5 MUST be as short as
-possible.
-          </li>
-          <li>
-All floating point values MUST be encoded as 64-bits, even for integral values.
-          </li>
-        </ul>
+<ul>
+<li>Integer values (A real number without a fractional part) are serialized as <a href="https://tools.ietf.org/html/rfc7049#section-2.1">CBOR integers (major types 0 and 1)</a>, choosing the shortest form.</li>
+<li>Double values (Number values with a fractional part) are serialized as <a href="https://tools.ietf.org/html/rfc7049#section-2.1">CBOR floating-point values (major type 7)</a>. All floating point values MUST be encoded as 64-bits (additional type value 27), even for integral values.</li>
+<li>Strings are serialized as <a href="https://tools.ietf.org/html/rfc7049#section-2.1">CBOR UTF-8 strings (major type 3)</a>. Note that, unlike JSON, CBOR does not require any escaping in strings, so all escape codes (if any) present in the alternate representations must be replaced beforehand by their corresponding character.</li>
+<li>Boolean values MUST be represented as a <a href="https://tools.ietf.org/html/rfc7049#section-2.3">CBOR Primitve type (major type 7) with value 21 for true and 20 for false). </a> </li>
+<li>List (A finite sequence of values) MUST be represented as a <a href="https://tools.ietf.org/html/rfc7049#section-2.1">CBOR Array type (major type 4) </a>.</li>
+<li>Unordered sets of values MUST be represented as a <a href="https://tools.ietf.org/html/rfc7049#section-2.1">CBOR Array type (major type 4).</a></li>
+<li>Ordered Map ( A Set of properties ) MUST be represented as a <a href="https://tools.ietf.org/html/rfc7049#section-2.1">CBOR map type (major type 5).</a></li>
+<li>Null (Empty values) MUST be represented as a <a href="https://tools.ietf.org/html/rfc7049#section-2.1">CBOR null literal Primitive type (major type 7) value 22.</a> </li>
+<li>Other Data Types not otherewise defined in the Abstract Data Model MUST be represented as a <a href="https://tools.ietf.org/html/rfc7049#section-2.1">CBOR String type. (major type 5)</a></li>
+<li>Indefinite-length items are not allowed and MUST be made <a href="https://tools.ietf.org/html/rfc7049#section-2.2.1">CBOR definite length.</a></li>
+<li>Datetime MUST be encoded as <a href="https://tools.ietf.org/html/rfc7049#section-2.4.1">CBOR UTF-8 (major type 3)</a>. This datetime value MUST be normalized to UTC 00:00, as indicated by the trailing "Z".</li>
+</ul>
 
-        <pre class="example nohighlight" title="An example DID Document represented as contrained CBOR and exported in diagnostic annotation mode for easy readibility">
-    a7                                                # map(7)
-    62                                                #   text(2)
-       6964                                           #     "id"
-    78 40                                             #   text(64)
-       6469643a6578616d706c653a31324433               #     "did:example:12D3"
-       4b6f6f574d4864727a6377706a626472               #     "KooWMHdrzcwpjbdr"
-       5a733547477145524176636771583362               #     "Zs5GGqERAvcgqX3b"
-       3564707550745061396f743639796577               #     "5dpuPtPa9ot69yew"
-    65                                                #   text(5)
-       70726f6f66                                     #     "proof"
-    a4                                                #   map(4)
-       64                                             #     text(4)
-          74797065                                    #       "type"
-       74                                             #     text(20)
-          656432353531395369676e617475726532303138    #       "ed25519Signature2018"
-       67                                             #     text(7)
-          63726561746564                              #       "created"
-       74                                             #     text(20)
-          323032302d30352d30315430333a30303a30325a    #       "2020-05-01T03:00:02Z"
-       67                                             #     text(7)
-          63726561746f72                              #       "creator"
-       78 8c                                          #     text(140)
-          6469643a6578616d706c653a31324433            #       "did:example:12D3"
-          4b6f6f574d4864727a6377706a626472            #       "KooWMHdrzcwpjbdr"
-          5a733547477145524176636771583362            #       "Zs5GGqERAvcgqX3b"
-          3564707550745061396f743639796577            #       "5dpuPtPa9ot69yew"
-          3b206578616d706c653a6b65793d6964            #       "; example:key=id"
-          3d626166797265696375627478357771            #       "=bafyreicubtx5wq"
-          6f336e6f73633463617a726b63746668            #       "o3nosc4cazrkctfh"
-          776436726577657a6770776f65347377            #       "wd6rewezgpwoe4sw"
-          69726c733465626468733269                    #       "irls4ebdhs2i"
-       6e                                             #     text(14)
-          7369676e617475726556616c7565                #       "signatureValue"
-       78 58                                          #     text(88)
-          6f3972364c78676f474e38466f616565            #       "o9r6LxgoGN8Foaee"
-          554136456444637631324776447a4645            #       "UA6EdDcv12GvDzFE"
-          6d43676a577a76707572325953517941            #       "mCgjWzvpur2YSQyA"
-          3857327230535357554b2b6e4835744d            #       "8W2r0SSWUK+nH5tM"
-          717a61464c756e3677775a31456f7433            #       "qzaFLun6wwZ1Eot3"
-          37616d4744673d3d                            #       "7amGDg=="
-    67                                                #   text(7)
-       63726561746564                                 #     "created"
-    74                                                #   text(20)
-       323031382d31322d30315430333a30303a30305a       #     "2018-12-01T03:00:00Z"
-    67                                                #   text(7)
-       75706461746564                                 #     "updated"
-    74                                                #   text(20)
-       323032302d30352d30315430333a30303a30305a       #     "2020-05-01T03:00:00Z"
-    68                                                #   text(8)
-       40636f6e74657874                               #     "@context"
-    78 1c                                             #   text(28)
-       68747470733a2f2f7777772e77332e6f               #     "https://www.w3.o"
-       72672f6e732f6469642f7631                       #     "rg/ns/did/v1"
-    69                                                #   text(9)
-       7075626c69634b6579                             #     "verificationMethod"
-    81                                                #   array(1)
-       a5                                             #     map(5)
-          62                                          #       text(2)
-             6964                                     #         "id"
-          78 85                                       #       text(133)
-             6261667972656963756274783577716f         #         "bafyreicubtx5wqo"
-             336e6f73633463617a726b6374666877         #         "3nosc4cazrkctfhw"
-             6436726577657a6770776f6534737769         #         "d6rewezgpwoe4swi"
-             726c7334656264687332693b6578616d         #         "rls4ebdhs2i;exam"
-             706c653a6b65793d6964626166797265         #         "ple:key=idbafyre"
-             6963756274783577716f336e6f736334         #         "icubtx5wqo3nosc4"
-             63617a726b6374666877643672657765         #         "cazrkctfhwd6rewe"
-             7a6770776f6534737769726c73346562         #         "zgpwoe4swirls4eb"
-             6468733269                               #         "dhs2i"
-          64                                          #       text(4)
-             74797065                                 #         "type"
-          6e                                          #       text(14)
-             45644473615075626c69634b6579             #         "EdDsaPublicKey"
-          65                                          #       text(5)
-             6375727665                               #         "curve"
-          67                                          #       text(7)
-             65643235353139                           #         "ed25519"
-          67                                          #       text(7)
-             65787069726573                           #         "expires"
-          74                                          #       text(20)
-             323031392d31322d30315430333a30303a30305a #         "2019-12-01T03:00:00Z"
-          6f                                          #       text(15)
-             7075626c69634b6579426173653634           #         "publicKeyBase64"
-          78 2c                                       #       text(44)
-             716d7a3774704c4e4b4b4b646c376344         #         "qmz7tpLNKKKdl7cD"
-             375062656a4469425670374f4e706d5a         #         "7PbejDiBVp7ONpmZ"
-             62666d633763454b396d673d                 #         "bfmc7cEK9mg="
-    6e                                                #   text(14)
-       61757468656e7469636174696f6e                   #     "authentication"
-    81                                                #   array(1)
-       78 83                                          #     text(131)
-          6469643a6578616d706c653a31324433            #       "did:example:12D3"
-          4b6f6f574d4864727a6377706a626472            #       "KooWMHdrzcwpjbdr"
-          5a733547477145524176636771583362            #       "Zs5GGqERAvcgqX3b"
-          3564707550745061396f743639796577            #       "5dpuPtPa9ot69yew"
-          3b6b65792d69643d6261667972656963            #       ";key-id=bafyreic"
-          756274783577716f336e6f7363346361            #       "ubtx5wqo3nosc4ca"
-          7a726b63746668776436726577657a67            #       "zrkctfhwd6rewezg"
-          70776f6534737769726c733465626468            #       "pwoe4swirls4ebdh"
-          733269                                      #       "s2i"
+
+<!-- <p>Additional considerations to enable lossless production of a CBOR representation and to enable a deterministic 
+  canonical CBOR representation of the DID Document between other core representations of the DID document is possible 
+  with additional constraints to the CBOR representation.  -->
+  <p>
+  To produce a deterministic canonical CBOR representation of a DID document and faciliate maximal lossless compatiblity with other 
+  core representations via the Abstract Data Model the following constraints of a CBOR representation of a DID Document model MUST be followed:
+ 
+<ul>
+    <li>Property names MUST be represented as text string (major type 3) and contain only UTF-8 strings. </li>
+    <li>Undefined Values of Required Properties as defined in <a href="#data-model">the Data Model</a> that are absent from the CBOR representation SHOULD be labeled with Primitive type (major type 7) with value 23 (Undefined value).  </li>
+    <li>Property names in each CBOR map MUST be unique. </li>
+    <li>Integer encoding MUST be as short as possible.</li>
+    <li>The expression of lengths in CBOR major types 2 through 5 MUST be as short as possible. </li>
+    <li>The keys in every map must be sorted lowest value to highest. Sorting is performed on the bytes of the representation of the keys. If two keys have different lengths, the shorter one sorts earlier.  If two keys have the same length, the one with the lower value in (byte-wise) lexical order sorts earlier. </li>
+  </ul>
+</p>
+
+
+<p class="issue">
+ How to represent Floating-point values that can exceed the range or the precision IEEE 754. See <a href=https://github.com/w3c/did-core/issues/361">issue #361. </a>
+</p>
+
+<p>
+  The property name `@context` MAY be present in the CBOR representation of a DID Document and if present SHOULD be ignored as this property is reserved for JSON-LD processing.
+</p>
+
+<p>
+  All properties of the DID document represented in CBOR MUST be included in the root map (major type 5). 
+  Properties MAY define additional data sub structures represented as nested CBOR maps (major type 5) 
+  and is subject to the value representation rules in the lists above and conformance to section <a href="#extensibility"> § 4.3 Extensibility</a>.
+</p>
+
+<p>
+</p>
+
+  <ul>
+  </ul>
+
+
+<pre class="example nohighlight" title="An example DID Document (the same as Example 2) represented as constrained CBOR presented as Hexadecimal">
+  A2626964781E6469643A6578616D706C653A313233343536373839616263
+  6465666768696E61757468656E7469636174696F6E81A462696478256469
+  643A6578616D706C653A313233343536373839616263646566676869236B
+  6579732D316474797065781A45643235353139566572696669636174696F
+  6E4B6579323031386A636F6E74726F6C6C6572781E6469643A6578616D70
+  6C653A3132333435363738396162636465666768696F7075626C69634B65
+  79426173653538782C483343324156764C4D7636676D4D4E616D33755641
+  6A5A70666B634A437744776E5A6E367A3377586D715056
+</pre>
+  
+<pre class="example nohighlight" title="The same example DID Document shown above but exported 
+in diagnostic annotation mode for easy readibility">  
+A2                                   # map(2)
+62                                   # text(2)
+   6964                              # "id"
+78 1E                                # text(30)
+   6469643A6578616D706C653A313233343536373839616263646566676869 # "did:example:123456789abcdefghi"
+6E                                   # text(14)
+   61757468656E7469636174696F6E      # "authentication"
+81                                   # array(1)
+   A4                                # map(4)
+      62                             # text(2)
+         6964                        # "id"
+      78 25                          # text(37)
+         6469643A6578616D706C653A313233343536373839616263646566676869236B6579732D31 # "did:example:123456789abcdefghi#keys-1"
+      64                             # text(4)
+         74797065                    # "type"
+      78 1A                          # text(26)
+         45643235353139566572696669636174696F6E4B657932303138 # "Ed25519VerificationKey2018"
+      6A                             # text(10)
+         636F6E74726F6C6C6572        # "controller"
+      78 1E                          # text(30)
+         6469643A6578616D706C653A313233343536373839616263646566676869 # "did:example:123456789abcdefghi"
+      6F                             # text(15)
+         7075626C69634B6579426173653538 # "publicKeyBase58"
+      78 2C                          # text(44)
+         483343324156764C4D7636676D4D4E616D337556416A5A70666B634A437744776E5A6E367A3377586D715056 # "H3C2AVvLMv6gmMNam3uVAjZpfkcJCwDwnZn6z3wXmqPV"
         </pre>
 
       </section>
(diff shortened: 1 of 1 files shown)


contexts/did-v0.11.jsonld
```diff
@@ -0,0 +1,58 @@
+{
```
rhiaro, line 1: The contexts live in https://github.com/w3c/did-spec-registries/tree/master/contexts, not here

iherman:
A formal comment: the DagCBOR is referenced to as a possibility for a special form of CBOR encoding in 6.3.3.1., which is a normative section. This means that DagCBOR must have a normative reference (and hopefully a reference that is indeed stable enough to be considered as normative).

That being said, I wonder whether the whole of section should really be normative. Normative means that it must have testable implementations of that version, and two mutually independent ones at that. It means having a separate serialization format alongside the vanilla CBOR. I wonder whether we should not instead label 6.3.3 as informative (while possibly keeping it in the document).

contexts/did-v0.11.jsonld
jonnycrunch: Mmm, not sure why the .json-ld files got included.

index.html
```diff
@@ -2766,134 +2728,114 @@ <h3>Consumption</h3>
...
+     766572696669636174696f6e4d6574686f64 # "verificationMethod"
+  81                                   # array(1)
+     78 40                             # text(64)
+        6469643a6578616d706c653a313244334b6f6f574d4864727a6377706a6264725a7335474771455241766367715833623564707550745061396f743639796577 # "did:example:12D3KooWMHdrzcwpjbdrZs5GGqERAvcgqX3b5dpuPtPa9ot69yew"
           </pre>
         </section>
+      </section>
+    </section>
+    <section>
+      <h3>Converting Data between Core Representations</h3>
```
peacekeeper, line 2815: I think this section may be problematic.. Even though CBOR-to-JSON conversion may be a special case where optimizations are possible, "converting" between representations should always go via the abstract data model (ADM). I.e. to "convert" from CBOR to JSON, you would first run the CBOR consumption process, and then the JSON production process. The end result should be the same anyway.

index.html
```diff
@@ -2766,134 +2728,114 @@ <h3>Consumption</h3>
...
+      </p>
         <ul>
-          <li>
-The keys in every map must be sorted lowest value to highest. Sorting is
-performed on the bytes of the representation of the keys.
-          </li>
-          <li>
-Indefinite-length items must be made into definite-length items.
-          </li>
+  <li>Duplicate key in the same CBOR map MUST through and Error.</li>
```
iherman, line 2733: through -> throw

OR13 (approved):

index.html
OR13: Since the core data model is defined by https://w3c.github.io/did-core/#data-model (essentially 2 sentences)... I find this paragraph extremely helpful.

If we just replace the word "JSON" with "INFRA" I think most of this still works.... I would prefer to see this section retained, with changes to make it between CBOR and INFRA, and I would argue that the other representations need a similar section or should be removed.

Suggest changing the section name to "Abstract Data Model Conversion" or similar.

@jonnycrunch I assume CBOR -> ADM -> JSON will retain all "properties", you may want to add an explicit sentence about that here.

index.html
OR13, line 2733:
```suggestion
  <li>Duplicate key in the same CBOR map MUST throw an Error.</li>
```

index.html
peacekeeper: If it's changed to say "CBOR to INFRA" instead of "CBOR to JSON", then shouldn't that simply be the "Consumption" section (which right now seems almost empty in this PR)?

index.html
```diff
@@ -2582,179 +2582,141 @@ <h2>CBOR</h2>
...
+<p>Semantic interoperability with other core representations of the DID document is possible with additional constaints 
+  to the CBOR representation.  Additionally, for lossless conversion between other core representations a deterministic 
+  representation is desirable when producing DID Documents that are represented as CBOR, in addition to the suggestions
+  in section 3.9 of the CBOR [[RFC7049]] specification for deterministic mappings, 
+  the following constraints of the DID Document model MUST be followed:
+  These additional constaints include: </p>
+
+<ul>
+    <li>Property names MUST be represented as a byte string array (major type 2) and contain only UTF-8 strings. </li>
+    <li>All member properties of a CBOR encoded DID document MUST be either represented in the `@context`, be present in the
```
peacekeeper, line 2627: Why would a CBOR representation have a `@context`? Is this the same as CBOR-LD?

jonnycrunch: @peacekeeper  

Totally valid to have `@context` in CBOR representation.   That is why I included: 

```
The property name @context MAY be used in the CBOR map and 
has the same semantic meaning as JSON-LD producers of a DID document in JSON representation.
```

 As  CBOR can support all JSON data types (including JSON-LD) for translation between the DID document model as the JSON generic data model is just a subset of the CBOR generic data model as such including the `@context`, which  gives semantics to the properties, but doesn't force json-LD processing.  My understanding of CBOR-LD is externally dictionary by reference using the `@context` in combination with an algorithm for iterating over properties and will require proper declaration to give the appropriate syntax for CBOR-LD `processing` required, something like `content-type = application/cbor+ld`.   We can do this, I think, with the current wording, but will eventually need to add that subsection on how that includes the appropriate CBOR tags and algorithm, etc.

index.html
jonnycrunch: 🙏

index.html
jonnycrunch: @OR13 ,  it would be cleaner to use CDDL as the ADM, in which case this is simple.  I am new to writing spec-text so I may have glossed over the details and would be grateful for any help.  personally, I find the definition of an Abstract Data Model, ... well too abstract, which is why I like the `Data Definition Language` that CDDL provides much more precise and something I can wrap my head around and do testing, while it does provide great support for explicit Type definition, Classes and well-defined required and optional Properties across representations. so CBOR <--> CDDL <--JSON --> and even YAML.

OR13: @jonnycrunch I completely agree, would prefer to use CDDL instead of english and handwaving at INFRA.

index.html
```diff
@@ -2766,134 +2728,114 @@ <h3>Consumption</h3>
...
+        <li>A byte string with an encoding hint (major type 6, tag number 21 through 23) is encoded as described and becomes a JSON string.</li>
+        <li>Indefinite-length items are made definite before conversion. </li>
+      </ul>
 
-          <p>
-A DID Document proof may be constructed using CBOR semantic tagging, such as
-tag 98 for CBOR Object Signing and Encryption (COSE) [[RFC8152]]
-          </p>
+      <p>
+        When interoperating with a text-based representation such as JSON or JSON-LD it is helpful to utilize additional sematics to aid convertion.  CBOR Tags numbers 21 to 23 indicate that a byte string value in DID document represented in CBOR might require a specific encoding to represent properly in JSON.
```
peacekeeper, line 2834: Language like this should really be avoided. The point of the ADM and production and consumption processes is really that you should be able to losslessly convert between ANY of the representations. If you add custom instructions for the special CBOR->JSON case, then would you also need custom instructions for every single pair of representations e.g. CBOR->YAML or CBOR->JSON-LD or CBOR->XML, etc.?

I think the Production section in this PR is really strong (perhaps minus the mention of `@context` which I don't really understand), but this section here about the special CBOR->JSON conversion case should be removed, and the "Consumption" section should be filled instead to explain how to go from a CBOR DID document to the ADM.

jonnycrunch: @peacekeeper I get where you are coming from.  My mandate was `lossless conversion` between JSON <--> CBOR and this give specific instructions.  I'm having a bit of difficulty with the Abstract Data Model as being too `abstract` for my taste.  If we were to focus on CDDL as representative of the Abstract Data Model, then this would be effortless.    CBOR <--> CDDL <--> JSON <--> CDDL <--> YAML.    Being new to spec text, I welcome help to nail this down and make it more concrete.

index.html
peacekeeper: Okay I understand that all JSON including JSON-LD can be directly mapped to CBOR. But if we define a CBOR representation of the DID document ADM, then the production rules should clearly say if a `@context` gets added or not.

- In the JSON-LD production rules, the addition of `@context` is clearly defined.
- In the JSON production rules, there is support for removing all mention of `@context` (https://github.com/w3c/did-core/pull/396).

The CBOR production rules should also make it clear whether or not `@context` gets added.

OR13: @peacekeeper why? should every section specifically state if properties related to processing directives are included or excluded?

Seems better to just say what is required, and how preserve by default works.

In a world where there are 10 representations, the approach of having each section comment about the other sections will not work.

index.html
OR13: I find this section helpful, especially given the lack of clarity regarding the ADM... I think its possible it could be refactored into the  "Data Model Section" and then referenced there, but thats probably not a good idea for this PR.

index.html
msporny: -1 to CDDL -- it is a concrete schema language, not an abstract data model. I'm concerned about the confusion around this... if we want to use CDDL to define everything, then we need to get rid of the abstract data model (which is technically possible... but it would require a fairly large rewrite of the specification at this point).

index.html
```diff
@@ -2766,134 +2728,114 @@ <h3>Consumption</h3>
...
+      </section>
+    </section>
+    <section>
+      <h3>Converting Data between Core Representations</h3>
+      <p></p>
 
-        <section>
-          <h4>COSE signatures</h4>
+    <section>
+      <h3>CBOR to JSON</h3>
```
msporny, line 2819: Yep, highly problematic because this is going from CBOR right through to JSON -- not via Abstract Data Model. if we go this route, we'll have to define NxN algorithms for each representation to go to each other representation.

msporny (changes requested):
Overall, I'm torn about this PR. It's good in that it provides a concrete and fairly detailed approach to how a CBOR encoding could work. So, many thanks to @jonnycrunch for proposing something concrete.

The problem with the PR is that it is providing a CBOR encoding with none of the advantages of CBOR. The payload size isn't greatly reduced, which is one of the primary reasons to use CBOR. I expect that there will be multiple -1s to the approach because it exposes that the group fundamentally does not understand why developers use CBOR.

index.html
```diff
@@ -139,12 +139,12 @@
         w3cid: 46729
       },
       {
-        name: "Jonathan Holt, DO, MS",
-        url: "https://www.linkedin.com/in/jonathan-holt-do-ms",
-        company: "ConsenSys Health",
-        companyURL: "https://www.consensyshealth.com",
-        w3cid: 95341
-      }
+      name: "Jonathan Holt, DO, MS",
```
msporny, line 142: Please don't add yourself as an Editor or Author of a spec. Those decisions are WG decisions and happen toward the end, when we can holistically look at who has contributed what to the specification.

OR13:
> The problem with the PR is that it is providing a CBOR encoding with none of the advantages of CBOR.

I could say the same thing of JSON (for which 0 examples exist today, at least CBOR has 1).

The question remains does this CBOR description preclude DAG_CBOR / CBOR-LD / FancyCBOR2025 ?

If it does than IMO, its a mistake, but if it does not preclude them, and defines a simple first approach to CBOR, I think we should take the PR.

index.html
OR13: I expect every developer who reads the spec to essentially do this... the ADM is the kind of complexity people avoid reading, and just guess at... in this case, the guess accurate, you can in fact convert CBOR to JSON or vice versa....without reading about INFRA or the opinions of this WG on data models... we should avoid making the "guessing" incompatible with normative statements, but I think it might also be wise to specifically call out sections like this with a disclaimer that "you have to go through the ADM to convert CBOR to JSON".

index.html
jonnycrunch: Given the amount of work I've put into this document, it only seems fair.

jonnycrunch: Given the amount that I have contributed, only seems fair.

msporny:
> I could say the same thing of JSON (for which 0 examples exist today, at least CBOR has 1).

Yes, this is true.

> The question remains does this CBOR description preclude DAG_CBOR / CBOR-LD / FancyCBOR2025 ?

It does not preclude DAG_CBOR / CBOR-LD / FancyCBOR2025 as long as it has a very specific mimetype.

I'll be sad if this goes through and grabs the application/did+cbor MIME Type...

That said, this PR is better than the current CBOR section, which DOES preclude CBOR-LD / FancyCBOR2025.

> If it does than IMO, its a mistake, but if it does not preclude them, and defines a simple first approach to CBOR, I think we should take the PR.

The question is -- will anyone find value in the representation? Other than "Hey, I implemented application/did+cbor! (and it gives me no benefit beyond application/did+json, application/did+ld+json)...

jonnycrunch:
```
The problem with the PR is that it is providing a CBOR encoding with none of the advantages of CBOR.
```

@msporny,  I disagree

According to the CBOR RFC 7049 authors including the late @jimsch, 

"Remember that CBOR prioritizes the simplicity of the encoder and decoder over the size of the encoded data."

That said, it doesn't preclude additional representation and flavors of CBOR including CBOR-LD, dagCBOR, etc which can take advantage of data compaction of CBOR.  But out of the gate, with this approach implementors get about a 25% size reduction compared to JSON approaches. This helps implementations such as myself with the cost of overhead of storing DID documents immediately and has a framework for use to move forward with additional more compact representations.

index.html
msporny: Please take a look at what others in the group have contributed to the specification... there are many others that are not listed that have (arguably) contributed more: https://github.com/w3c/did-core/graphs/contributors?from=2017-07-02&to=2020-10-08&type=c

The goal here is that everyone that should be recognized for contributing to the specification will be recognized in time, and doing so is a WG decision after (ideally) taking a data-based approach to contributions across all areas (issues, PRs, calls, etc.). We typically do this right before we publish the Proposed Recommendation (many months from now). Until then, adding and removing people from the authors list tends to lead to a lot of gnashing of teeth.

If you feel strongly about this, you could always escalate it to the Chairs to get a Chair opinion on the matter.

OR13: @brentzundel @burnburn for future reference, if we retain CBOR, jonathan holt should be listed as a contributor... potentially he should regardless, given how little contribution this note has seen from WG members.

OR13:
> I'll be sad if this goes through and grabs the application/did+cbor MIME Type...

I don't agree.... this is actually what I would expect, I would expect:

`application/did+cbor` => CBOR
`application/did+ld+cbor` => CBOR-LD
`application/did+dag+cbor` => DAG_CBOR

index.html
jonnycrunch: @OR13 🙏,  that means a lot!

index.html
```diff
@@ -2582,179 +2582,141 @@ <h2>CBOR</h2>
...
       <section>
         <h3>Production</h3>
         <p>
-When producing DID Documents that are represented as CBOR, in addition to the
-suggestions in section 3.9 of the CBOR [[RFC7049]] specification for
-deterministic mappings, the following constraints of the DID Document model
-MUST be followed:
-        </p>
+A DID document encoded in CBOR must be represented as a CBOR Map (major type 5) conforming to [[RFC7049]]. 
+Similar to the JSON encoding of a DID document, all top-level properties of the DID document MUST be represented by using the
```
msporny, line 2601:
```suggestion
Similar to the JSON encoding of a DID document, all properties of the DID document MUST be represented by using the 
```

index.html
```diff
@@ -2582,179 +2582,141 @@ <h2>CBOR</h2>
...
         <h3>Production</h3>
         <p>
-When producing DID Documents that are represented as CBOR, in addition to the
-suggestions in section 3.9 of the CBOR [[RFC7049]] specification for
-deterministic mappings, the following constraints of the DID Document model
-MUST be followed:
-        </p>
+A DID document encoded in CBOR must be represented as a CBOR Map (major type 5) conforming to [[RFC7049]]. 
+Similar to the JSON encoding of a DID document, all top-level properties of the DID document MUST be represented by using the 
+property name as the name of the member of the CBOR map.  The values of properties of the data model described in <a href="#core-properties"> § 5. Core Properties</a>, including
```
msporny, line 2602: We should point to the CBOR spec for all type definitions.

index.html
```diff
@@ -2582,179 +2582,141 @@ <h2>CBOR</h2>
...
-deterministic mappings, the following constraints of the DID Document model
-MUST be followed:
-        </p>
+A DID document encoded in CBOR must be represented as a CBOR Map (major type 5) conforming to [[RFC7049]]. 
+Similar to the JSON encoding of a DID document, all top-level properties of the DID document MUST be represented by using the 
+property name as the name of the member of the CBOR map.  The values of properties of the data model described in <a href="#core-properties"> § 5. Core Properties</a>, including 
+all extensions, MUST be encoded in CBOR [[RFC7049]] by mapping property values to CBOR types as follows: </p>
+
+<ul>
+ <li><a href="https://tools.ietf.org/id/draft-ietf-cbor-7049bis-12.html#section-3.2.2">Numeric</a> values representable as IEEE754 MUST be represented as a Number type (major type 0 or 1).</li>
```
msporny, line 2606: All of the links to CBOR are non-normative, they point to the BIS version, which is not an IETF standard yet. We need to point to the existing specification.

msporny: This specific line states that floating point numbers should be expressed as integers... which would lose all precision for those IEEE754 values. We need to be clear on how floats are expressed, how integers are expressed, and what the acceptable value ranges are if an implementation hopes to round-trip information between different languages -- ECMA defines these ranges, see the JSON spec for how they talk about acceptable ranges.

index.html
```diff
@@ -2582,179 +2582,141 @@ <h2>CBOR</h2>
...
+<li><a href="https://tools.ietf.org/id/draft-ietf-cbor-7049bis-12.html#section-3.2.2">Boolean</a> values MUST be represented as a Boolean literal (primitive type 21 for true and primitive type 20 for false).  </li>
+<li><a href="https://tools.ietf.org/id/draft-ietf-cbor-7049bis-12.html#section-3.2.2">Sequence</a> value MUST be represented as an Array type (major type 4) .</li>
+<li><a href="https://tools.ietf.org/id/draft-ietf-cbor-7049bis-12.html#section-3.2.2">Unordered sets</a> of values MUST be represented as an Array type (major type 4).</li>
+<li><a href="https://tools.ietf.org/id/draft-ietf-cbor-7049bis-12.html#section-3.2.2">Sets of properties</a> MUST be represented as an map type (major type 5).</li>
+<li><a href="https://tools.ietf.org/id/draft-ietf-cbor-7049bis-12.html#section-3.2.2">Empty values</a> MUST be represented as a null literal (primitive type 22). </li>
```
msporny, line 2611: I'm not certain that the abstract data model supports null values... I don't think we have anything in the spec on expressing null values (as we haven't had that need). So, the option here is to stay silent or to define what a null value is in the abstract data model. I suggest we stay silent, we don't have a use case for a null value.

index.html
```diff
@@ -2582,179 +2582,141 @@ <h2>CBOR</h2>
...
+<li><a href="https://tools.ietf.org/id/draft-ietf-cbor-7049bis-12.html#section-3.2.2">Other values</a> MUST be represented as a String type. (major type 5)</li>
+<li><a href="https://tools.ietf.org/id/draft-ietf-cbor-7049bis-12.html#section-3.2.2">Indefinite-length items</a> MUST be made definite legth.</li>
```
msporny, line 2613:
```suggestion
<li><a href="https://tools.ietf.org/id/draft-ietf-cbor-7049bis-12.html#section-3.2.2">Indefinite-length items</a> MUST be made definite length.</li>
```

index.html
```diff
@@ -2582,179 +2582,141 @@ <h2>CBOR</h2>
...
         <p>
-When producing DID Documents that are represented as CBOR, in addition to the
-suggestions in section 3.9 of the CBOR [[RFC7049]] specification for
-deterministic mappings, the following constraints of the DID Document model
-MUST be followed:
-        </p>
+A DID document encoded in CBOR must be represented as a CBOR Map (major type 5) conforming to [[RFC7049]]. 
+Similar to the JSON encoding of a DID document, all top-level properties of the DID document MUST be represented by using the 
+property name as the name of the member of the CBOR map.  The values of properties of the data model described in <a href="#core-properties"> § 5. Core Properties</a>, including 
+all extensions, MUST be encoded in CBOR [[RFC7049]] by mapping property values to CBOR types as follows: </p>
```
msporny, line 2603: This section should really be talking about how the INFRA values map to CBOR... we don't define any of the CBOR values in the spec. The mappings are INFRA -> CBOR, not CBOR Abstract Type -> CBOR representation.

index.html
```diff
@@ -2582,179 +2582,141 @@ <h2>CBOR</h2>
...
+<li><a href="https://tools.ietf.org/id/draft-ietf-cbor-7049bis-12.html#section-3.2.2">Sequence</a> value MUST be represented as an Array type (major type 4) .</li>
+<li><a href="https://tools.ietf.org/id/draft-ietf-cbor-7049bis-12.html#section-3.2.2">Unordered sets</a> of values MUST be represented as an Array type (major type 4).</li>
+<li><a href="https://tools.ietf.org/id/draft-ietf-cbor-7049bis-12.html#section-3.2.2">Sets of properties</a> MUST be represented as an map type (major type 5).</li>
+<li><a href="https://tools.ietf.org/id/draft-ietf-cbor-7049bis-12.html#section-3.2.2">Empty values</a> MUST be represented as a null literal (primitive type 22). </li>
+<li><a href="https://tools.ietf.org/id/draft-ietf-cbor-7049bis-12.html#section-3.2.2">Other values</a> MUST be represented as a String type. (major type 5)</li>
+<li><a href="https://tools.ietf.org/id/draft-ietf-cbor-7049bis-12.html#section-3.2.2">Indefinite-length items</a> MUST be made definite legth.</li>
+<li><a href="https://www.rfc-editor.org/rfc/rfc3339.txt">Datetime </a> MUST be encoded in UTF-8 (major type 3). This datetime value MUST be normalized to UTC 00:00, as indicated by the trailing "Z".</li>
```
msporny, line 2614: This raises the number of options for Datetime to 3... ISO8601, RFC3339, and XMLDateTime -- we need to pick one as a group.

index.html
```diff
@@ -2582,179 +2582,141 @@ <h2>CBOR</h2>
...
+</ul>
+
+
+<p>Semantic interoperability with other core representations of the DID document is possible with additional constaints
```
msporny, line 2618:
```suggestion
<p>Semantic interoperability with other core representations of the DID document is possible with additional constraints 
```

index.html
```diff
@@ -2582,179 +2582,141 @@ <h2>CBOR</h2>
...
+
+<ul>
+    <li>Property names MUST be represented as a byte string array (major type 2) and contain only UTF-8 strings. </li>
+    <li>All member properties of a CBOR encoded DID document MUST be either represented in the `@context`, be present in the 
+      <a href="https://www.w3.org/TR/did-spec-registries/">DID spec registries</a> or be explicitly documented as an extension of the DID document conformant to <a href="#extensibility"> § 4.3. Extensibility</a>.  </li>
+    <li>The byte string array's length in bytes is represented following the rules for positive integers (major type 0).  </li>
+    <li>Property names in each CBOR map MUST be unique. </li>
+    <li>Integer encoding MUST be as short as possible.</li>
+    <li>The expression of lengths in CBOR major types 2 through 5 MUST be as short as possible. </li>
+    <li>All floating point values MUST be encoded as 64-bits, even for integral values.</li>
```
msporny, line 2633: Will need to discuss loss of precision at 56-bits in environments like node.js / javascript / browser.

index.html
```diff
@@ -2582,179 +2582,141 @@ <h2>CBOR</h2>
...
-          <li>
-The expression of lengths in CBOR major types 2 through 5 MUST be as short as
-possible.
-          </li>
-          <li>
-All floating point values MUST be encoded as 64-bits, even for integral values.
-          </li>
-        </ul>
+<p>
+  The property name `@context` MAY be used in the CBOR map and has the same semantic meaning as JSON-LD producers of a DID document in JSON representation.
```
msporny, line 2640: Strange for CBOR to talk about JSON-LD. While the statement is true, it should probably just stay silent on this item.

index.html
```diff
@@ -2582,179 +2582,141 @@ <h2>CBOR</h2>
...
+</p>
+
+<p>
+  All properties of the DID document MUST be included in the root map (major type 5).
```
msporny, line 2644: This statement is a bit strange as it repeats something stated earlier in the document. I don't quite understand what you're going for here.

index.html
```diff
@@ -2766,134 +2728,114 @@ <h3>Consumption</h3>
...
+("=") are removed from the encoded string.  Tag number 22 suggests
+conversion to classical base64 encoding (Section 4 of [[RFC4648]] ), with
+padding as defined in [[RFC4648]].  For both base64url and base64,
+padding bits are set to zero (see Section 3.5 of [[RFC4648]] ), and
+encoding is performed without the inclusion of any line breaks,
+whitespace, or other additional characters.  Tag number 23 suggests
+conversion to base16 (hex) encoding, with uppercase alphabetics (see
+Section 8 of [[RFC4648]]).  Note that, for all three tag numbers, the
+encoding of the empty byte string is the empty text string of other representations.
+</p>
```
msporny, line 2766: I don't understand what I would need to do as an implementer with this text. How does my implementation change? I don't think any of this PR suggests that we use tags 21-23, so why are we talking about them here?

index.html
jonnycrunch: I think this is in-line with the conclusion of last Thursday's special topic call.  `@context` MAY exist in `application/json` DID documents, NO json-ld processing is expected.  Same is true of the `application/cbor` DID document.

index.html
msporny: There is also a section in the CBOR spec that speaks to this -- can't we just refer to that section? 

https://tools.ietf.org/html/rfc7049#section-4

index.html
jonnycrunch: just being explicitly clear on production rules for CBOR representation.  Once we have an unambiguous ADM then it should be defined there.

msporny (changes requested):
The general structure is fine, there are some details in the PR that need to be sorted. We probably also need to define a MIME Type for application/did+cbor, but that can happen in another PR.

index.html
jonnycrunch: I think this is even more constrained as it is XMLDateTime with timezone set to `Z`.

index.html
jonnycrunch: I think INFRA is still lacking in some areas for all mappings.

index.html
jonnycrunch: CBOR tags facilitate a `lossless` conversion between core representations with use of semantic sugar.  It would be helpful to implementors who need to represent binary data and give `batteries included` clear instruction for how to `losslessly` convert to other representations during DID Resolution.   This section highlights the method to convert between core representations.

index.html
jonnycrunch: will do

index.html
jonnycrunch: You are correct, thanks for catching that, floating point numbers are major type 7.  will update and point to normative version.  As I mentioned numbers are a pain and I think we just need to settle on a representation.

index.html
jonnycrunch: I suppose the counter argument is `What happens when a DID document does not contain a required property after conversion to a different core representation?`.  Seem like Null would be appropriate then to explicitly state that this required value as determined in the ADM isn't there.

index.html
jonnycrunch: agreed.

index.html
jonnycrunch: Done, but it just highlighted all of the problems of the ADM including the lack of clarity on number representation that we discussed and the gap with INFRA.  I added a bunch more details regarding bigNum and BigFloat and yet we don't need those types in the specification.  If we want to dwell on these, then I can make `lossless` conversions between CBOR and JSON via an ADM (ideally in CDDL), but it may be a exercise in futility.

index.html
jonnycrunch: The language for floating points is that to enforce precision even when the floating point value is equivalent to an integral value, ie. `1.0` is still `1.0` not `1`, which is just explicitly stating how to handle numbers and thus achieve `lossless` conversion while retaining canonical representation.

index.html
jonnycrunch: Turns out there is a value of `undefined`, which may be more appropriate. But, I'm open for additional discussion on how to handle this in the context of `unknown` properties.  I'm also more in line with @OR13, that it is critical to explicitly drop unknown properties rather than `preserve`/`retain` them as this will lead to a buffer overflow and definite security vulnerability.

index.html
jonnycrunch: I added a bunch more explicit declarations for numbers, again an area that INFRA is lacking.

index.html
jonnycrunch: see:  https://github.com/w3c/did-core/issues/435

index.html
msporny: Agree with @OR13 -- what is being suggested by @peacekeeper is not a scalable approach; it would require representations to know about other representations, which is strange.

index.html
msporny: Yes, /at that time/ we should add @jonnycrunch ... not before. If we're going to start adding people to Editors and authors, we're going to have to discuss other people as well... which will block this PR.

Please remove the author credit for now to unblock the PR, please.

msporny:
> "Remember that CBOR prioritizes the simplicity of the encoder and decoder over the size of the encoded data."

I was skeptical of this statement, so decided to go and look at how simple CBOR encoders/decoders are wrt. JSON encoders/decoders. Turns out that CBOR encoders/decoders are more complex, sometimes by an order of magnitude from JSON encoders/decoders. So, while that was a design goal of the CBOR WG, they failed to achieve something simpler than JSON in the end due to all the variational stuff done to ensure that encoding size was optimized (within reason).

For example, there are many JSON encoder/decoders that are implemented purely as C header files, and small ones for microcontrollers:

421 sloc -- https://github.com/rafagafe/tiny-json
1367 sloc -- https://github.com/cesanta/frozen 

The smallest CBOR encoder/decoders I found were multiple times that size and complexity:

785 sloc -- https://github.com/cabo/cn-cbor
2275 sloc -- https://github.com/intel/tinycbor

All that to say that the statement is a wash... you can do simple constrained encoder/decoders in JSON, JSON-LD, and CBOR. This isn't something that is only applicable to CBOR, it's applicable to each representation wrt. encoder/decoder simplicity.

index.html
jonnycrunch: @msporny CDDL is a normative Data Definition Language and is more than just a schema check for concrete representations.  CDDL provides a machine-processable notation to express CBOR (and JSON) data structures. It  is used in many specifications and other design documents for describing the structures that one party is meant to send to another, enabling informed discussion, input validation, and even code generation. It also works for both JSON (including JSON-LD) and CBOR.  I would like everyone to seriously consider CDDL. 

My understanding regarding what we are attempting to solve with the Abstract Data Model is class of objects whose logical behavior is defined by a set of values and a set of operations.  

CDDL allows us to unambiguously define classes of object with logical behaviors and works with JSON and CBOR.  CDDL in combination with sets of operations defined in the Core Spec achieves this goal.   

I would also invite @cabo to weigh in his thoughts and appreciate any advice he may have.  As I mentioned, I was anticipating feedback from Jim Schaad (@jimsch) who unfortunately died last month.

index.html
jonnycrunch: @msporny I'm actually already listed as a contributor,  this PR only updates my affiliation.  Given the amount that I have contributed, it seems only fair.

index.html
cabo: We are pretty rich in data modeling and poor in information modeling formal languages.
Sometimes a data modeling language, with some understanding of its limitations, can stand in for an information modeling language.
I haven't analyzed this document, so I don't have an opinion yet.  
But I do know that CDDL has helped in other cases.
(There are converters from CDDL to the formats proposed on json-schema.org, see how we made use of automatic conversion in https://www.ietf.org/archive/id/draft-onedm-t2trg-sdf-00.html#name-json-schemaorg-rendition-of -- we had to model CDDL features by generating two versions of the json-schema.org format.)

index.html
cabo: Please refer to https://www.ietf.org/archive/id/draft-ietf-cbor-7049bis-16.html which is already in RFC-EDITOR stage and probably will have an RFC number before you are done.

msporny:
@jonnycrunch -- apologies for the delay in merging... all three times that I've sat down to re-review this PR, I've been pulled away. I hope to do a final review on this during the week.

index.html
msporny: *sigh* -- I missed that. Looks that it happened here: https://github.com/w3c/did-core/commit/7a4bac4a3da280c03961c960030aea527e4135fa ... which never should have gotten into the spec.

I had asked that you remove your name at that point as well and thought you had done that:

https://github.com/w3c/did-core/pull/282/files#diff-0eb547304658805aad788d320f10bf1f292797b5e6d745a3bf617584da017051R127-R132

 I'll raise the issue with the chairs and see how they want to proceed.

burnburn: @jonnycrunch Manu is right that individual contributors should NEVER add themselves to the author/editor/etc. lists.  These determinations are complex and can best be handled as we go to CR or even PR.  @manu , since Jonathan's name is currently there please update his affiliation. @jonnycrunch understand that the lists of editors, authors, and contributors will be updated later and your name may shift to a different list or drop off as appropriate at that time.  I am speaking in my official capacity as Co-chair.

msporny: @burnburn wrote:
> @msporny, since Jonathan's name is currently there please update his affiliation.

Acknowledged, will do.

brentzundel: I second what @burnburn has said.
The final determination of who will be listed as an author/editor/etc. has yet to be made and is complex. The editors and chairs will be the ones to make the determination, which will be based primarily on data-driven evaluation of contributions.

index.html
brentzundel: It may be better to talk about what to do when consuming properties that are not defined in DID Core nor the registries, rather than specifically mentioning a particular property from another representation.

index.html
```diff
@@ -2766,134 +2729,118 @@ <h3>Consumption</h3>
...
-          <li>
-The keys in every map must be sorted lowest value to highest. Sorting is
-performed on the bytes of the representation of the keys.
-          </li>
-          <li>
-Indefinite-length items must be made into definite-length items.
-          </li>
-        </ul>
+  <li>Duplicate key in the same CBOR map MUST throw an Error.</li>
+  <li>All member properties of a CBOR encoded DID document MUST be either represented in the `@context`, be present in the
```
brentzundel, line 2735: That this CBOR representation actively supports an `@context` property is strange to me.

index.html
brentzundel: I understand the desire to provide a means for possible lossless conversion to another representation, but am unsure what this description has to do with CBOR Extensibility. Perhaps the section should be re-named?

index.html
brentzundel: Whether or not CDDL should replace INFRA as the language of the ADM is a question that needs to be resolved by the working group. Until such time as the working group decides to use CDDL in the ADM, I agree with @OR13 that replacing "JSON" with "INFRA" is the appropriate thing to do here. And I agree with @peacekeeper that a description of CBOR -> INFRA belongs in the consumption section.

index.html
brentzundel: I second @msporny 's concerns. 
Representation descriptions should define production and consumption rules to and from the ADM, and should not include translation rules between specific representations.

index.html
brentzundel: While this information may be valuable, the mandate was:

> Clear specification in DID Core for producing and consuming CBOR such that it can be "losslessly converted" (via the abstract data model) to the other representations.

Any additional information or guidance for how to better ensure that the CBOR values may be converted into JSON might fit better in the implementation guide.

index.html
brentzundel: Perhaps this section could be as simple as
"Conversion from the CBOR representation to another supported representation is a process of using the CBOR consumption rules defined here, followed by the using the production rules for the chosen representation. Conversely, conversion from another representation to CBOR is a process of following that representation's consumption rules, followed by using the CBOR production rules defined here."

brentzundel (commented):
I think this PR introduces a lot of great content, but that the organization of that content needs some work.
I leave it to the editors to decide if it would be better to fix the organizational issues before or after merging this PR.

index.html
jonnycrunch: I think that it would be helpful to flush out the Abstract Data Model and I'd like to make my case that CDDl as a Data Definition Language can stand in for the data model aspects of the ADM as I have demoed before.  Happy to present next week for the TPAC sessions next week to make my argument.  @cabo, as I have discussed with the Chair and will their approval, I'd like to extend an invitation to be an external Invited Expert and help use through this dilemma.   Despite being invited, there is a formal [process](https://www.w3.org/participate/invited-experts/) to accept and join our working group.

index.html
jonnycrunch: well, the ADM using JSON-LD is strange to me as evident by the `@type` issue that we are currently having.    @peacekeeper,  I think that from the discussion regarding ADM that we agreed that `@context` is optional in the JSON representations. It is only when you combine it with MIME = `application/did+json+ld that you explicitly state that JSON-LD processing is required.

OR13: Technically JSON-LD processing is still not required with `application/did+json+ld` per the spec language today.

index.html
jonnycrunch: I would like to argue for CDDL to facilitate the data representation aspects of the Abstract Data Model and thus JSON <--> CDDL <--> CBOR.  Infra just isn't doing it and I think CDDL will.

index.html
jonnycrunch: interesting, thanks for the clarification!

index.html
jonnycrunch: updated, see #361

jonnycrunch:
@msporny [section 4 of RFC 7049](https://tools.ietf.org/html/rfc7049#section-4) give non-normative advice about converting between JSON and CBOR.  It is up to spec implementers to be clear about what rules they want to explicitly convert between the two formats. Seems that this group is converging on an Abstract Data Model that I believe is too abstract to be effectively implemented.  It would be great to formalize the ADM (including whether or not that includes the '@context`) then work on clear and precise rules such as some of the ones that I suggested in this PR.   

with regard to MIME type: my previous PR had it included, but seems to have been dropped when I merged upstream changes to the document.  I'll update it to include `application/did-cbor'.  Thanks for catching that.

cabo:
> with regard to MIME type: my previous PR had it included, but seems to have been dropped when I merged upstream changes to the document. I'll update it to include `application/did-cbor'. Thanks for catching that.

You may want to consider structured syntax suffixes (RFC 6313, see also Section 7.5 of RFC 7049).  This would make something like application/did+cbor the media type name.

Grüße, Carsten

msporny:
Now that #455 has been merged, I'm going to try to focus on this PR and work with @jonnycrunch to get it across the line.

My suggestion here is:

1. For @jonnycrunch to fix the merge conflicts in this PR as well as the merge history to reflect what he wants to go into the main branch.
2. For @jonnycrunch to rework the PR in whatever way he sees fit based on #455 and #454. 
3. For @jonnycrunch to mark the entire section as a work in progress via an issue marker.
4. For the Editors to merge the PR as-is, as it's an improvement on the current CBOR section, and put in PRs to update the CBOR section to align with other data model changes in the specification.

I want to be very clear that this section isn't done, we're not getting good engagement to improve it (aside from the work that @jonnycrunch has done on it) and changes are going to have to be made to get it into shape for CR. I'm willing to do the work to make that happen... thus the request to merge as-is after @jonnycrunch makes a final pass over the PR.

Folks should object to this path forward for this PR if they have alternatives that may help us get better progress on the vanilla CBOR representation.

OR13:
Merge conflict and suggested changes need to be incorporated so we can get this in.

talltree:
I too am in favor of merging this as soon as the merge conflict and suggested changes can be attended to. The sooner the better.

index.html
```diff
@@ -2766,134 +2729,118 @@ <h3>Consumption</h3>
...
+  81                                   # array(1)
+     78 40                             # text(64)
+        6469643a6578616d706c653a313244334b6f6f574d4864727a6377706a6264725a7335474771455241766367715833623564707550745061396f743639796577 # "did:example:12D3KooWMHdrzcwpjbdrZs5GGqERAvcgqX3b5dpuPtPa9ot69yew"
           </pre>
         </section>
+      </section>
+    </section>
+    <section>
+      <h3>Converting Data between Core Representations</h3>
+      <p></p>
```
peacekeeper, line 2821:
```suggestion

      <div class="warning" title="Converting via DID document data model">
        <p>
This section conflicts with other parts of the specification, since
conversion has to be defined not between any two representations directly,
but rather in terms of parsing and serializing via the abstract
DID document data model (see <a href="#data-model"></a> and 
<a href="#representations"></a>). The content of this section can
potentially be adapted and re-used in other sections such as the one
about the CBOR representation (see <a href="#cbor"></a>).
        </p>
      </div>
```

peacekeeper (changes requested):
I propose to add this warning per the various comments in this PR, to point out that conversion must happen via the abstract data model, not between any two representations directly. With this added, I'd be happy to approve this PR.

msporny:
Ping @jonnycrunch -- still waiting on merge conflicts to be resolved and the PR to be brought up to date with the rest of the specification.

jonnycrunch:
updated CBOR section with particular focus on CBOR <- ADM -> JSON language as per reviewer suggestions and merge conflicts with head branch.  Please let me know if there are additional details that I missed.

msporny:
There are still merge conflicts, possibly due to previously merged PRs, please fix the merge conflicts.

@jonnycrunch while you're fixing the merge conflicts, can you clean up the merge history so I don't have to squash-merge... namely, these commits (which modified files that shouldn't have been modified): 

Delete package.json - jonnycrunch committed on Sep 29 - 7755fc1 
Delete did-v1.jsonld - jonnycrunch committed on Sep 29 - a9dabdf 
Delete did-v0.11.jsonld - jonnycrunch committed on Sep 29 - 77f4c39 

There are also a lot of "Update index.html" commits that either need to have better commit comments, or be rolled up into a single commit. It's optional if you want to do this or not... if you don't do it, I'll just squash-merge everything together.

index.html
```diff
@@ -2806,216 +2799,180 @@ <h2>CBOR</h2>
...
-  id : did-url
-  type : text
-  serviceEndpoint : uri
-  ? description : text
-  * tstr => any
-}
-      </pre>
+<p>
+  The property name `@context` MAY be present in the CBOR representation of a DID Document and if present SHOULD be ignored as this property is reserved for JSON-LD processing.
+</p>
```
jricher, line 2861: This block should be removed as a similar statement is removed from the JSON section in another PR and we shouldn't add them. If anything, this should just be a note that it might show up, but no normative language here as it's not a defined property.

index.html
jricher: I agree that the conversion section should be removed from here.

index.html
jricher: We definitely should not be defining a direct conversion like this, or encouraging people to do this kind of thing. This is a semantically-driven specification and so the ADM is the only defined method of translation. Anything else (including treating JSON-LD as plain JSON) is liable to work based on luck.

jricher (changes requested):

index.html
msporny: Agreed, the Editors have discussed items such as this in the PR and will clean it up in another PR given how long this PR has been hanging out there. At this point, it's an improvement on what is in the specification and the Editors will do their best to align it with the rest of the specification while giving folks the opportunity to object/revise in a future PR.

index.html
msporny: Agreed this this section doesn't make sense given the direction of the specification. We have an ADM for a reason and general rules for production/consumption as well. The Editors will be cleaning these sections up in future PRs.

msporny (approved):

msporny:
Normative, multiple reviews, changes requested and made, Editors will align this section with the rest of the specification over the next several weeks, merging as a work in progress to avoid objections over the details.

PR #422, opened 2020-09-30
Fix echidna

OR13:
https://github.com/w3c/did-core/blob/master/w3c.json#L5
https://github.com/w3c/did-core/blob/master/ECHIDNA#L2
https://github.com/w3c/did-core/blob/master/index.html#L34

I believe these all need to match, for echidna to work correctly.

### DIFF
--- a/w3c.json
+++ b/w3c.json
@@ -1,6 +1,10 @@
- {
-    "group":      [117488]
-,   "contacts":   ["iherman"]
-,   "repo-type":  "rec-track"
-,   "shortName":  "did-spec"
+{
+    "group": [
+        117488
+    ],
+    "contacts": [
+        "iherman"
+    ],
+    "repo-type": "rec-track",
+    "shortName": "did-core"
 }


msporny:
Looks like Echidna is working correctly w/ @rhiaro's fixes? https://www.w3.org/TR/did-core/

@iherman -- the change @OR13 is suggesting in this PR seems correct... is there some backend W3C system that still refers to the spec as did-spec?

iherman:
> @iherman -- the change @OR13 is suggesting in this PR seems correct... is there some backend W3C system that still refers to the spec as did-spec?

Actually, it is the other way round: more and more of W3C's backend systems _rely_ on the content of this file, i.e., this file _must_ be correct. And @OR13 is absolutely right: it was not. (Which may have been my mistake, I should have changed it when we changed the repo and document name. This probably means I should be burnt on the stake...)

iherman (approved):

iherman:
> Looks like Echidna is working correctly w/ @rhiaro's fixes? https://www.w3.org/TR/did-core/

All that being said, I'm indeed not sure that this was the problem with Echidna. But regardless, `w3c.json` must be correct and it is not.

msporny (approved):

msporny:
Editorial, bug fix, multiple reviews, objectors will be burnt at the stake :), merging.

Issue #402, opened 2020-09-17
Handling additional errors in resolution/dereferencing


clehner:
Three error codes are defined for DID resolution and DID URL Dereferencing:
- `invalid-did`
- `unauthorized`
- `not-found`

Source:
- [§ 8.1.2 DID Resolution Metadata Properties](https://w3c.github.io/did-core/#did-resolution-metadata-properties)
- [§ 8.2.2 DID URL Dereferencing Metadata Properties](https://w3c.github.io/did-core/#did-url-dereferencing-metadata-properties)
- [did-spec-registries § 6.2 *error*](https://w3c.github.io/did-spec-registries/#error-0)

What should an implementation do if it encounters errors during resolution or dereferencing other than the three specified? For example:
- a DID resolves to a supposed JSON-encoded DID, but the JSON is invalid
- a DID resolves to a supposed DID document, but the document doesn't fit the Data Model, such as by missing a required property like "id", or would have other errors produced during processing
- a transient network failure prevents resolving the DID or dereferencing the URL
- the DID method or URL scheme is not implemented

Should implementers make up additional *error* values as needed? Or return additional errors outside the metadata structure?

Is this an appropriate place for a question like this?

peacekeeper:
@clehner Good question, I actually wanted to raise a similar issue myself. When we added the DID Resolution and DID URL Dereferencing sections, we said at some point that we would still have to decide on the exact error values.

My answer would be that we should only define a minimal set of error values in the DID Core spec, and that additional error values can be defined in other places, such as the DID Resolution spec.

Based on what we currently have, my preference would be to do this:

- Remove the `unauthorized` value. (Because it would be very hard to define what it means to authorize a resolution process; in fact we usually consider resolution something that doesn't require authentication/authorization of the client).
- Add a `deactivated` value. (This would be returned if the DID has been deactivated).
- Maybe add a `method-not-supported` value. (To indicate that the DID resolver doesn't support this particular DID method).

So we would then have the following error values defined in DID Core: `invalid-did`, `not-found`, `deactivated`, (maybe) `method-not-supported`.

Again, additional error values can be defined later using the DID Spec Registries, but the above seem like a useful "core set" of errors to me.

msporny:
The PR has been merged into the main specification. This issue will be closed in 7 days unless there are objections.

kdenhartog:
No objections made and 7 days have passed. Closing this issue.

PR #440, opened 2020-10-19
Remove "unauthorized" and add "deactivated" error code.

peacekeeper:
Addresses https://github.com/w3c/did-core/issues/402.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/440.html" title="Last updated on Oct 29, 2020, 10:02 PM UTC (ddd8bb5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/440/5dd10b5...ddd8bb5.html" title="Last updated on Oct 29, 2020, 10:02 PM UTC (ddd8bb5)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -3447,18 +3447,18 @@ The <a>DID</a> supplied to the <a>DID resolution</a> function does not conform
 to valid syntax. (See <a href="#did-syntax"></a>.)
                         </dd>
                         <dt>
-unauthorized
-                        </dt>
-                        <dd>
-The caller is not authorized to resolve this <a>DID</a> with this <a>DID
-resolver</a>.
-                        </dd>
-                        <dt>
 not-found
                         </dt>
                         <dd>
-The <a>DID resolver</a> was unable to return the <a>DID document</a>
+The <a>DID resolver</a> was unable to find the <a>DID document</a>
 resulting from this resolution request.
+                        </dd>
+                        <dt>
+deactivated
+                        </dt>
+                        <dd>
+The <a>DID</a> supplied to the <a>DID resolution</a> function has been
+deactivated. (See <a href="#deactivate"></a>.)
                         </dd>
                     </dl>
                 </dd>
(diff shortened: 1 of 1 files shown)


peacekeeper:
The question of resolution/dereferencing error codes started earlier this year when we added the DID Resolution section. Based on the discussions back then, and the comments in issue https://github.com/w3c/did-core/issues/402, the idea is that we include a minimal set of error codes in DID Core, and allow additional codes to be assigned via the DID Spec Registries in the future.

With this PR, we would have:

- For DID resolution: `invalid-did`, `not-found`, `deactivated`
- For DID URL dereferencing: `invalid-did-url`, `not-found`, `deactivated`

kdenhartog (commented):
I like adding the deactivated error, and I think it would remain beneficial to keep unauthorized as well. In the case of resolving dids from a public resolver connected to a private VDR (such as an intranet web server with did:web or a private hyperledger fabric network) I think it's beneficial to indicate the DID exists, but the resolver cannot return the DID Document without some form of authorization provided in the request. This bodes well for a zero trust architecture where the resolver service is acting as an authorization point for a resource (the DID Document).

With that in mind, this does open the question do we need to be able to provide authorization proof in the resolution options parameter?

peacekeeper:
I feel pretty strongly that in the DID Core spec we should not talk about authorization of DID resolution and therefore remove the error code. This doesn't mean that such an error code and also some kind of authorization input metadata is not possible or not allowed, those things can be done and even added later to the DID Spec Registries. But if we want to address this now in DID Core, it would be a major new feature that we don't have time to discuss. As far as I remember, by default, we have always considered DID resolution to not require authorization.

If we want to specify this at some point, we could define the required metadata and error codes elsewhere, e.g. in DID Resolution, and then register those extensions in the DID Spec Registries.

agropper:
+1 Markus. I think this is a complicated issue in terms of privacy. Happy
to engage if the group thinks it’s important but we would need to develop
some use cases to discuss further.

msporny (approved):

OR13 (approved):

brentzundel (approved):

peacekeeper:
Merge conflicts should be resolved now.

msporny (approved):

msporny:
Normative, multiple positive reviews, no objections, merging.

Issue #403, opened 2020-09-18
DID Core editorial review (2)


iherman:
Here is the second (and last) part of my editorial review.

* [x] (Generic comment) maybe we should remove the references to closed issues from the document (e.g., Issue  204, referred to from §6.1.1)

## 5. Core properties

* [x] If, as a reader, I have to find the normative definition of the properties, I would look at the table of content to find that. But there is a bunch of inconsistencies:
    * Some section titles are simply the name of the property (e.g., "authentication") others are not (e.g., "DID Subject", "Service Endpoint")
    * Some properties are not in their own subsection like `id`, `type` (also `alsoKnownAs` but there is a separate PR to take care of that).
    * Some properties are actually shared among others, like `id` and `type` but are not separately defined
    * The only place where there is a reference to all core properties is in the introductory paragraph of §5, but that is not clear from the TOC 

    I think what is missing is a clear, formal specification of the full DID Core vocabulary, with all the properties and classes (types), their shape constraints, domain and range definitions, etc. That specification should be at one place in the document with links to the section where these terms are defined in English prose. That vocabulary would also be defined, for the LD world, as a formal, RDFS/OWL ontology in its own namespace. The [Web Annotation Vocabulary](https://www.w3.org/TR/annotation-vocab/) might be a good example for what I am referring to, although that vocabulary is many time larger and more complex than DID Core.

    (Note that the definition should _*not*_ be made  using `rdf:domain` and `rdf:range` to define constraints. Those statements are _licenses to infer_ and _not a constraint_. If we want to define constraints we either must use English prose, or formalism like [SHACL](https://www.w3.org/TR/shacl/). Or both, actually.)

* [x] This is related to the previous comment, but is a comment in itself. For a clean ontological definition of the verification methods and verification relationships, I wonder whether it is not worth defining, formally, something like a `VerificationMethod` type, which would be a supertype for the types like `Ed25519VerificationKey2018`, and which would also act as a clean range constraint for methods like `authentication` as well as _part of_ the domain constraints for properties like `controller` or `id`.  All this should be part of the aforementioned vocabulary specification. (Clearly, this is not really an editorial comment only. Sorry about that…)

## 5.3.1 Key types and formats

[...]

PR #448, opened 2020-10-27
references: DID Spec Registries informative

rhiaro:
Pointed out by @iherman in #403 this is not a rec-track document, so needs to be an informative rather than normative reference.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/448.html" title="Last updated on Oct 27, 2020, 7:12 PM UTC (7439cd3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/448/5dd10b5...7439cd3.html" title="Last updated on Oct 27, 2020, 7:12 PM UTC (7439cd3)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -925,7 +925,7 @@ parameter is OPTIONAL. If present, the associated value MUST be an
 Implementers as well as <a>DID method</a> specification authors MAY use
 additional DID parameters that are not listed here. For maximum
 interoperability, it is RECOMMENDED that DID parameters use the official W3C
-DID Specification Registries mechanism [[DID-SPEC-REGISTRIES]], to avoid
+DID Specification Registries mechanism [[?DID-SPEC-REGISTRIES]], to avoid
 collision with other uses of the same DID parameter with different semantics.
         </p>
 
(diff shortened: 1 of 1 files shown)


rhiaro:
Having re-skimmed some of the W3C process docs about normative references, we might be okay to normatively reference the Registries given that we're just saying stuff should be added there, rather than depending on a particular functionality defined therein. Also worth checking the emerging 'living document' stuff for clarity too.

iherman:
@rhiaro it may be that with a careful formulation the ambiguities can be reduced; can one of you try it?

The living document is a different issue. Today (ie, process 2020) this is meant for rec-track documents, ie, it does not apply here. 2021 will probably make it possible to publish a somewhat special class of document for registries which may get a normative 'weight'. I doubt that this will be at our disposal before the end of the WG's life, but it is not impossible. We can cross that particular bridge when we get there...

msporny (approved):

msporny:
Purely editorial, presumably at least two reviews, no objections, merging.

Issue #401, opened 2020-09-17
DID Core editorial review (1)


iherman:
I spent some time today on thoroughly re-reading the core spec. I did not complete the whole document, only up to §5.3.1, but I thought I would dump my issues here, and continue on another day with the rest.

There are lots of mini issues, and many (most?) of them are minor editorial issues. Some are more complex, but still editorial I believe.

## Abstract

* [x] "DIDs are URLs": Strictly speaking, they are not. They are URI-s and there is a thing called DID URL… This is only an abstract, but it should still be precise…
* [x] "A DID document might contain the DID subject itself (e.g. a data model).": I do not understand this statement. The DID subject is defined as: _"The entity identified by a DID and described by a DID document. A DID has exactly one DID subject. Anything can be a DID subject: person, group, organization, physical thing, digital thing, logical thing, etc."_ But the document cannot contain a person…

## Introduction

* [ ] Just a food for thought: wouldn't it be a better style to use a neutral form in this section? I.e., "Because the user controls" instead of "Because we control", for example?
* [x] "This specification does not require any particular technology or cryptography to underpin the generation, persistence, resolution or interpretation of DIDs." I am not sure this is well formulated. The _specification_ does not require, but implementations _do_ require a bunch of particular technologies. I think the intention here is to say something like "This specification does not _depend_ on any particular technology…"
* [x] "(e.g., blockchain, distributed ledger, ...)": isn't blockchain a specific form of distributed ledger? In this context I would propose to remove the reference to blockchain, as too specific.
  
## 1.2 Design Goals (Table)

* [x] Introductory paragraph: the second sentence sounds like repeating the essence of the first one, with the repeated term "design goals for this specification".
* [ ] **Decentralization**, "Eliminate the requirement for centralized authorities or single point failure in identifier management...": playing possible opponent here I could say that the _specification_ of, say, the HTTP URL scheme does not _require_ the existence of a DNS either and, similarly, the _specification_ of a DID does not _eliminate_ this. I think it is better to say that the definition of DID-s "makes it possible" or "makes it easier" (or something like that) to be implemented without the requirement of central authorities.
* [ ] **Interoperability** and **Portability**: both of these statements are, though true, a bit too obvious for my taste. These statements are true for all Standards, W3C or otherwise. (I would think just removing these two would be fine.)
* [ ] **Simplicity**: this again is some sort of a generic, feel-good statement. Let alone the fact that simplicity is a matter of personal opinion, and others may not agree that DIDs are simple...
  
[...]

PR #450, opened 2020-10-27
terms: add 'immutable' to DLT definition

rhiaro:
Suggested by @iherman in #401. Obviously not an expert, my understanding is a DLT is kinda shoulda supposed to be immutable but under certain situations this might not technically be the case; but for the purposes of this definition stating it as Ivan suggests I think is probably right. Look, now this comment took longer to write than making the change itself. Just close the PR if it's not appropriate to add "immutable" anyway.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/450.html" title="Last updated on Nov 11, 2020, 2:19 PM UTC (dba8371)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/450/b426d48...dba8371.html" title="Last updated on Nov 11, 2020, 2:19 PM UTC (dba8371)">Diff</a>

### DIFF
--- a/terms.html
+++ b/terms.html
@@ -213,11 +213,13 @@ function for a given <a>DID URL</a> or <a>DID document</a>.
   <dt><dfn data-lt="distributed ledger technology|DLT">distributed ledger</dfn> (DLT)</dt>
 
   <dd>
-A <a href="https://en.wikipedia.org/wiki/Distributed_database">distributed database</a>
-in which the various nodes use a
-<a href="https://en.wikipedia.org/wiki/Consensus_(computer_science)">consensus protocol</a>
-to maintain an immutable shared ledger in which each transaction is cryptographically
-signed and chained to the previous transaction.
+A non-centralized system for recording events. These systems establish
+sufficient confidence for participating users to rely upon the data recorded
+by others to make operational decisions. They typically use distributed
+databases where different nodes use a consensus protocol to confirm the
+ordering of cryptographically signed transactions. The linking of digitally
+signed transactions over time often makes the history of the ledger
+effectively immutable.
   </dd>
 
   <dt><dfn data-lt="proofPurpose">proof purpose</dfn></dt>
--- a/terms.html
+++ b/terms.html
@@ -216,7 +216,7 @@ function for a given <a>DID URL</a> or <a>DID document</a>.
 A <a href="https://en.wikipedia.org/wiki/Distributed_database">distributed database</a>
 in which the various nodes use a
 <a href="https://en.wikipedia.org/wiki/Consensus_(computer_science)">consensus protocol</a>
-to maintain a shared ledger in which each transaction is cryptographically
+to maintain an immutable shared ledger in which each transaction is cryptographically
 signed and chained to the previous transaction.
   </dd>
 
--- a/terms.html
+++ b/terms.html
@@ -214,7 +214,7 @@ function for a given <a>DID URL</a> or <a>DID document</a>.
 
   <dd>
 A non-centralized system for recording events. These systems establish
-sufficient confidence for participating users to rely upon the data recorded
+sufficient confidence for participants to rely upon the data recorded
 by others to make operational decisions. They typically use distributed
 databases where different nodes use a consensus protocol to confirm the
 ordering of cryptographically signed transactions. The linking of digitally


iherman (approved):
I am obviously o.k. with it, but it is not my voice that counts...

peacekeeper (approved):

jandrieu:
-1
Even DLTs are not "immutable". This is an unfortunate term that has caused considerable hype and grief.

Perhaps "distributed" addresses your goal better:
> to maintain an distributed ledger in which each transaction is cryptographically

iherman:
@jandrieu I am obviously not an expert, and I am not sure what the optimal term is. However, in our world isn't the fact that a node put on the DLT is not supposed to change an essential feature?

msporny:
While there are corner cases where what @jandrieu says is true, DLTs are viewed to be statistically immutable in the general case. Could you elaborate more on your concern, @jandrieu? I don't think your rewording addresses @iherman's initial issue raised.

jandrieu:
Hmmm.

I didn't realize my suggestion would end up defining DLTs in terms of distributed ledgers, which definitely doesn't work.

So, I would suggest just not adding "immutable".

My concern is both technical and cultural. 

DLTs are not, in fact, immutable. They fork all the time. Sometimes as a result of intentional manipulation. At best, what is established is a probabilistic conviction that it is economically infeasible for attackers to make more money subverting the chain for a double spend than they would make by mining. 

In many cases, this economic feasibility argument fails--either because there are not enough interested computation to keep the system honest or because the goals of the attacker are not economic. So, only the best DLTs actually achieve ANY kind of immutability--and it still based on probabilistic economics rather than factual immutability. Many, perhaps most DLTs are constantly at risk of this type of attack, which is why foundational layer 1 ledgers are deemed so valuable: the feedback loop of value increases the security profile so valuable chains get more valuable. So, technically, it is not only incorrect to say a DLT is immutable, it is at best aspirational hopes and at worst willful disinformation for those DLTs that have not achieved the right economic balance. 

Conditional immutability is a goal, not a fact, of DLTs.

Culturally, the problem is that "immutable" was the "sizzle" in many businesses' pitch for ICOs and blockchain fairy dust. As such, it has acquired a patina of huksterism both within the technical communities working on blockchain tech as well as the investors and customers that bought into the hype of "immutability". 

So, although well-intended, I think it diverts more than it adds to the conversation.

msporny:
> So, only the best DLTs actually achieve ANY kind of immutability--and it still based on probabilistic economics rather than factual immutability.

Ah, so here is the point of contention -- the above is not true for Paxos, RAFT, Veres One's Continuity consensus algorithm as well as the Hashgraph consensus algorithm. There are consensus algorithms with provable finality. Bitcoin and Ethereum don't have it, the best they can do is statistical, but there are a whole class of consensus algorithms that do achieve finality. 

That's what I was trying to point out... your statement doesn't hold for the entire class of technologies.

jandrieu:
Indeed. There are exceptions. Which means that arguably, *some* DLTs are immutable. Some aren't.

msporny:
> Indeed. There are exceptions. Which means that arguably, _some_ DLTs are immutable. Some aren't.

Ah, ok, we're on the same page then... you said this before:

> DLTs are not, in fact, immutable.

... which is what raised my eyebrow. 

Eyebrow is now lowered again.

msporny:
Soooooo... this means that this PR should be closed w/o being merged?

iherman:
> Soooooo... this means that this PR should be closed w/o being merged?

Well... I go back to my original problem that led to this issue here. I am a layperson and read this in the terminology of our spec:

> **distributed ledger (DLT)**
> A distributed database in which the various nodes use a consensus protocol to maintain a shared ledger in which each transaction is cryptographically signed and chained to the previous transaction.

For that layperson reader I believe one of the main point of using a DLT is that if I create something that is put on a DLT, that something cannot change any more. It is cast in concrete, marble, titanium, whatever, but I (or somebody) else cannot go and change it. The "which each transaction is cryptographically signed and chained to the previous transaction" is a technical mumbo-jumbo that does not help the reader to grasp the essence of it.

I frankly do not care whether the term "immutable" is the right technical one here. But I still believe part of an essential information gets lost.

msporny:
@iherman wrote:
> I frankly do not care whether the term "immutable" is the right technical one here. But I still believe part of an essential information gets lost.

Got it -- that's a reasonable expectation -- @jandrieu thoughts and concrete language that address @iherman's concerns?

jandrieu:
It's not our job to explain why DLTs are useful. We're just trying to define them in a way that is accurate. "immutable" is not accurate.

Here's the existing language:
> distributed ledger (DLT)
> A distributed database in which the various nodes use a consensus protocol to maintain a shared ledger in which each transaction is cryptographically signed and chained to the previous transaction.

This is accurate and complete.

I don't think we want to put marketing buzzwords like "immutable" in a terminological definition in a specification.

msporny:
What about this, @jandrieu, @iherman:

> distributed ledger (DLT)
> A non-centralized system for recording events. These systems often achieve sufficient trust among those that use the system to make decisions based on the information in the system. In more technical terms, solutions often consist of distributed databases in which the various database nodes in the system use a consensus protocol in which each transaction is cryptographically signed and chained to the previous transaction, often making the history of the ledger effectively immutable.

iherman:
Works for me, @msporny

peacekeeper:
"**append-only** ledger"?

Just an idea, no strong opinion here...

jandrieu:
+1 @msporny The basics of that are great. An iteration:
> distributed ledger (DLT)
> A non-centralized system for recording events. These systems establish sufficient confidence for participating users to rely upon the data recorded by others to make operational decisions. They typically use distributed databases where different nodes use a consensus protocol to confirm the ordering of cryptographically signed transactions. The linking of digitally signed transactions over time often makes the history of the ledger effectively immutable.

OR13 (approved):

kdenhartog (approved):

rhiaro:
I've updated the definition to use @jandrieu's iteration on @msporny's text https://github.com/w3c/did-core/pull/450#issuecomment-722515284

Sorry to those who recently approved. Not sure if you were approving the original change or Joe's update. I'll re-request review just in case.

terms.html
```diff
@@ -213,11 +213,13 @@
   <dt><dfn data-lt="distributed ledger technology|DLT">distributed ledger</dfn> (DLT)</dt>
 
   <dd>
-A <a href="https://en.wikipedia.org/wiki/Distributed_database">distributed database</a>
-in which the various nodes use a
-<a href="https://en.wikipedia.org/wiki/Consensus_(computer_science)">consensus protocol</a>
-to maintain a shared ledger in which each transaction is cryptographically
-signed and chained to the previous transaction.
+A non-centralized system for recording events. These systems establish
+sufficient confidence for participating users to rely upon the data recorded
```
msporny, line 217:
```suggestion
sufficient confidence for participants to rely upon the data recorded
```

msporny (approved):
Minor editorial nit, otherwise LGTM.

peacekeeper (approved):

kdenhartog (approved):
Happy with the new definition as well!

brentzundel (approved):

msporny:
Editorial, multiple positive reviews, changes requested and made, no objections, merging.

PR #459, opened 2020-11-16
editorial: remove biometrics examples

rhiaro:
Personal preference to remove biometrics examples from the spec. This is too dystopian, and I don't think we should be encouraging it even if it's technically possible. It's also dancing close to the PII and bound-forever problems (even though the example refers to an endpoint, rather than biometric data in the DID doc). I've replaced the example with something less specific about credentials, but also happy to see a better more specific example suggested instead.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/459.html" title="Last updated on Nov 30, 2020, 11:41 AM UTC (9631fd3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/459/b426d48...9631fd3.html" title="Last updated on Nov 30, 2020, 11:41 AM UTC (9631fd3)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -270,8 +270,7 @@ The Decentralized Identifiers (DIDs) defined in this specification are a new
 type of globally unique identifier designed to enable individuals and
 organizations to generate our own identifiers using systems we trust, and to
 prove control of those identifiers (authenticate) using cryptographic proofs
-(for example, digital signatures, privacy-preserving biometric protocols, and so
-on).
+(for example, digital signatures).
     </p>
     <p>
 Because we control the generation and assertion of these identifiers, each of us
@@ -1438,8 +1437,8 @@ usage, it verifies that the signer possessed the associated private key.
 Verification methods might take many parameters. An example of this is a set
 of five cryptographic keys from which any three are required to contribute to
 a threshold signature. Methods need not be cryptographic. An example of this
-might be the contact information for a biometric service provider that compares
-a purported <a>DID controller</a> against a candidate biometric vector.
+might be the contact information for a verifiable credential verifier that
+verifies a credential presented by a purported <a>DID controller</a>.
       </p>
 
       <p>
--- a/index.html
+++ b/index.html
@@ -1436,10 +1436,7 @@ usage, it verifies that the signer possessed the associated private key.
       <p>
 Verification methods might take many parameters. An example of this is a set
 of five cryptographic keys from which any three are required to contribute to
-a threshold signature. Methods need not be cryptographic. An example of this
-might be the use of a service endpoint that returns a verifiable credential
-for the purported <a>DID Controller</a> which the <a>DID Controller</a> can
-provide as proof of control.
+a threshold signature. Methods need not be cryptographic.
       </p>
 
       <p>
--- a/index.html
+++ b/index.html
@@ -1437,8 +1437,9 @@ usage, it verifies that the signer possessed the associated private key.
 Verification methods might take many parameters. An example of this is a set
 of five cryptographic keys from which any three are required to contribute to
 a threshold signature. Methods need not be cryptographic. An example of this
-might be the contact information for a verifiable credential verifier that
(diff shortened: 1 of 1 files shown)


jandrieu (approved):
Agreed. I also don't believe we have two implementations that do this, so this section would be at risk in any case.

Do we know of any implementations that claim biometric verification of any kind?

peacekeeper:
Hmm I agree in general with not encouraging (or actively discouraging) things that can be dangerous and misused. But I also believe it's important to not lose the idea that verification methods can be very flexible and not limited to cryptographic keys.

I think there will be non-dystophian use cases where for some reason you are unable to use cryptographic keys, but you still want a decentralized identifier and a way to prove control over it (including by using biometrics).

So I'd be fine with removing the example, as long as it's still possible to do it.

peacekeeper:
Besides, I don't really understand the new example. Is this saying that anyone who can present a certain (bearer?) credential will have proven control over the DID?

Isn't it often the other way round, i.e. you have to prove control over a DID before a verifier will accept your credential as valid?

msporny (approved):

msporny:
> Besides, I don't really understand the new example. Is this saying that anyone who can present a certain (bearer?) credential will have proven control over the DID?

I believe the new text is saying that a DID Document can provide a verification process (that is not cryptographic in nature), where a verifier could perform that process (like picking up the phone and calling someone and doing some Knowledge-based process where the DID controller knows the answers to the verifier's questions because they have a physically printed card with the answers on it plus a pin code, for example).

I do agree that the explanation is a bit vague. @rhiaro, any chance you could provide a more concrete example? I'm going to give this another 3 days and merge it if we don't have an update.

peacekeeper:
> I believe the new text is saying that a DID Document can provide a verification process (that is not cryptographic in nature)

Yes I think that's great, and I like the example of picking up the phone, etc., but it would be good to improve the changed text a bit. Right now it says that "contact information for a verifiable credential verifier" is itself a verification method for DIDs, which doesn't make much sense to me.

rhiaro:
Yeah you're right, that text is not good. I had thought about removing it altogether because I wasn't sure of a specific alternative example. I was actually thinking about a generic out-of-band VC verification process, or possibly a VC verification service endpoint? Ideally this would be an example of a real world thing that actually happens today rather than something imaginary. I'd appreciate a further nudge in the right direction.

index.html
```diff
@@ -1438,8 +1437,8 @@ <h2>Verification Methods</h2>
 Verification methods might take many parameters. An example of this is a set
 of five cryptographic keys from which any three are required to contribute to
 a threshold signature. Methods need not be cryptographic. An example of this
-might be the contact information for a biometric service provider that compares
-a purported <a>DID controller</a> against a candidate biometric vector.
+might be the contact information for a verifiable credential verifier that
```
kdenhartog, line 1440:
```suggestion
might be through the usage of a service endpoint that can be used to fetch a verifiable credential from the <a>DID Controller</a> so that the <a>DID Controller</a> can provide proof of control via a verifiable credential.
```

kdenhartog: @rhiaro this is an attempt to reword it. It still feels a bit messy to me, but thought it might help nudge you towards what you were describing.

rhiaro:
I worked on @kdenhartog's suggestion in https://github.com/w3c/did-core/pull/459/commits/5870811c45268bb329c9cc1f2e381aa77a0aef9a but I still don't think I could get it to make sense. I removed it altogether in the next commit. I think it's better to be short an example than have a confusing or made up one, and I don't think the paragraph loses a lot without it. I'll add one back as soon as I can find something properly concrete, but that shouldn't hold up this PR.

kdenhartog (approved):

msporny:
Editorial, changes requested and made, multiple positive reviews, no objections, merging.

PR #465, opened 2020-11-19
Change MUST to SHOULD for lossless conversion

peacekeeper:
I think this is simply a mistake.. During the Amsterdam F2F we said that representations may support extension mechanisms that only work in a specific representation, and that can't be losslessly converted to others.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/465.html" title="Last updated on Dec 13, 2020, 8:25 PM UTC (c8fba4b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/465/26ad2eb...c8fba4b.html" title="Last updated on Dec 13, 2020, 8:25 PM UTC (c8fba4b)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1262,7 +1262,9 @@ that ensures that two different representations will be able to work together.
       <li>
 Representations MAY define other extensibility mechanisms including methods for
 decentralized extensions. Such extension mechanisms SHOULD support lossless
-conversion into any other conformant representation.
+conversion into any other conformant representation. Extension mechanisms for a 
+representation SHOULD define a mapping of all properties and representation syntax 
+into the abstract data model and its type system.
       </li>
       </ol>
       <p class="note">
--- a/index.html
+++ b/index.html
@@ -1261,7 +1261,7 @@ that ensures that two different representations will be able to work together.
         </li>
       <li>
 Representations MAY define other extensibility mechanisms including methods for
-decentralized extensions. Such extension mechanisms MUST support lossless
+decentralized extensions. Such extension mechanisms SHOULD support lossless
 conversion into any other conformant representation.
       </li>
       </ol>


msporny (approved):

kdenhartog (approved):

index.html
```diff
@@ -1261,7 +1261,7 @@ <h2>Extensibility</h2>
         </li>
       <li>
 Representations MAY define other extensibility mechanisms including methods for
-decentralized extensions. Such extension mechanisms MUST support lossless
+decentralized extensions. Such extension mechanisms SHOULD support lossless
 conversion into any other conformant representation.
```
jricher, line 1265: I would be more comfortable with this if we also say they have to tell us at least SOME way to get things into and out of the data model, even if it's just because of the default type rules. Otherwise we could end up with ambiguous behaviors due to extensions.

I'm not sure if this captures it but I'd recommend something like this:

```suggestion
decentralized extensions. Such extension mechanisms SHOULD support lossless
conversion into any other conformant representation. Extension mechanisms for a 
representation MUST define a mapping of all properties and representation syntax 
into the abstract data model and its type system.
```

peacekeeper: Hmm I'm wondering about cases such as CBOR where AFAIK you can have numbers as keys in a map. Would we say that a CBOR DID document representation that uses this CBOR-specific feature is not conformant, because the data model only supports property names that are strings?

@jonnycrunch interested in your opinion on this..

msporny: CBOR where you have numbers as keys in a map is only true BEFORE consumption or AFTER production. The representation rules would state how you consume numbers-as-keys in the particular CBOR representation to the ADM, so, AFAICT, there isn't an issue here, @peacekeeper.

jricher: The conversion rules for CBOR (not yet written) would need to explicitly say how to map member names (which can be numbers or strings) into property names (which need to be strings). Similar logic needs to be written for JSON-LD, where the value of `context` can change the semantic mapping of an object member name such that it might map to a different property name.

msporny: @peacekeeper ... waiting on you to merge this PR. You either need to accept @jricher's suggestion, or provide an alternate that @jricher is happy with.

peacekeeper: I dont really agree with this. It should be allowed to have representation-specific extensions that don't fit into the ADM. In the CBOR example, I don't think it's possible to define ADM production/consumption rules that properly deal with numbers-as-keys. 

Of course such extensions that don't fit in the ADM can't be losslessly converted to other representations, but that has always been part of the model.

So I'd prefer to merge the PR without applying @jricher 's suggestion (maybe open a separate issue for that afterwards?)

brentzundel (approved):

index.html
peacekeeper: We could keep your extra sentence, but change MUST to SHOULD in that extra sentence as well.

msporny: Waiting on @jricher to agree to @peacekeeper's suggestion, provide a different modification, etc.

msporny:
Please retarget this PR to the `main` branch. You can do this via the drop-down at the top of the page beside the `w3c:master` target branch indicator for the pull request.

index.html
peacekeeper: I propose to commit this suggestion, but with MUST changed to SHOULD. @jricher ?

msporny: Applying suggestion by @peacekeeper and merging we've been waiting 18 days for a response and haven't gotten one. It's a single word change, if there are objections, a future PR can fix this easily.

msporny (approved):

msporny:
Normative, multiple reviews, changes requested and made (and one timed out for a response), no objections, merging.

PR #467, opened 2020-11-21
respec: fix restrictReferences

rhiaro:
Less of a 'fix' and more of a complete rewrite.

`restrictRefs` is now triggered by `postProcess` in the respec config. It removes any definitions from the terminology section that are not actually used in the spec.

It is not smart enough to detect circular definition cycles within the terminology section itself, but I've eyeballed it and don't think there are any.

There *are* definitions that are *only* referred to from other definitions (and not in the main body of the spec) - but we can't drop these without removing their reference or we get respec errors. Will do another editorial pass over terminology at some point to reduce this.

There's a really hacky bit to stop the invisible 'definition panels' triggering the local-refs-exist linter, because I didn't figure out how to remove the dfns before the DOM is built. But for the time being it works..

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/467.html" title="Last updated on Nov 21, 2020, 1:45 AM UTC (dcd30fd)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/467/26ad2eb...dcd30fd.html" title="Last updated on Nov 21, 2020, 1:45 AM UTC (dcd30fd)">Diff</a>

### DIFF
--- a/common.js
+++ b/common.js
@@ -3,7 +3,6 @@
 /*  ... stolen from Gregg Kellogg of the JSON-LD 1.1 Working Group */
 /*  ... who stole it from Manu Sporny of the JSON-LD 1.0 Working Group */
 /*  ... who stole it from Shane McCarron, that beautiful, beautiful man. */
-/* exported restrictReferences */
 var ccg = {
   // Add as the respecConfig localBiblio variable
   // Extend or override global respec references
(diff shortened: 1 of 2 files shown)


msporny (approved):
Haven't tested it yet, but if it works, ship it. Anything is better than the broken form we have currently. Thank you @rhiaro for working on this... I tried and failed a few weeks ago.

msporny:
Bug fix, a review, no objections, doubt people will object to fixing a bug, merging.

PR #454, opened 2020-11-08
Add representation language to address syntax and "unknown" properties.

dlongley:
Here's an attempt at getting generic/unknown/syntax-related property representation rules into the spec per the resolutions/discussions this past week.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/454.html" title="Last updated on Dec 1, 2020, 6:25 PM UTC (35e6c03)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/454/b86bed4...35e6c03.html" title="Last updated on Dec 1, 2020, 6:25 PM UTC (35e6c03)">Diff</a>

PR #472, opened 2020-12-01
(editorial) minor language refinement

TallTed:
Speaking into the void is never pleasant. There was no comment in response to [my suggested language change in #454](https://github.com/w3c/did-core/pull/454#discussion_r529983390), and it would be lost entirely if I were not saying something now.

To reiterate, I suggested that --
```
Note that no additional processing will be applied to the value of the 
<code>@context</code> object member, if present. The original value will 
be added verbatim to the abstract data model.
```
-- would be clearer than (line breaks added) --
```
Note that the value of the <code>@context</code> object member will be
ignored, if present. That is, this field will not have additional 
processing applied to its value and will be added verbatim to the 
abstract data model.
```

The major reason for this change is that "ignored" simply isn't correct here, as — even with the subsequent clarifying statement — "ignored" implies dropping the value on the floor, not carrying it forward to the abstract data model.

Upon re-re-re-re-reading both of the above, I believe this would be better still, and this is what I've put in this PR --
```
Note that the <code>@context</code> object member, if present, will not have
additional processing applied to its value, which will be added verbatim to the
abstract data model.
```

***
<a href="https://pr-preview.s3.amazonaws.com/TallTed/did-core/pull/472.html" title="Last updated on Dec 1, 2020, 8:13 PM UTC (57e2c45)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/472/11622f3...TallTed:57e2c45.html" title="Last updated on Dec 1, 2020, 8:13 PM UTC (57e2c45)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2628,7 +2628,9 @@ data-cite="INFRA#convert-a-json-derived-javascript-value-to-an-infra-value">JSON
 consumption rules</a> in the [[INFRA]] specification.
         </p>
         <p>
-Note that the value of the <code>@context</code> object member will be ignored, if present. That is, this field will not have additional processing applied to its value and will be added verbatim to the abstract data model.
+Note that the <code>@context</code> object member, if present, will not have
+additional processing applied to its value, which will be added verbatim to the
+abstract data model.
         </p>
       </section>
 


msporny (approved):
Editorial. I expect that this language is going to be removed in a future PR, but if it's not, this change is fine.

Issue #463, opened 2020-11-18
Naming classes of properties and syntax


peacekeeper:
https://github.com/w3c/did-core/pull/455 introduced the concept of "representation properties". Per comments in that PR, we still wanted to discuss the exact terminology. I propose to distinguish properties and syntax according to 1. whether they are core, registered, or unregistered, and 2. whether they are representation-independent or representation-specific.

See this matrix:

![properties and syntax](https://user-images.githubusercontent.com/290941/99601146-217a6200-29ff-11eb-885d-b88cfabf6331.png)

Additional notes:

- We re-organize the DID Spec Registries to no longer use the term "Base properties" (since this isn't defined anywhere)
- We also introduce the term "representation-foreign syntax", but that's not a separate class. It's the same class as "representation-specific syntax" if that syntax show up in another representation (e.g. `@context` in a YAML DID document).

If people agree with this, I can do PRs to update DID Core and DID Spec Registries.

OR13:
is there an example of "representation-independent syntax" ?

I think its probably safe to take a stab at a PR for this regardless.

peacekeeper:
> is there an example of "representation-independent syntax" ?

No there is no such thing, that would be a contradiction.

Based on the above diagram, we would have:

- Core Properties (representation-independent): `id`, `alsoKnownAs`, `verificationMethod`, `service`, ...
- Registered Properties (repesentation-independent): includes Core Properties, and also: `ethereumAddress`, ...
- Unregistered Propertes (repesentation-independent): `foo`, ...

and:

- Core Syntax (representation-specific): `@context`
- Registered Syntax (representation-specific): includes Core Syntax, and also: (nothing yet)
- Unregistered Syntax (representation-specific): `%YAML`, `xmlns`, ...

talltree:
@peacekeeper, you nailed it. Great diagram (as always), and I'm totally aligned with your proposed terminology.

One question: do you think we need to add these to the terminology section? My gut feeling is no, but that we should introduce them (and I would even vote to include your diagram) in the Abstract Data Model section.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2019/did-wg/Meetings/Minutes/2020-11-24-did#section3) on 2020-11-24)

- no resolutions were taken

<details><summary><i>View the transcript</i></summary>

### 3. PR 454 and 455... 

[...]

PR #474, opened 2020-12-06
Add JSON-LD Production and Consumption rules.

msporny:
Add JSON-LD Production and Consumption rules by just referring to the JSON production rules, and layering `@context` above those rules.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/474.html" title="Last updated on Dec 13, 2020, 8:27 PM UTC (ea2e8e9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/474/3d76cea...ea2e8e9.html" title="Last updated on Dec 13, 2020, 8:27 PM UTC (ea2e8e9)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1255,8 +1255,8 @@ that ensures that two different representations will be able to work together.
       <li>
 Representations MAY define other extensibility mechanisms including methods for
 decentralized extensions. Such extension mechanisms SHOULD support lossless
-conversion into any other conformant representation. Extension mechanisms for a 
-representation SHOULD define a mapping of all properties and representation syntax 
+conversion into any other conformant representation. Extension mechanisms for a
+representation SHOULD define a mapping of all properties and representation syntax
 into the abstract data model and its type system.
       </li>
       </ol>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -2682,9 +2686,10 @@ <h3>
         </h3>
 
         <p>
-The <a>DID document</a> MUST be serialized as a JSON document, with one
-additional requirement: The <a>DID document</a> MUST include the
-<code>@context</code> property.
+The <a>DID document</a> MUST be serialized according to the
+<a href="#production">production rules for JSON</a>, with one additional
+requirement: The <a>DID document</a> MUST include the <code>@context</code>
+property.
```
jricher, line 2692: Given the discussion around terminology, we probably don't want to call this a "property" anymore, or at the very least not just "property" without qualifiers.

jricher:
An important additional rule is that the context defines how to interpret the JSON object member names into a "property name" and therefore an ADM property definition with associated type data. This makes me wonder if "property name" should be in the conversion tables for all representations as its own entry, since all representations do need to define how that works.

index.html
msporny: Agree with "at the very least not just 'property' without qualifiers. That discussion is happening in an issue right now (can't remember the number off of the top of my head). Once we bike shed that name, the Editors will go through the spec and fix all occurences.

msporny:
> This makes me wonder if "property name" should be in the conversion tables for all representations as its own entry, since all representations do need to define how that works.

Yes, I wondered that too. At present, "all property names MUST be strings" is stated for at least 2 of the 3 representations, I'm getting a PR together to get all representations to state that in the introductory information for production/consumption.

In any case, we do need to state this somewhere for all representations, ideally in a clean and straightforward manner.

index.html
msporny: Given that we don't know when that discussion is going to be resolved, would you object to making that change in a future PR?

jricher: If it's tracked in a separate issue that's fine, I just don't want that detail to get lost when we move forward.

jricher:
The property names in the ADM are strings, but that doesn't mean they have to be strings in the representation. As long as they can be mapped deterministically, any value is fine in the representation. JSON-LD specifically can map JSON object member names to ADM property names by processing the document's `context` value as well as the the string value of the object member name -- the object member names are never taken in a vacuum.

brentzundel (approved):

OR13 (approved):

index.html
msporny: Tracked in https://github.com/w3c/did-core/issues/463

msporny:
Normative, multiple reviews, no changes requested (but issues raised and tracked), no objections, merging.

PR #475, opened 2020-12-06
Remove duplicate data conversion section.

msporny:
This removes the data conversion section in the CBOR section as it duplicates the Representation->Abstract Data Model->Representation rules defined in the Representation section.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/475.html" title="Last updated on Dec 13, 2020, 8:30 PM UTC (2022d43)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/475/f1b5b00...2022d43.html" title="Last updated on Dec 13, 2020, 8:30 PM UTC (2022d43)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -3031,44 +3031,6 @@ Use no CBOR tags other than the CID tag (42)
         </section>
       </section>
     </section>
-    <section>
-      <h3>Converting Data between Core Representations</h3>
-      <p>Transformation between various core representations SHOULD be performed through parsing and serializing via the Abstract DID Document Data Model (see <a href="#data-model"></a> and <a href="#representations">Production and Consumptions rules between core representation via the Abstract Data Model</a>).
-        Given that CBOR is a superset of JSON, the following non-normative advice is given for converting a DID document between CBOR and JSON.  </p>
-
-      <div class="warning" title="Converting via DID document data model">
-        <p>
-This section conflicts with other parts of the specification, since
-conversion has to be defined not between any two representations directly,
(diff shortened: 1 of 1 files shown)


OR13 (approved):

jricher (approved):

kdenhartog (approved):

brentzundel (approved):

msporny:
Hey @jonnycrunch are you ok with this PR?

jonnycrunch:
Yeah, I'm `OK` with it, but it would be nice to leave it as a non-normative Note to help implementers with some guidance.

msporny:
> Yeah, I'm `OK` with it, but it would be nice to leave it as a non-normative Note to help implementers with some guidance.

The issue with the section is that it is not only duplicates instructions on how you should do conversion, but it also suggests that going directly from CBOR to JSON is where we had consensus in the group (which is not where we had consensus). The group came to consensus on the point that to do conversion, you go from a representation to the abstract data model to another representation. This section doesn't say that and provides guidance that goes against guidance given further up in the specification.

We probably /do/ want to say that as long as you get the same result as going from representation_A -> ADM -> representation_B that you're good to go... which would allow people to "convert directly from CBOR to JSON". What we don't want to do is give different guidance that could diverge from what we have in the spec.

msporny:
Normative, multiple reviews, changes suggested and responded to, no objections to this particular PR, merging.

Issue #464, opened 2020-11-19
DID Core must define how to compute the digest to secure did spec registry entries contexts


OR13:
https://github.com/w3c/did-spec-registries/pull/146

This is pretty straightforward,  we need to explain how you use curl + SHA-256, to get a hash, and compare it with an expected value.

If we don't do this, folks will be registering multihashes, MD5, Blake2b, etc... and its won't be useful at all...

this is an option:

```
curl -s https://www.w3.org/ns/did/v1  | openssl sha256
```

msporny:
This feels like a reasonable way to provide guidance for how developers should generate a hash.

msporny:
This issue will be closed once #485 is merged.

msporny:
PR #485 has been merged, closing.

PR #485, opened 2020-12-12
Add advisement on computing the cryptographic hash of a context

OR13:
Addresses https://github.com/w3c/did-core/issues/464

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/485.html" title="Last updated on Dec 17, 2020, 2:24 PM UTC (1bb6dfa)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/485/16ba782...1bb6dfa.html" title="Last updated on Dec 17, 2020, 2:24 PM UTC (1bb6dfa)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2755,6 +2755,13 @@ machine-readable information about the context. These URIs SHOULD be
 associated with a cryptographic hash of the content of the JSON-LD Context as
 part of registration in the DID Specification Registries [[?DID-SPEC-REGISTRIES]].
             </p>
+            <p class="advisement">
+              If included, the cryptographic hash of the content of the JSON-LD Context, MUST be computed in a manner equivalent to:
+              <pre class="example">
+                curl -s https://www.w3.org/ns/did/v1#910cd6648f6f7b72a7896a8da83e63460eb8355a0af4b56b699c9281452ac8bb  | openssl sha256
+              </pre>
+              The integrity check is considered valid if and only if the hex digest of the resource matches the fragment.
+            </p>
           </dd>
         </dl>
         <p>
--- a/index.html
+++ b/index.html
@@ -2760,7 +2760,7 @@ If included, the cryptographic hash of the content of the JSON-LD Context MUST b
 computed in a manner equivalent to the following:
             </p>
             <pre class="example">
-curl -s https://www.w3.org/ns/did/v1  | openssl sha256
+curl -s https://www.w3.org/ns/did/v1 | openssl sha256
             </pre>
             <p>
 The command above will result in the following hash, expressed in hexadecimal notation:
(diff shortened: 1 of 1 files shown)


OR13:
Personally, I think it might be better to remove the language regarding computing hashes of contexts, but if it remains, it must be defined in sufficient detail to be useful.

index.html
```diff
@@ -2755,6 +2755,13 @@ <h3>
 associated with a cryptographic hash of the content of the JSON-LD Context as
 part of registration in the DID Specification Registries [[?DID-SPEC-REGISTRIES]].
             </p>
+            <p class="advisement">
+              If included, the cryptographic hash of the content of the JSON-LD Context, MUST be computed in a manner equivalent to:
+              <pre class="example">
+                curl -s https://www.w3.org/ns/did/v1#910cd6648f6f7b72a7896a8da83e63460eb8355a0af4b56b699c9281452ac8bb  | openssl sha256
+              </pre>
+              The integrity check is considered valid if and only if the hex digest of the resource matches the fragment.
```
msporny, line 2763: This particular line isn't needed is it? We're not telling people how to write their software... just how to register their contexts?

msporny (changes requested):
Overall LGTM... minor corrections to tag nesting (`p` and `pre` are block level HTML elements and can't be put inside each other)... removed fragment as there is no normative spec that states how to integrity check URLs -- could use Hashlink here? Or IPFS? None of it could be normative, though.

msporny:
> Personally, I think it might be better to remove the language regarding computing hashes of contexts, but if it remains, it must be defined in sufficient detail to be useful.

Yeah, the more I think about it, the more I think it would be better to remove the context hashes until we can point to a spec that can refer to them in a way that is normative (with normative processing rules). Hashlink could do that, but it's years away from being a standard.

index.html
OR13: I think this line is necessary to explain how they can register their contexts.

an alternative would be to tell them to register the URL and integrity digest separately.... which seems like a missed opportunity for making things easy, for both humans and computers.

index.html
```diff
@@ -2755,6 +2755,19 @@ <h3>
 associated with a cryptographic hash of the content of the JSON-LD Context as
 part of registration in the DID Specification Registries [[?DID-SPEC-REGISTRIES]].
             </p>
+            <p class="advisement">
+              If included, the cryptographic hash of the content of the JSON-LD Context MUST be computed in a manner equivalent to the following:
```
TallTed, line 2759: white space, line length
```suggestion
If included, the cryptographic hash of the content of the JSON-LD Context MUST be
computed in a manner equivalent to the following:
```

index.html
```diff
@@ -2755,6 +2755,19 @@ <h3>
...
+            </p>
+            <pre class="example">
+curl -s https://www.w3.org/ns/did/v1  | openssl sha256
```
davidlehn, line 2762:
```suggestion
curl -s https://www.w3.org/ns/did/v1 | openssl sha256
```

msporny (approved):

msporny:

Issue #417, opened 2020-09-25
How should resolvers handle the accept header?


OR13:
https://github.com/decentralized-identity/universal-resolver/issues/150

^ the universal resolver is one example of a resolver, I would like to establish working group consensus for what should happen in the following cases:

1. `application/did+json` is requested but did method only supports `application/did+ld+json`.

2. `application/did+ld+json` is requested but did method only supports `application/did+json`.

3. `*/*` is requested but did method only supports `application/did+ld+json`.

4. `*/*` is requested and did method supports `application/did+ld+json` and `application/did+json`.

5. `application/did+ld+json, application/did+json` is requested and did method supports `application/did+ld+json`.

6. `application/did+json, application/did+ld+json` is requested and did method supports `application/did+ld+json`.

iherman:
Shouldn't we have a closure on the media type issue first (#208)? `application/did+ld+json` may look very different in fact...

OR13:
@iherman I completely agree, but since I'm reviewing PRs in the universal resolver that are assuming `application/did+ld+json`... I feel we need to start having this conversation now.

In many ways, this might be a better way to approach the questions about the ADM / Producer / Consumer issues for representations, its possible if I understood how we plan to handle this issue, I might not object to as much of the ambiguity in the ADM.

@jricher can you add your thoughts on this? I know you worked on the did resolution spec, I think most of my concerns with representations are actually just variations of not understanding how you, @talltree and @peacekeeper see this issue being addressed.

peacekeeper:
I'd like to make one subtle meta-comment about this topic.. DID resolvers do not have a dependency on HTTP. Even though the Universal Resolver and some other implementations expose the DID resolution function at an HTTP endpoint, this isn't a requirement for resolvers.

The DID Resolution spec defines an [HTTP(S) Binding](https://w3c-ccg.github.io/did-resolution/#bindings-https), which maps the inputs and outputs of the abstract `resolveStream()` function to an HTTP request/response. This binding and other implementation details are out of scope for DID Core.

For this reason, we shouldn't be talking about an "Accept header" here in DID Core, but about the "Accept input metadata property".

Having said that, all your questions about the different cases are still valid and very interesting, and I agree this could be a helpful approach to the representation/producer/consumer topics!

peacekeeper:
[...]

PR #487, opened 2020-12-12
Define representation-not-supported error

OR13:
Addresses https://github.com/w3c/did-core/issues/417

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/487.html" title="Last updated on Dec 12, 2020, 6:42 PM UTC (365cecc)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/487/16ba782...365cecc.html" title="Last updated on Dec 12, 2020, 6:42 PM UTC (365cecc)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -3562,6 +3562,13 @@ not-found
 The <a>DID resolver</a> was unable to find the <a>DID document</a>
 resulting from this resolution request.
                         </dd>
+<dt>
+representation-not-supported
+</dt>
+<dd>
+This error code is returned if the representation requested via the <code>accept</code> input metadata 
+property is not supported by the DID method and/or DID resolver implementation.
+</dd>
                         <dt>
 deactivated
                         </dt>


index.html
```diff
@@ -3562,6 +3562,13 @@ <h3>
 The <a>DID resolver</a> was unable to find the <a>DID document</a>
 resulting from this resolution request.
                         </dd>
+<dt>
+representation-not-supported
```
msporny, line 3566:
```suggestion
unsupported-representation
```

Don't know if this is better than the one in the original PR... thoughts?

msporny (approved):
Minor potential improvement, otherwise LGTM.

kdenhartog (approved):

index.html
OR13: it's shorter, I don't have strong feeling one way or the other.

Seems to align better with https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/415

I think technically https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/406

is more accurate, perhaps "representation-not-acceptable" is better?

peacekeeper (approved):
Looks good.

index.html
msporny: hmm... perhaps we want to align w/ HTTP response codes when possible... this would be one such case, and agree that "representation-not-acceptable" is probably the more technically accurate error.

Thoughts from others? Remember that these codes, once defined, will stick around forever, so bikeshedding at this point is appropriate.

msporny: No additional thoughts from the group, merging as-is.

msporny:
Normative, multiple reviews, changes suggested (but not made), no objections, merging.

Issue #359, opened 2020-07-24
Change serviceEndpoint from string-or-object to array-or-object, or just an object with a URI array property


csuwildcat:
The `serviceEndpoint` property is already a polymorphic property, allowing a value that may either be a string URI or an object with custom properties. It appears that in many use cases, the most desirable value type is an array of URI strings. This is the case with just about any service type where there are multiple endpoints, instances, or locations involved. I would propose one of the following resolutions be adopted:

1. The `serviceEndpoint` property can be a URI string, an array of URI strings, or an object with custom properties.
2. The `serviceEndpoint` property can be an array of URI strings or an object with custom properties (instead of adding another type, just make the default an array of URIs, wherein people can just add one if they don't have multiples)

Please let me know what you think, because right now we probably going to see two of the most common service types (DID-to-domain linkage and SDS/Hubs) use a custom object for this property just so they can specify multiple endpoints.

agropper:
Everything in the DID Document is public and will be crawled and indexed by thousands, maybe millions of “directories” and other data brokers. Even for “private” DIDs this kind of thing poses a large correlation risk. 

On the other hand, I don’t understand why this would be useful to anyone. Are we just trying to reduce the cost of running a search engine to a minimum? If not, then adding a mediator or policy decision point as a best practice for service endpoints seems reasonable.

csuwildcat:
> On the other hand, I don’t understand why this would be useful to anyone. Are we just trying to reduce the cost of running a search engine to a minimum? If not, then adding a mediator or policy decision point as a best practice for service endpoints seems reasonable.

It's not about reducing costs, it's about providing a value signature for Service Endpoints (whatever their function) that serves the most frequent cases. One case we're thinking about is a service endpoint that describes a list of Web origins that are associated with the DID, for example: a company may have a DID and own three different websites. If they wanted to use the DID Configuration mechanism of linking to the proofs of control across those sites, they would want to list the three Web origins of those sites in the DID Configuration Service Endpoint. Today, we would have to add an object to that value and some subproperty that takes an array, all because we can't just add three strings in an array directly. Same goes for personal datastores: if a DID owner wants to link to 1+N instances of their personal datastores, we will again need to define a custom service endpoint object with some random subproperty that has an array value, all because we can't add an array with URI strings directly.

csuwildcat:
[...]

PR #489, opened 2020-12-12
Add LinkedDomain example

OR13:
Related https://github.com/w3c/did-spec-registries/pull/168

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/489.html" title="Last updated on Dec 12, 2020, 7:08 PM UTC (8388af7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/489/16ba782...8388af7.html" title="Last updated on Dec 12, 2020, 7:08 PM UTC (8388af7)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2236,6 +2236,10 @@ standard specification.
       <pre class="example nohighlight" title="Various service endpoints">
 {
   "service": [{
+    "id":"did:example:123#linked-domain",
+    "type": "LinkedDomains",
+    "serviceEndpoint": "https://bar.example.com"
+  }, {
     "id": "did:example:123456789abcdefghi#openid",
     "type": "OpenIdConnectVersion1.0Service",
     "serviceEndpoint": "https://openid.example.com/"


index.html
```diff
@@ -2236,6 +2236,10 @@ <h2>Service Endpoints</h2>
       <pre class="example nohighlight" title="Various service endpoints">
 {
   "service": [{
+    "id":"did:example:123#linked-domain",
+    "type": "LinkedDomains",
+    "serviceEndpoint": "https://bar.example.com"
```
msporny, line 2241: Interested in hearing what @jandrieu has to say about this.

`service` feels like a footgun feature... but one can't deny that it also has a variety of useful features. I'm still not clear on group consensus wrt. when it's fine to put in things that would cause trouble from a GDPR/PII perspective and when it's fine.

For example, if people take issue with `"type": "Person"`, then they should probably also have a huge problem with the service endpoint above?

index.html
```diff
@@ -2236,6 +2236,10 @@ <h2>Service Endpoints</h2>
       <pre class="example nohighlight" title="Various service endpoints">
 {
   "service": [{
+    "id":"did:example:123#linked-domain",
+    "type": "LinkedDomains",
```
msporny, line 2240: Is this a real use case? Are we going to define "LinkedDomain"? If so, I'm good... if not, let's use a real example.

msporny (changes requested):
I'd like to see more WG input on this one... feels like we're back on the `type` slippery slope.

tplooker:
@OR13 interesting proposal, if the idea with this example is to express the linkage of the DID to another identifier (in this case a domain name) would we not be better placed to just use the `alsoKnownAs` property?

awoie:
@Manu Sporny <msporny@digitalbazaar.com> the LinkedDomains example is real
and was taken from the DIF Well-Known DID Configuration (aka well-known)
specification:
https://identity.foundation/.well-known/resources/did-configuration/#linked-domain-service-endpoint

OR13:
Unlike the other listed `service.type` examples "LinkedDomain" has an actual spec behind it. 

Perhaps we should remove all the other examples ?

OR13:
> @OR13 interesting proposal, if the idea with this example is to express the linkage of the DID to another identifier (in this case a domain name) would we not be better placed to just use the `alsoKnownAs` property?

The idea is to add an example of support for the spec you are editor of :)

`alsoKnownAs` has no cryptographic binding associated with it.

msporny:
> Unlike the other listed `service.type` examples "LinkedDomain" has an actual spec behind it.
> 
> Perhaps we should remove all the other examples ?

Yes, we should remove examples that don't have specs behind them... or create specs for the examples. There is a small caveat to that, which are examples that demonstrate how one might create extensions (where we specifically want to use `xyzExample` style names). 

I read the LinkedDomain spec... the `origin` thing feels wrong, why wasn't an array used instead? Is that set in stone?

OR13:
> > Unlike the other listed `service.type` examples "LinkedDomain" has an actual spec behind it.
> > Perhaps we should remove all the other examples ?
> 
> Yes, we should remove examples that don't have specs behind them... or create specs for the examples. There is a small caveat to that, which are examples that demonstrate how one might create extensions (where we specifically want to use `xyzExample` style names).
> 
> I read the LinkedDomain spec... the `origin` thing feels wrong, why wasn't an array used instead? Is that set in stone?

@msporny I think Microsoft and a number of other companies, including Transmute, Mattr, UPort, Spherity have some support for that spec as it is written. There are some open issues on the repo.

msporny:
> @msporny I think Microsoft and a number of other companies, including Transmute, Mattr, UPort, Spherity have some support for that spec as it is written. There are some open issues on the repo.

What was the reasoning behind `origins`?

Namely, why did the group decide to do this:

```
{
  "id":"did:example:123#foo",
  "type": "LinkedDomains",
  "serviceEndpoint": {
    "origins": ["https://foo.example.com", "https://identity.foundation"]
  }
}
```

instead of this?

```
{
  "id":"did:example:123#foo",
  "type": "LinkedDomains",
  "serviceEndpoint": ["https://foo.example.com", "https://identity.foundation"]
}
```

OR13:
@msporny we debated the typing of serviceEndpoints as "string" || "array" || "object"... which still makes me cringe typing it out... and ultimately @csuwildcat did what he wanted here... I will note that `origins` is a new term, and is thus not valid JSON-LD in the default context without extension, hence my example is singular https://github.com/w3c/did-core/pull/489/files#diff-0eb547304658805aad788d320f10bf1f292797b5e6d745a3bf617584da017051R2241

and the did spec registries attempts to add a single known "service.type" to the did core context, so that no extension is required for the singular terms: https://github.com/w3c/did-spec-registries/pull/168/files#diff-2e25992eb310ac6568a9e9185473769fee1f26806eb08947239c1924de64b571R20

csuwildcat:
@OR13  "we debated the typing of serviceEndpoints as "string" || "array" || "object"... which still makes me cringe typing it out... and ultimately @csuwildcat did what he wanted here..." - this is factually incorrect, so I must clarify the record:

People using the spec wanted to add multiple origins to their service endpoint, and the DID Core spec says it should be either one string URI or an object with custom properties, so for folks who wanted to include multiple origins we use the object to option the spec provides. This was just a plain use of the DID Core spec as it is written, not something I decided or wanted. I actually advocated for something different than what we ended up with, I wanted DID Core to change to use an object with a `uri` property with an array as its value, but got shot down by DID Core.

csuwildcat:
@msporny we had this discussion here, and one option I highlighted, ironically, was to make the `serviceEndpoint` prop an array or always an object with a `uri` array property, as you alluded to above: https://github.com/w3c/did-core/issues/359

I was not successful in convincing the folks on that thread to move to either one, thus we had to deal with what we were given by the DID Core spec.

OR13:
@csuwildcat thanks for clarifying.... IMO this highlights why types matter... ever JSON member should have a clearly defined set of allowed types... If DID Core says `serviceEndpoint` is a string or an object.... thats the same thing as saying it's "not an array"... thats why the group did what we did.

IMO 

```
{
  "id":"did:example:123#foo",
  "type": "LinkedDomains",
  "serviceEndpoint": ["https://foo.example.com", "https://identity.foundation"]
}
```

Is the correct way of solving this issue, but would require changes to DID Core, see https://w3c.github.io/did-core/#dfn-service

> The value of the serviceEndpoint property MUST be a valid URI conforming to [RFC3986] and normalized according to the rules in section 6 of [RFC3986] and to any normalization rules in its applicable URI scheme specification, OR a set of properties which describe the service endpoint further.

msporny:
```
  "type": "LinkedDomains",
  "serviceEndpoint": ["https://foo.example.com", "https://identity.foundation"]
```

What we could do is fix DID Core, so it allows a URL (string) or an array of URLs (strings) -- we could forbid the use of object (unless someone has a use case for that, which I doubt). Then the spec that specifies "LinkedDomains" could further enforce that serviceEndpoint MUST always be an array of URLs (strings). I think the most important thing here is we try to get rid of "origins" to make it slightly easier on developers and establish a real usage pattern for `serviceEndpoint`.

Thoughts?

csuwildcat:
Ah, but this needs to be an origin in form that matches the same origin
model of the Web, not just some URL. They are different, and for security
purposes, must be separated from just and URL:
https://html.spec.whatwg.org/multipage/origin.html#origin

dlongley:
@csuwildcat,

> Ah, but this needs to be an origin as defined by the the limitations that match the same origin
model of the Web, not just some URL.
> If this change was made, could a service endpoint definition place more structures on the strings? For example: foo.domain.com/alice is === to foo.domain.com/bob in secure origin terms, and the path is irrelevant and discarded?

Both DID methods and specific service types could always impose additional restrictions on acceptable values and processing rules for users of the expressed data. The did core spec would just say that non-URL values are invalid. Since origins are valid URLs, I don't think there's an issue there.

msporny:
I'd like to merge this, waiting on at least one other review -- looks like a fair number of people have looked at this... but no one has approved yet. Need some concrete positions from the group.

msporny (approved):

kdenhartog:
> What we could do is fix DID Core, so it allows a URL (string) or an array of URLs (strings) -- we could forbid the use of object (unless someone has a use case for that, which I doubt)

The only one I could come up with is when did:example:123#serviceEndpoint is included (since it's a URL) and de-referenced by a documentLoader to the full object. Does that present us with any concerns?

dlongley:
@kdenhartog,

> The only one I could come up with is when did:example:123#serviceEndpoint is included (since it's a URL) and de-referenced by a documentLoader to the full object. Does that present us with any concerns?

Are you talking about a self-referential case like this?

```js
{
  "service": [{
    "id":"did:example:123#serviceEndpoint",
    "type": "SomeService",
    "serviceEndpoint": "did:example:123#serviceEndpoint"
  }]
}
```

If so, I don't think its a concern at this level. If you dereference `did:example:123#serviceEndpoint` and use the result to get the value of `serviceEndpoint` **and then** dereference that value, that's a separate step. A step we don't talk about here in core, I don't think. It is clear that a DID resolution spec would want to consider cycles like this and prevent them, but that's not an issue here for the data model.

msporny (approved):

msporny:

Issue #423, opened 2020-10-01
Add NOTE about Verification Method types and remove the redundant languages on key representations from DID Core


OR13:
https://github.com/w3c/did-core/pull/406#discussion_r498455863

I think we are wasting our time defining the same thing with slight differences 4 or 5 times in difference places.

We have the ADM, it defines supported key types (actually it doesn't but I am assuming it will, possibly by using the `type` property value.).

We have the did spec registries, it defines supported key types for all supported representations, in some cases it also links to the definitions elsewhere.

We have the linked data signature suites, like `Ed25519Signature2018` and `JsonWebSignature2020` which define the verification method types, like : `Ed25519VerificationKey2018` and `JsonWebKey2020`.

Regardless of opinions on JWK vs not JWK....

We really ought not to need to define "all supported key types" in did core and did spec registries.... We should be able to say something like:

"Verification Methods are responsible for defining what key representations and proof formats are supported, for example:
P-256 JWK supports ES256, and `Ed25519VerificationKey2018` supports `Ed25519Signature2018`."

"DID Core defines support for a few verification methods, and additional methods may be found in the did spec registries."

Then list the Verification Method types that DID Core Supports ( without any extensions )...

Then delete the entire section about verification method types, but leave the details about required properties such as `id`, `type`, `controller`, `publicKeyBase58` and `publicKeyJwk`...

OR13:
We also want to be able to explain support for `secp256r1` aka P-256 and the other curves listed here:

https://www.iana.org/assignments/jose/jose.xhtml#web-key-elliptic-curve

I personally don't think its wise for us to repeat a version of this table in DID Core, and instead I advocate for relying on the Linked Data Suite to define the supported crypto and note that JsonWebSignature2020 was created to resolve this exact issue... If you don't want agility, don't use it :)

msporny:
Agree on the proposal, this issue will need to result in a PR in the next 30 days or the spec text will remain as it is right now.

msporny:
This issue is related to issue #240 and will be closed when there is a PR for that issue (or this issue).

msporny:
Issue #240 was closed via PR #490, which removed redundant language regarding key representations from DID Core and has been merged. Closing.

Issue #240, opened 2020-03-24
Should did-core restrict the use of JWK?


OR13:
There was some debate on the did core registries about whether JWK is safe to allow in did documents, without this spec making some restrictions: 

- https://github.com/w3c/did-core-registries/pull/20

This thread has some similar arguments:

https://github.com/transmute-industries/lds-jws2020/issues/4

Probably a special call is needed to address this....

OR13:
Noting that existing verification methods, are currently not defined in did core, at all.... https://github.com/w3c/did-core-registries/pull/21

So this will indeed be difficult to do :)

OR13:
I'd argue this ticket is not possible to address, and is outside of the scope of this WG... DID Method Implementers may choose to use `jwks` and NONE of the existing verificationMethods defined in the current spec, and rely on all the the machinery that is defined in JOSE and OIDC / OAuth2.0... the only thing preventing interop via `jwks` would be its inclusion as a verification method in the did core registries, and even if it never gets added there... DID methods could still implement support for it and have interop.

Because how a DID Method treats publicKeyJwk is defined in the DID Method software, we cannot mandate that the software ignore / or enforce properties defined in the DID Core data model... 

There is no way to reflect the RFC compliant data model for a JWK in JSON-LD other than using `@json`, although it is possible to do this more accurately in JSON Schema.

From the charter:

https://www.w3.org/2019/09/did-wg-charter.html

`Defining specific authentication, signing, or cryptography mechanisms. Scope is limited to defining extension points for these mechanisms. (See "Scope".)`

I consider altering the expected behavior of JWKs to be defining a specific cryptographic mechanism for expressing keys, and their relationship to signing / encryption (limiting JWK functionality defines through exclusion a novel cryptographic representation).

From the in scope section:

`Define extension points enabling authentication, signing and cryptography mechanisms, but not defining specific authentication, signing, or cryptography mechanisms. (See "Out of Scope".)`

extending JWK to work with DIDs seems fine to me (by adding them to a did document).... disabling expected functionality defined by IETF for JWKs, is not an extension, it is a restriction, one which alters what can be expressed as a JWK... this means we would be defining a new cryptographic standard for "DID Document JWKs", which are like "JWKs" but limited by the DID Core working group.

[...]

PR #490, opened 2020-12-12
Refactor verification methods section

OR13:
Addresses:

https://github.com/w3c/did-core/issues/423
https://github.com/w3c/did-core/issues/240

Describes how verification method types are related to key representations.

Removes "key types".

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/490.html" title="Last updated on Dec 20, 2020, 4:20 PM UTC (c3da0b6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/490/16ba782...c3da0b6.html" title="Last updated on Dec 20, 2020, 4:20 PM UTC (c3da0b6)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1729,7 +1729,7 @@ A suite definition is responsible for specifying the verification method
 <code>type</code> and proof <code>type</code>. 
 For example, see <a href="https://w3c-ccg.github.io/lds-jws2020/">JSON Web Signature 2020</a> and 
 <a href="https://w3c-ccg.github.io/lds-ed25519-2018/">Ed25519 Signature 2018</a>.
-For all registered and supported verification method types available for DIDs
+For all registered and supported verification method types available for DIDs,
 please see [[?DID-SPEC-REGISTRIES]].
 </p>
 
--- a/index.html
+++ b/index.html
@@ -1763,8 +1763,8 @@ please see [[?DID-SPEC-REGISTRIES]].
 <a>verification relationship</a>.
 </p> 
 <p>
-If a verification method does not exist in the <a>DID document</a>, the method was revoked
-or is invalid. 
+If a referenced verification method is not in the <a>DID Document</a> used
+to dereference it, then that verification method is considered invalid or revoked.
 </p>
 <p>
 Each <a>DID method</a> specification is expected to detail how revocation is performed and tracked.
--- a/index.html
+++ b/index.html
@@ -1581,11 +1581,12 @@ href="#did-syntax"></a>.
 
 <p>
 A verification method MUST contain verification material, 
-such as <a>publicKeyJwk</a> or <a>publicKeyBase58</a> 
+such as <a>publicKeyJwk</a> or <a>publicKeyBase58</a>.
 </p>
 
 <p>
-  A <a>verification method</a> MUST NOT contain multiple verification material properties.
+  A <a>verification method</a> MUST NOT contain multiple verification material properties
+  for the same material.
   For example, expressing key material in a <a>verification method</a> using both
   <code>publicKeyJwk</code> and <code>publicKeyBase58</code> at the same time
   is prohibited.
@@ -1593,7 +1594,7 @@ such as <a>publicKeyJwk</a> or <a>publicKeyBase58</a>
         </dd>
         <dt><dfn>publicKeyJwk</dfn></dt>
         <dd>
-         A JSON Web Key that conforms to [[RFC7517]]. MUST NOT contain "d", 
+         A JSON Web Key that conforms to [[RFC7517]]. This value MUST NOT contain "d", 
          or any other members of the private information class as described in 
          <a href="https://tools.ietf.org/html/rfc7517#section-8.1.1">Registration Template</a>.
         </dd>
@@ -1727,7 +1728,9 @@ and supporting formats that have historically had broad deployment.
 A suite definition is responsible for specifying the verification method 
 <code>type</code> and proof <code>type</code>. 
 For example, see <a href="https://w3c-ccg.github.io/lds-jws2020/">JSON Web Signature 2020</a> and 
-<a href="https://w3c-ccg.github.io/lds-ed25519-2018/">Ed25519 Signature 2018</a>. 
+<a href="https://w3c-ccg.github.io/lds-ed25519-2018/">Ed25519 Signature 2018</a>.
+For all registered and supported verification method types available for DIDs
+please see [[?DID-SPEC-REGISTRIES]].
 </p>
 
 <pre class="example nohighlight" title="Various verification method types">
--- a/index.html
+++ b/index.html
@@ -1523,7 +1523,7 @@ a threshold signature. Methods need not be cryptographic.
 
       <p>
 In order to maximize interoperability, support for public keys as verification
-methods is restricted: see <a href="#key-types-and-formats"></a>. For other
+methods is restricted: see <a href="#verification-method-types"></a>. For other
 types of verification method, the verification method SHOULD be registered in
 the [[?DID-SPEC-REGISTRIES]].
       </p>
@@ -1563,7 +1563,7 @@ formats such as JWK [[RFC7517]]. It is RECOMMENDED that JWK <code>kid</code>
 values are set to the public key fingerprint [[RFC7638]]. It is RECOMMENDED
 that verification methods that use JWKs to represent their public keys utilize
 the value of <code>kid</code> as their fragment identifier. See the first key
-in <a href="#example-15-various-public-keys"></a>
+in <a href="#example-15-various-verification-method-types"></a>
 for an example of a public key with a compound key identifier.
           </p>
 
@@ -1578,9 +1578,30 @@ The value of the <code>controller</code> property MUST be a <a
 data-cite="INFRA#string">string</a> that conforms to the rules in Section <a
 href="#did-syntax"></a>.
           </p>
+
+<p>
+A verification method MUST contain verification material, 
+such as <a>publicKeyJwk</a> or <a>publicKeyBase58</a> 
+</p>
+
+<p>
+  A <a>verification method</a> MUST NOT contain multiple verification material properties.
+  For example, expressing key material in a <a>verification method</a> using both
+  <code>publicKeyJwk</code> and <code>publicKeyBase58</code> at the same time
+  is prohibited.
+</p>
+        </dd>
+        <dt><dfn>publicKeyJwk</dfn></dt>
+        <dd>
+         A JSON Web Key that conforms to [[RFC7517]]. MUST NOT contain "d", 
+         or any other members of the private information class as described in 
+         <a href="https://tools.ietf.org/html/rfc7517#section-8.1.1">Registration Template</a>.
+        </dd>
+        <dt><dfn>publicKeyBase58</dfn></dt>
+        <dd>
+         A base58btc encoded public key.
         </dd>
       </dl>
-
 
       <p class="note" title="Verification method controller(s) and DID controller(s)">
 The semantics of the <code>controller</code> property are the same when the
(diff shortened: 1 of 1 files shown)


OR13:
I expect this may be semi controversial, please review the linked issues as we have discussed this many, many, many times...

Essentially, it's a terrible idea to replicate https://www.iana.org/assignments/jose/jose.xhtml in this spec, so instead we should rely on the verification method type to define support for keys / signature formats.

This PR makes this explicit, and removes some leftover language related to "keys" when we really mean "verification method types".

msporny (approved):
Overall, LGTM.

There are minor nits that we can handle once this PR is in. 

Also, *\*thank you\** -- addressing this item was on my todo list for today, which means instead of writing that PR, I can go and hammer a beer by the bonfire in your and Amy's honor. :)

index.html
```diff
@@ -1682,132 +1708,29 @@ <h2>Verification Methods</h2>
...
-expressed and encoded.
-        </p>
-
-<p class="issue" data-number="423">
-The Working Group is still debating how best to communicate support for
-specific verification method types.
+A suite definition is responsible for specifying the verification method 
+<code>type</code> and proof <code>type</code>. 
+For example, see <a href="https://w3c-ccg.github.io/lds-jws2020/">JSON Web Signature 2020</a> and 
+<a href="https://w3c-ccg.github.io/lds-ed25519-2018/">Ed25519 Signature 2018</a>.
```
kdenhartog, line 1730:
```suggestion
<a href="https://w3c-ccg.github.io/lds-ed25519-2018/">Ed25519 Signature 2018</a>. For all registered and supported verification method types available for DIDs please see [[?DID-SPEC-REGISTRIES]].
```

kdenhartog (approved):
Looks good, happy with text as is, just thought it might be useful to point people to the registries for this work as well and as a point of criteria to register a verification method type in the registries it needs a spec.

index.html
```diff
@@ -1578,10 +1578,31 @@ <h2>Verification Methods</h2>
...
+<p>
+  A <a>verification method</a> MUST NOT contain multiple verification material properties.
+  For example, expressing key material in a <a>verification method</a> using both
+  <code>publicKeyJwk</code> and <code>publicKeyBase58</code> at the same time
+  is prohibited.
+</p>
+        </dd>
+        <dt><dfn>publicKeyJwk</dfn></dt>
+        <dd>
+         A JSON Web Key that conforms to [[RFC7517]]. MUST NOT contain "d",
```
rhiaro, line 1596:
```suggestion
         A JSON Web Key that conforms to [[RFC7517]]. This value MUST NOT contain "d", 
```

rhiaro (approved):
Much better!

selfissued (approved):
LGTM

index.html
```diff
@@ -1578,10 +1578,31 @@ <h2>Verification Methods</h2>
 data-cite="INFRA#string">string</a> that conforms to the rules in Section <a
 href="#did-syntax"></a>.
           </p>
+
+<p>
+A verification method MUST contain verification material, 
+such as <a>publicKeyJwk</a> or <a>publicKeyBase58</a>
```
dlongley, line 1584:
```suggestion
such as <a>publicKeyJwk</a> or <a>publicKeyBase58</a>.
```

index.html
```diff
@@ -1578,10 +1578,31 @@ <h2>Verification Methods</h2>
...
 href="#did-syntax"></a>.
           </p>
+
+<p>
+A verification method MUST contain verification material, 
+such as <a>publicKeyJwk</a> or <a>publicKeyBase58</a> 
+</p>
+
+<p>
+  A <a>verification method</a> MUST NOT contain multiple verification material properties.
```
dlongley, line 1588:
```suggestion
  A <a>verification method</a> MUST NOT contain multiple verification material properties
  for the same material.
```

index.html
```diff
@@ -1832,12 +1755,17 @@ <h3>Key types and formats</h3>
...
-<a>verification relationship</a>. Each <a>DID method</a> specification is
-expected to detail how revocation is performed and tracked.
-        </p>
+
+<p>The <a>DID document</a> MUST NOT express revoked keys using a
+<a>verification relationship</a>.
+</p> 
+<p>
+If a verification method does not exist in the <a>DID document</a>, the method was revoked
+or is invalid.
```
dlongley, line 1764: I want to make sure we don't introduce any spec contradictions here that were clearly not intended.

First, this statement could be read to mean that any verification method that isn't referenced in the DID Document is invalid -- thus invalidating everyone else's verification methods and transitively your own (obviously absurd, but this is spec text, so we should be careful).

Second, a DID Document may have a `controller` property that refers to another DID Document in which there are verification methods that could be used to satisfy proofs that would be considered equivalent to those provided by the DID subject. We don't want readers to get confused where one section of the spec says this is ok, and a simple reading here says it's not. Again, that wasn't the intent here.

Lastly, I don't think we want to rule out use cases where other verification methods are expressed via VCs (e.g., for privacy purposes) but are ultimately rooted by a verification method in the DID Document. What we want to ensure is that the base or root is always the DID Document (or a DID Document referenced as a `controller`).

This is my attempt at saying that simply:

```suggestion
Proofs provided by the <a>DID subject</a> must be verifiable by a verification method
that is in the <a>DID Document</a>. This does not preclude the use of protocols that
involve chains of proofs provided that the base proof is verifiable by a verification
method that is in the <a>DID Document</a>. If the verification method for verifying
a base proof provided by the <a>DID subject</a> is not in the <a>DID Document</a>,
the verification method was revoked or is invalid. See <a href="#did-controller"></a> for
more information on verification methods for proofs that are not provided by the
<a>DID subject</a>.
```

dlongley (changes requested):

index.html
dlongley: @OR13, if this suggestion is accepted, I'll approve, this is great. Thanks!

OR13: I don't think I can accept this change, I didn't add anything regarding proofs, that statement regarding absence of a verification method was the result of working group consensus, and is logically simpler than your proposed text.

I don't think we should be describing "proofs" in general in did core, and the sentence as written does not prevent your proposed text.

I think you are opening a can of worms regarding proofs, which is not needed.

whether the vm is for a "base proof (undefined term) or proof chain (undefined term)" is not relevant. if a single verification method `id` is not present in the did document, the verification method (not the proof) is considered revoked.

index.html
```diff
@@ -1682,132 +1709,29 @@ <h2>Verification Methods</h2>
...
-expressed and encoded.
-        </p>
-
-<p class="issue" data-number="423">
-The Working Group is still debating how best to communicate support for
-specific verification method types.
+A suite definition is responsible for specifying the verification method 
+<code>type</code> and proof <code>type</code>. 
+For example, see <a href="https://w3c-ccg.github.io/lds-jws2020/">JSON Web Signature 2020</a> and 
+<a href="https://w3c-ccg.github.io/lds-ed25519-2018/">Ed25519 Signature 2018</a>. For all registered and supported verification method types available for DIDs please see [[?DID-SPEC-REGISTRIES]].
```
TallTed, line 1731: Just line length...

```suggestion
<a href="https://w3c-ccg.github.io/lds-ed25519-2018/">Ed25519 Signature 2018</a>.
For all registered and supported verification method types available for DIDs
please see [[?DID-SPEC-REGISTRIES]].
```

TallTed (approved):
+1 after the suggested changes

index.html
dlongley: @OR13, the "proofs" language I used here mirrors the language used in the controller section. That's why I chose it -- to make it clear that we're not creating a contradiction between two different sections of the spec. That language also has consensus from the group.

dlongley: >  if a verification method id is not present in the did document, the verification method (not the proof) is considered revoked.

This statement is problematic because it presumes that the verification method ID was, at some point, *in* the DID Document and because it makes it sound like a verification method that is in an entirely different DID Document is invalid or revoked simply because it is not present in the DID Document you're looking in. This is why referring to proofs is better -- the whole reason you're looking at a verification method is because you're dealing with a thing (a proof of some sort) that you're trying to verify. If that proof has led you to the current DID Document, and the verification method isn't there, then it cannot be used to verify that proof. That's the concept we want to get across here.

index.html
```diff
@@ -1832,12 +1756,17 @@ <h3>Key types and formats</h3>
...
-<a>verification relationship</a>. Each <a>DID method</a> specification is
-expected to detail how revocation is performed and tracked.
-        </p>
+
+<p>The <a>DID document</a> MUST NOT express revoked keys using a
+<a>verification relationship</a>.
+</p> 
+<p>
+If a verification method does not exist in the <a>DID document</a>, the method was revoked
+or is invalid.
```
dlongley, line 1765: Another option I would be ok with that I think says less but doesn't create the problems I mentioned would be this:

```suggestion
Proofs provided by the <a>DID subject</a> must be verifiable by a verification method
that is in the <a>DID Document</a>. See <a href="#did-controller"></a> for
more information on verification methods for proofs that are not provided by the
<a>DID subject</a>.
```

dlongley: @OR13, please take a look at the above new suggestion. I think I found a way to avoid the contradictions without introducing new language that isn't already in the spec.

index.html
dlongley: Resolving this conversation and referencing a new suggestion here: https://github.com/w3c/did-core/pull/490#discussion_r543618406

brentzundel (approved):

index.html
OR13: I am stuck on the words "provided by"....

I think bringing did subject and did controller in this way is harmful.

I suggest first stating:

1. "If the verification method for verifying a proof is not in the <a>DID Document</a> that results from resolving that verification method, then that verification method is considered invalid or revoked for that proof."

Then adding something to the effect of 

2. "verification methods for a subject can be controlled by the subject or a third party controller. See
<a href="#did-controller"></a> for more information on verification methods
for proofs that are not provided by the <a>DID subject</a>."

1 applies to both subject === controller and subject !== controller.

2 is essentially a warning that controller might not be did subject... which is a case for which we have 0 examples, and which is clearly not well understood (KERI is reinventing this concept IMO)... I am in favor of providing much better clarity on 2, including some actual examples.... I would suggest we raise an issue for it / mark it at risk as well.

OR13: @dlongley I would accept your proposed changes, as is if you can raise an issue for this, noting we are still discussing it.

dlongley: @OR13,

> I am stuck on the words "provided by"...

That is hard fought language to reach consensus from elsewhere in the spec (emphasis mine):

> When a controller property is present in a DID Document, its value expresses one or more DIDs. Any verification methods contained in the DID Documents for those DIDs SHOULD be accepted as authoritative, such that proofs that satisfy those verification methods are to be considered equivalent to proofs **provided by the DID Subject**.

This is why I used it -- to ensure it's consistent. That being said, your suggestion may help us avoid all of this. I don't think we even need to talk about proofs now but I've tweaked your suggestion to make it more correct, IMO. This is about *dereferencing* verification methods there were referenced elsewhere (in proofs, for example) and then not finding them int the DID Document where they should be:

```suggestion
If a referenced verification method is not in the <a>DID Document</a> used
dereference it, then that verification method is considered invalid or revoked.
```

What do you think?

OR13: I am good with that language, but I think you are missing a "to"...

> If a referenced verification method is not in the <a>DID Document</a> used to
dereference it, then that verification method is considered invalid or revoked.

dlongley: @OR13, added the missing "to", thanks!

index.html
```diff
@@ -1728,7 +1728,9 @@ <h3>Verification method types</h3>
 A suite definition is responsible for specifying the verification method 
 <code>type</code> and proof <code>type</code>. 
 For example, see <a href="https://w3c-ccg.github.io/lds-jws2020/">JSON Web Signature 2020</a> and 
-<a href="https://w3c-ccg.github.io/lds-ed25519-2018/">Ed25519 Signature 2018</a>. For all registered and supported verification method types available for DIDs please see [[?DID-SPEC-REGISTRIES]].
+<a href="https://w3c-ccg.github.io/lds-ed25519-2018/">Ed25519 Signature 2018</a>.
+For all registered and supported verification method types available for DIDs
```
TallTed, line 1732:
```suggestion
For all registered and supported verification method types available for DIDs,
```

msporny (approved):

msporny:

PR #508, opened 2020-12-18
Add additional Metadata properties, fix #483

### LINKED ISSUE #483: Issue with using `version-id` and key revocation
I've discovered a quite severe issue that can occur when you use the `version-id` (or the `version-time`) query parameters for DID documents.

## Problem
In the example below we use JWS and assume that kid in the header is set to `did?version-id<id>#fragment`. The same issue is however present if you use other signature systems, and if you use `version-time`.

Consider the following scenario:
```
d1 <- t1 <- d2 <- t3
      /           /
Blockchain---------------> time
          \           \
    c1 <- t2 <- c2 <- t4
```

- `t<n>` - blockchain timestamp
- `d1` - DID document created
- `d2` - DID document updated (key revocation)
- `c1` - JWS signature with content
- `c2` - JWS signature with content

With timestamps from a blockchain we should be able to determine which signatures was made in which order, and thus invalidate any signature that was made by a revoked key. However there is a problem. 

Let's assume that
- `c1` was signed by `did?version-id=t1#key1`
- `c2` was also signed with `did?version-id=t1#key1`

Here `c1` should be a valid signature when we combine it with the information from `t2`. However, `c2` should not be valid when combined with `t4` because the key `#key1` was revoked at `t3`. This is problematic because we don't get any information about when the DID document was updated when resolving it. The only information we get about updates are from the **DID Document Metadata**, but only includes `created` and `updated`. In this case this information is irrelevant. 

## Mitigation

A possible solution would be to introduce some new properties to the **DID Document Metadata**:

- `version-id` - the version id of the document that was resolved
- `version-time` - the time the document version that was resolved
- `next-version-id` - the version-id of the next update of the document
- `next-version-time` - the version-time of the next update to the document.

Adding these properties would allow us to pinpoint when in time a particular `version-id` of a document was valid and thus we would be able to invalidate update `c2` from our example above.

---

Happy to hear other thoughts and possible solutions!

cc @OR13 @awoie @kdenhartog

oed:
Addresses issue #483

***
<a href="https://pr-preview.s3.amazonaws.com/oed/did-core/pull/508.html" title="Last updated on Dec 22, 2020, 3:21 PM UTC (28b10c9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/508/6434f17...oed:28b10c9.html" title="Last updated on Dec 22, 2020, 3:21 PM UTC (28b10c9)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -3769,10 +3769,50 @@ datetime value MUST be normalized to UTC 00:00, as indicated by the trailing
                 </h4>
                 <p>
 DID document metadata SHOULD include an <code>updated</code> property
-to indicate the timestamp of the last <a href="#update">Update operation</a>.
+to indicate the timestamp of the last <a href="#update">Update operation</a>
+for the document version which was resolved. This property MAY not be
+supported by a given <a>DID method</a>. The value of the property MUST
+follow the same formatting rules as the <code>created</code> property.
+                </p>
+              </section>
+
+            <section>
+                <h4>
+                  next-update
+                </h4>
+                <p>
+DID document metadata SHOULD include a <code>next-update</code> property if
+the resolved document version is not the latest version of the document. It
+indicates the timestamp of the next <a href="#update">Update operation</a>.
 This property MAY not be supported by a given <a>DID method</a>.
 The value of the property MUST follow the same formatting rules as the
 <code>created</code> property.
+                </p>
+              </section>
+
+            <section>
+                <h4>
+                  version-id
+                </h4>
+                <p>
+DID document metadata SHOULD include a <code>version-id</code> property
+to indicate the version of the last <a href="#update">Update operation</a>
+for the document version which was resolved. This property MAY not be
+supported by a given <a>DID method</a>. The value of the property MUST be
+an <a data-lt="ascii string">ASCII string</a>.
+                </p>
+              </section>
+
+            <section>
+                <h4>
+                  next-version-id
+                </h4>
+                <p>
+DID document metadata SHOULD include a <code>next-version-id</code> property
+if the resolved document version is not the latest version of the document.
+It indicates the version of the next <a href="#update">Update operation</a>.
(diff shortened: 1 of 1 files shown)


msporny (approved):

brentzundel (approved):

peacekeeper (approved):
We have to be a bit careful not add new features after the WG announced "feature freeze". But in this case I think we can argue that this is an improvement to an existing feature that can fix a real problem. So +1 to adding these, thanks @oed for the contribution!

msporny:
> We have to be a bit careful not add new features after the WG announced "feature freeze". But in this case I think we can argue that this is an improvement to an existing feature that can fix a real problem. So +1 to adding these, thanks @oed for the contribution!

Good call Markus... we should ensure that no one thinks that this is a new feature since we are in feature freeze. I think it could be argued both ways, but since this is a fairly self-contained PR with no big discussion needed, it's fine. 

If this PR turns into a large conversation thread, it'll signal that it's probably more in the "new feature" category. To be clear @oed , at that point it wouldn't be blocked, but would need to be added to the DID Spec Registries as an extension.

kdenhartog (approved):

jonnycrunch:
would this be a great use of the did-spec registries to extend the did-document for a property like `next-version-id` rather than a new feature?

oed:
@jonnycrunch This doesn't extend the *did-document* but rather the *did-document-metadata* that you get when you resolve a DID. 

I think it makes sense to add it here since it seems like #483 could be a quite common issue for people using DIDs in conjunction with blockchain systems. So I see this more like a bug fix rather than a new feature.

msporny:
Normative, multiple reviews, concerns raised about feature creep -- but this is identified as fixing an existing feature in the spec and the thread didn't explode into a feature bashing thread, no objections, merging.

PR #510, opened 2020-12-20
Allow serviceEndpoint to be an array. Address #506.

msporny:
This PR fixes an oversight in the definition of serviceEndpoint to be an array. This should address #506.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/510.html" title="Last updated on Dec 20, 2020, 9:57 PM UTC (668ca83)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/510/6434f17...668ca83.html" title="Last updated on Dec 20, 2020, 9:57 PM UTC (668ca83)">Diff</a>

Issue #506, opened 2020-12-15
Correct the serviceEndpoint property to allow for a string or array of strings


csuwildcat:
The current text in the spec says that the `serviceEndpoint` property in Service Endpoint entries must be either a string or an object. Apparently this is not what it should have been (per @msporny), so it should be corrected to allow an array.

See this earlier Issue for history on this: https://github.com/w3c/did-core/issues/359

msporny:
@csuwildcat I've created PR #510 to address this issue. Please review. 

This issue will be closed once that PR is merged.

msporny:
PR #510 has been merged. Closing.

PR #511, opened 2020-12-20
Add section on "Service Privacy" to Privacy Considerations.

msporny:
This PR adds a section on Service Privacy to the Privacy Considerations section and attempts to address issue #382 by implementing the following WG resolutions: https://github.com/w3c/did-core/issues/382#issuecomment-703119593

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/511.html" title="Last updated on Dec 20, 2020, 11:03 PM UTC (178ace8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/511/6434f17...178ace8.html" title="Last updated on Dec 20, 2020, 11:03 PM UTC (178ace8)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -4805,6 +4805,126 @@ keep negotiated options to a minimum on wire protocols, use encrypted transport
 layers, and pad messages to standard lengths.
       </p>
     </section>
+
+    <section>
+      <h2>
+Service Privacy
+      </h2>
+
+      <p>
+This specification provides a mechanism for expressing services that an
+entity might use to interact with a <a>DID subject</a>
+(see <a href="#service-endpoints"></a>). Implementers are advised that
+while service endpoints in a <a>DID Document</a> increase control and agency
+for a <a>DID subject</a>, that they might also become a privacy risk if misused.
+      </p>
+
+      <p>
+For example, consider the range of privacy implications for the following
+service endpoints (ordered by an entity's expectation of no privacy to
+strong privacy protection):
+      </p>
+
+      <ul>
+        <li>
+A governmental DID Document expressing the government's public website.
+        </li>
+        <li>
+A celebrity DID Document expressing all of their official social media profiles.
+        </li>
+        <li>
+An individual's DID Document expressing their professional profile.
+        </li>
+        <li>
+A corporate DID Document expressing contact information for their paying
+customers.
+        </li>
+        <li>
+An individual's DID Document expressing a service endpoint containing their
+email address (questionable idea).
+        </li>
+        <li>
+An individual's DID Document expressing a service endpoint containing their
+personal phone number and home address (bad idea).
+        </li>
+        <li>
+An individual's DID Document expressing endpoints for managing electronic
+health records (bad idea).
+        </li>
+      </ul>
+
+      <p>
+The examples above demonstrate a spectrum of privacy challenges that
+implementers creating service endpoints want to consider. For these reasons,
+it is important that implementers ask themselves at least the following
+questions when creating new types of service endpoints:
+      </p>
+
+      <ul>
+        <li>
+What is the expected level of privacy that an entity using the endpoint will
+have?
+        </li>
+        <li>
+Is it possible to misuse the service endpoint in a way that exposes more
+data than a <a>DID subject</a> might expect?
+        </li>
+        <li>
+Is it possible to attack the service endpoint in a way that results in an
+unexpected exposure of private or confidential information?
+        </li>
+        <li>
+Is there another way of communicating the information expressed by the
+service endpoint in a way that is safer, more privacy preserving, or less
+likely to be misused or abused?
+        </li>
+        <li>
+What will a privacy impact assessment for the service conclude for the known
+use cases for the service?
+        </li>
+      </ul>
+
+      <p>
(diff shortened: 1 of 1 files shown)


agropper (changes requested):
I strongly disagree with these changes. Why is "An individual's DID Document expressing endpoints for managing electronic health records (bad idea)." a bad idea?

In this comment I make exactly the opposite argument: https://github.com/w3c/did-core/pull/480#issuecomment-748646616 We should be recommending as few service endpoints as possible and protecting them by an authorization server in most cases.

index.html
```diff
@@ -4805,6 +4805,126 @@ <h2>
...
+An individual's DID Document expressing a service endpoint containing their
+email address (questionable idea).
+        </li>
+        <li>
+An individual's DID Document expressing a service endpoint containing their
+personal phone number and home address (bad idea).
+        </li>
+        <li>
+An individual's DID Document expressing endpoints for managing electronic
+health records (bad idea).
```
dlongley, line 4852: I don't think we need these parentheticals. This is a considerations section -- we should let the reader consider whether or not these things are a good or bad idea. We should just elevate attention to them so they will receive that consideration.

agropper:
I think it's much worse than we let on in this section. The examples talk
about institutions, public documents, and individuals and don't consider
DIDs for IoT. We can reasonably expect each of us to have a dozens of
connected things in our homes, cars, and pockets. Many of them will connect
directly into the Ether via 5G. Any unprotected service endpoint in the DID
Document of these devices will be indexed and correlated by data brokers
using AI.

For example:
https://venturebeat.com/2020/12/17/bigids-quick-1-billion-plus-valuation-shows-rise-of-data-intelligence-and-privacy/


I suggest we add a paragraph about DIDs for devices and documents that are
associated with individuals, groups, and companies. DID Documents using
service endpoints for indexing convenience will become a standardized
surveillance protocol across everything we do.

dhh1128 (approved):
I am okay with this section as written. I agree with Dave's comment that the "(bad idea)" parenthetical is not necessary, but it doesn't bother me either way.

I also agree with Adrian's note that we're failing to talk about IOT as it relates to privacy, so an additional paragraph on that topic would be nice. (Maybe @agropper can suggest one?)

agropper:
@dhh1128 and @msporny - I appreciate the work @msporny did to provide the first example of this section as a PR. I've reviewed the resolutions and propose the following to replace the entire Service Privacy section:

The ability for a controller to optionally state at least one service endpoint in the DID Document increases their control and agency. Stating more than one endpoint in the DID Document always adds privacy risk either due to correlation across the endpoint descriptions or because the services are not protected by an authorization mechanism, or both. (Resolution 1)

The degree of added privacy risk for using multiple service endpoints in one DID Document can be difficult to estimate. Privacy harms, are typically due to unintended consequences. DIDs can identify documents, things, and schemas as well as services. As such, they will be associated with individual people, households, clubs, and employers and correlation of their service endpoints will become a powerful surveillance and inference tool. (Resolution 3)

DID Documents are often public and will be stored and indexed efficiently by their very nature as standards-based. This risk is worse if DID Documents are published to immutable Verifiable Data Registries. Access to a history of the DID Documents referenced by a DID represents a form of traffic analysis made more efficient through our standards process. (Resolution 4)

Examples of service endpoints in DID Documents come in three broad categories: (i) intentional public disclosures; (ii) DIDs about natural persons (as might be covered by GDPR and similar privacy regulations); and (iii) about things and documents that are associated with natural persons. For category (i), consider non-DID publication mechanisms with only a single service endpoint as the root. In some cases, the publication mechanism might reference a DID Document with no service endpoints at all. For category (ii), prefer using only one service that points to an authorization server or to a mediator / proxy that can provide a kind of herd immunity, or both. For category (iii), avoid the use of multiple service endpoints for a DID because some of these (e.g. an authorization server) are likely to be reused with other, related DIDs. Place correlatable service endpoints behind a “firewall”, if possible, or introduce a mediator / proxy as a sole service endpoint in order to obscure related devices and documents through herd immunity.  (Resolution 2)

Help in turning this into a PR would be appreciated. 

I also hope we can review #506 and https://github.com/w3c/did-core/pull/510 and reference it as privacy anti-pattern in this section.

dhh1128:
I am quite happy with Adrian's writeup and would be glad to turn it into a PR if Manu feels it covers the same semantic territory as his. @msporny ?

msporny:
> I am quite happy with Adrian's writeup and would be glad to turn it into a PR if Manu feels it covers the same semantic territory as his. @msporny ?

Yes, please, and thank you @dhh1128.

I have some minor nits and a few additions, but would be happy in general with @agropper's wording.

I'll leave this PR open until you raise one @dhh1128, then point to that PR as a continuation of this one, and then close this PR.

dhh1128:
I turned Adrian's latest suggestions into PR #515, which is an alternative embodiment of the resolutions made [here](https://github.com/w3c/did-core/issues/382#issuecomment-703119593). If accepted, that new PR would supersede this one.

msporny:
PR #515 supercede's this PR, closing.

Issue #518, opened 2020-12-29
Minor editorial issue: discrepancy between alsoKnownAs and others


iherman:
At present, `alsoKnownAs` is defined as 

> be a list where each item in the list is a URI conforming to [RFC3986].

In all other, similar occurrences of URI-s, like e.g. for controllers, the definition is of the sort:

>  a string or an **ordered set** of strings that conform to the rules in Section § 3.1 DID Syntax.

(Emphasis is mine). Any reason why using lists for one and ordered sets for the others? This is actually the only place where lists are used for the core property definition.

B.t.w., the only other place in the document where lists are used is for `equivalentID`. The same question may apply there: isn't it better to use ordered sets everywhere? One less INFRA notion to be used in the spec...

csuwildcat:
Honestly wasn't aware of the discrepancy until now - is there any practical effect? If not, I don't care personally.

TallTed:
Defining something as an "ordered set" (or an ordered list) implies that there is some importance, some meaning, to that order.

If (as I believe) there is no importance to the ordering of some, any, or all of these, then I suggest that those definitions be changed to say "unordered" (or to have no mention of order).

I have no immediate opinion about "set" vs "list" or "collection" or other group noun/structure, but perhaps we should explore whether there are any implications or stronger meanings associated with these (as has been the case in some other similar situations).

iherman:
@TallTed,

while I agree with your general comments, we have to rely here on the Infra](https://infra.spec.whatwg.org) standard, because that is what we refer to.

As far as I could see, the available data structures in [Infra](https://infra.spec.whatwg.org) are lists, stacks, queues, ordered sets, maps, and structs.  Strangely enough, there are no explicit _unordered_ sets, and the document says, in the [relevant section](https://infra.spec.whatwg.org/#sets):

> Almost all cases on the web platform require an ordered set, instead of an unordered one, since interoperability requires that any developer-exposed enumeration of the set’s contents be consistent between browsers. In those cases where order is not important, we still use ordered sets; implementations can optimize based on the fact that the order is not observable.

(I personally do not agree with the direction taken in Infra, and I think having an explicit unordered set would be a welcome addition to Infra, but that is not a matter of discussion here…)

[...]

Issue #404, opened 2020-09-19
Precise specification of the DID Core Vocabulary


iherman:
(I decided to spin-off this issue from #403, because it becomes a bit intricate and having it as a separate issue may be helpful.)

At the moment the Core spec contains property specifications (`authentication`, `alsoKnownAs`, etc) in English prose. These are scattered all over the place, and that is confusing. There is no one place where all the terms, classes, and constraints are properly defined, are referencable, and possibly easily usable for testing or by implementations. I think it is necessary to provide a clean, crisp, and unambiguous specification as part of DID-Core. This specification should account both the JSON-LD (ie, Linked Data) and pure JSON world, as well as CBOR.

The only reference I found at the moment is [The Security Vocabulary](https://w3c-ccg.github.io/security-vocab/) which contains the DID Core terms (although it is incomplete, `alsoKnownAs` is, for example, missing). This is the vocabulary referred to (via the `https://w3id.org/security#` namespace) from the JSON-LD `@context`, i.e., this is where one get by "following one's nose". (Note that this set of references is only meaningful for JSON-LD.)

But there are problems with the [Security Vocabulary](https://w3c-ccg.github.io/security-vocab/):

- I did not find any RDF(S) version of that vocabulary specification, ie, no formal vocabulary definition. (That may be my fault.)
- The vocabulary seems to be much larger than what is used by DID Core. It is, therefore _not_ a faithful representation of the DID Core vocabulary, and that may become a source of confusion
- It is also not clear what the exact relationship is between the DID Core Registry and this vocabulary.
- The vocabulary is a CG Report. It is therefore not considered as a stable document, i.e., it is _not appropriate as a [normative reference](https://w3c-ccg.github.io/security-vocab/)_ in the DID Core spec (as an aside, it is not referred from the spec at all…)

---

I believe what may want to do:

1. There should be a separate section (appendix or elsewhere) in the DID Core spec, including the definition of _all_ the DID Core terms, with constraints defined in English prose using, when appropriate, the INFRA vocabulary. The terms, the relevant constraints, etc, should be seen as part of the abstract data model_, the common reference point for JSON, JSON-LD, and CBOR.
2. We should have a separate, RDFS ontology file for the DID Core Vocabulary. This should be part of the DID Core spec (exactly how is an editorial question) and should probably be normative.
[...]

PR #519, opened 2020-12-29
Added an appendix as a vocabulary overview

iherman:
This is an attempt to get #404 resolved through some possible consensus.

I have gone through the latest version of the spec and pulled out the vocabulary related statements into one place as an Appendix. This reflects, to the best of my knowledge, the various constraints on the vocabulary. I believe it makes it easier for an author and/or implementers to have a clear idea of the vocabulary, and I did it in English text to be as serialization independent as possible.

I believe making the various machine-readable formats (CDDL, RDFS, SHACL) may still be important and useful, and we should do that in some way or other. But, in contrast to what [I said on the special call](https://www.w3.org/2019/did-wg/Meetings/Minutes/2020-12-15-did#section2), it may not have to be normative after all. 

There are some minor discrepancies that I found when doing all that, and this may not fully reflect the intention. E.g., the text does not say that when a URI is used as a value of a, say, `verificationMethod` then it must also be DID URL, although all examples are doing that. So I added a _SHOULD_ for the usage of a DID URL, and left it as a generic URI in the text (reflecting the current text). Also, the value of the `type` property is left unspecified; this value has a constraint when it is used in JSON-LD, but there is no such automatic constraint for pure JSON or CBOR. We may have to specify that.

See also #518.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/519.html" title="Last updated on Dec 29, 2020, 1:55 PM UTC (ff1dd06)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/519/9605a1e...ff1dd06.html" title="Last updated on Dec 29, 2020, 1:55 PM UTC (ff1dd06)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -5236,7 +5236,7 @@ practice to avoid dislosing unnecessary information in JWE headers.
             <td>MUST be present for any <code>DIDDocument</code>, <code>VerificationMethod</code>, or <code>ServiceEndpoint</code> instance. The value MUST be
               <ul>
                 <li>for a <code>DIDDocument</code>: a <a data-cite="INFRA#string">string</a> that conforms to the rules in Section <a href="#did-syntax"></a>;</li>
-                <li>for a <code>VerificationMethod</code>: a <a data-cite="INFRA#string">string</a> that conforms to the rules in Section <a href="#did-url-syntax"></a>;</li>
+                <li>for a <code>VerificationMethod</code>: a <a data-cite="INFRA#string">string</a> that conforms to the rules of [[RFC3986]] for <a>URI</a>-s and which SHOULD also conform to the rules in Section <a href="#did-url-syntax"></a>;</li>
                 <li>for a <code>ServiceEndpoint</code>: a <a data-cite="INFRA#string">string</a> that conforms to the rules of [[RFC3986]] for <a>URI</a>-s.</li>
               </ul>
             </td>
--- a/index.html
+++ b/index.html
@@ -5186,6 +5186,142 @@ practice to avoid dislosing unnecessary information in JWE headers.
   </section>
 
   <section class="appendix">
+    <h2>Vocabulary Overview</h2>
+
+    <p>This section provides an overview of all core properties defined in this specification in Section <a href="#core-properties"></a>, including the constraints on the usage of the properties.</p>
+
+    <section>
+      <h2>Classes</h2>
+
+      <p>The classes used in this section are not used DID Documents, and are only used to provide a more concise overview of the relevant core properties below. The are all represented, in a DID Document, as a <a data-cite="INFRA#maps">map</a> with specific properties.</p>
+
+      <table class=simple>
+        <thead>
+          <tr>
+            <th>Class</th>
+            <th>Description</th>
+          </tr>
+        </thead>
+        <tbody>
+          <tr>
+            <td><code>DIDDocument</code></td>
+            <td>Class representing the “top level” of the <a>DID Document</a>. Each instance MUST have an <code>id</code> property, and MAY have the <code>alsoKnownAs</code>, <code>controller</code>, <code>verificationMethod</code>, <code>authentication</code>, <code>assertionMethod</code>, <code>keyAgreement</code>, <code>capabilityInvocation</code>, <code>capabilityDelegation</code>, and <code>service</code> properties.</td>
+          </tr>
+          <tr>
+            <td><code>VerificationMethod</code></td>
+            <td>Class representing the generic concept of <a href="#verification-methods"></a>. Each instance MUST have the <code>id</code>, <code>type</code>, and <code>controller</code> properties, as well as <em>one of</em> the <code>publicKeyJwk</code> or <code>publicKeyBase58</code> properties.</td>
+          </tr>
+          <tr>
+            <td><code>ServiceEndpoint</code></td>
+            <td>Class representing the generic concept of <a href="#service-endpoints"></a>. Each instance MUST have the <code>id</code>, <code>type</code>, and <code>serviceEndpoint</code> properties.</td>
+          </tr>
+      </table>
+    </section>
+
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -5185,6 +5185,142 @@ <h2>Encrypting</h2>
...
+            <th>Property</th>
+            <th>Usage constraints</th>
+          </tr>
+        </thead>
+        <tbody>
+          <tr>
+            <td><code>id</code></td>
+            <td>MUST be present for any <code>DIDDocument</code>, <code>VerificationMethod</code>, or <code>ServiceEndpoint</code> instance. The value MUST be
+              <ul>
+                <li>for a <code>DIDDocument</code>: a <a data-cite="INFRA#string">string</a> that conforms to the rules in Section <a href="#did-syntax"></a>;</li>
```
rhiaro, line 5238: I'm concerned that introducing the idea of classes in general but this class in particular might lead to more confusion. I feel like the next logical step is people saying "oh, well in that case `@type: DIDDocument` etc..

rhiaro (changes requested):
I certainly support clarifying the core properties information. This appendix seems like a repetition of what is in the Core Properties section, minus the explanatory prose. I don't think we can use normative language in an appendix? (or if we do it's pointless as appendices are by definition non-normative.) 

Perhaps instead we could use this as a basis to iterate on the structure of Core Properties?

Alternatively I could see this as a helpful summary in, dare I say it, an implementation guide.. (which we could and should link to from the core spec, to make it more obvious to future readers.)

iherman:
> I certainly support clarifying the core properties information. This appendix seems like a repetition of what is in the Core Properties section, minus the explanatory prose. I don't think we can use normative language in an appendix? (or if we do it's pointless as appendices are by definition non-normative.) 

You are right. I got a bit carried away, I guess:-). I am happy to prepare a PR that changes the style of the text to make it possibly less precise (pointing back at the formal definition) but maybe more fluent.

> 
> Perhaps instead we could use this as a basis to iterate on the structure of Core Properties?
> 

That may indeed be another alternative, although if we do it, some sort of collection/index at the end of the document would still be useful. But that may be doable automatically via some scripts.

(The approach chosen in the [Publication manifest](https://www.w3.org/TR/pub-manifest/#value-localizable-string) may be an example on  how it would look.)

I actually thought of going down that direction, but that would require a much more radical change of the current text, so I shied away. 

> Alternatively I could see this as a helpful summary in, dare I say it, an implementation guide.. (which we could and should link to from the core spec, to make it more obvious to future readers.)

Well... Until we know what this document will contain (if ever it sees the light of the day) this is difficult to say. We can of course keep the table as is (or in a simplified form that I referred to) and move it out of the core document into the guide if it comes. It is not a normative section, meaning that it can be removed from the core spec at any time, even during the CR phase.

index.html
iherman: Yeah, and that is also why I added an extra note in the text. But, unfortunately, if we do not use something like that then the language would become extremely convoluted (how would I describe the constraints on the `id` property, for example?) Any idea?

msporny (changes requested):
I like that there is one location for this information -- that's very useful. I also like the concept of saying that a "DIDDocument" and "VerificationMethod" is a class of information -- although, that should probably go in each section on the topic... perhaps at the end of the Data Model section?

I'm concerned that the PR repeats a subset of normative statements in the specification -- we definitely shouldn't be doing that. Instead of treating this like an appendix, perhaps we could put it in the "Introduction" section somewhere in the spec -- perhaps in the data model section? Summarize there, and then go into detail in the rest of the specification? 

Maybe we can do introductory normative text in the data model section and then say "additional requirements for this property can be found in Section X.Y"? That way we can summarize in a normative fashion and put additional constraints in each section? It would be a fairly non-trivial rework of the spec, but might work? Up to you if you want to go to that effort @iherman, or if you'd rather say things non-normatively in data model and then point to each section where the property is defined?

rhiaro:
> Maybe we can do introductory normative text in the data model section

I think this makes sense, and I'd be happy to attempt the re-work using Ivan's text from this PR, if you don't have time/energy @iherman? I'm not sure if you meant Data Model section or Core Properties section though @msporny. Slight preference for the latter (replacing the list summary that is there now), but fine with either.

msporny:
> I'm not sure if you meant Data Model section or Core Properties section though @msporny. Slight preference for the latter (replacing the list summary that is there now), but fine with either.

I meant the former, but I think you're correct, and it should be the latter. The only issue with "Core Properties" is that the classes (did document, verification method, service endpoint, etc.) aren't necessarily properties... but that's an editorial detail that can probably be worked through. You've convinced me that we shouldn't put this in the Data Model section.

rhiaro:
To minimise confusion between the 'classes' being helpful narrative concepts vs vocabulary terms / RDF-type classes, I'd like to see them in normal text, ie. "DID Document" rather than `DIDDocument` etc. - does that sound okay? And given that they aren't actually a serializable part of the data model, I don't see any contradiction with mentioning them in the Core Properties section.

iherman:
> I think this makes sense, and I'd be happy to attempt the re-work using Ivan's text from this PR, if you don't have time/energy @iherman?

@msporny @rhiaro thanks for the comments. I will try to find time today to make an update on the PR, but thank you for the offer. I may take it up later:-)

iherman:
@msporny @rhiaro I decided to create a separate PR (#528) instead of modifying this one. There has been a series of PR merges since I did this PR and I was worried about a bunch of merge conflicts if I continue this commit.

If we all agree that #528 represents a more acceptable direction, then we should move the discussion there and close this one without any further action.

msporny:
> If we all agree that #528 represents a more acceptable direction, then we should move the discussion there and close this one without any further action.

No objections to #528, closing this one in favor of #528.

Issue #363, opened 2020-07-29
Clarify Authorization requirements.


Fak3:
Authorization of operations performed on behalf of the DID subject is defined in several places in the spec and requirements for implementers are unclear.

Section 5.3.2 defines Authorization as "the mechanism used to state how operations are performed (by controller) on behalf of the DID subject. " - it is unclear which operations it is talking about? Does it include DID Document Update operation performed by DID controller? Or does it include assertion of a statement on behalf of the DID subject? (discussed in 5.4.2 assertionMethod). These two exemplified operations have completely different requirements and such vague definition increases confusion about the whole section.

Then it also say that "Each DID method MUST define how authorization and delegation are implemented, including any necessary cryptographic operations." - why core spec delegates the responsibility to the DID method spec, and at the same time define authorization mechanisms as 5.4.2 assertionMethod? What a DID method spec should say about implementation of assertionMethod and other operations (verification relationships) defined in the core spec?

At the end it also suggests methods that a verifiable data registry could implement to support Authorization and delegation. This increases the confusion further - what role the verifiable data registry has in interpreting behavior of 5.4.2 assertionMethod ? Could assertionMethod be delegated too?

It seems that "Authorization" as defined in this section is very vague and confusing. The whole section 5.3.2 should be somehow reworded and probably reorganized to address the relationship between section 5.4 _Verification Relationships_, the DID method-specific notion of authorizing of DID controller, interpreting _controller_ property of DID Document, requirements for verifiable data registry, and delegation.

Fak3:
Rlated: #318 #269 #199 #122

msporny:
Agreed, we need to rewrite how Authorization is described in the spec and the Editors will need to do that in a rewrite.

msporny:
Still waiting on a PR.

rhiaro:
The Authorization section has now gone, and the whole Core Properties section has been reorganised and tidied up since this issue would raised. @Fak3 could you review [Verification Methods](https://w3c.github.io/did-core/#verification-methods) and [Verification Relationships](https://w3c.github.io/did-core/#verification-relationships) to see if the latest versions goes some way towards answering your questions?

OR13:
recommend closing, since the spec has drifted

msporny:
PR #525 has been raised to address this issue. This issue will be closed once PR #525 is merged. @Fak3, please let us know if this addresses the concerns you raised above.

Fak3:
Thank you. PR looks good to me

[...]

PR #525, opened 2021-01-03
Address issue #363 by clarifying use of word "authorization".

msporny:
Attempt to address issue #363 by clarifying the use of the word "authorization" in the specification by being really pedantic about its usage.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/525.html" title="Last updated on Jan 3, 2021, 6:45 PM UTC (15ffd2b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/525/c2841ed...15ffd2b.html" title="Last updated on Jan 3, 2021, 6:45 PM UTC (15ffd2b)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1492,9 +1492,9 @@ DID Controller
       </h2>
 
       <p>
-Authorization is the mechanism used to state how operations are performed
-on behalf of the <a>DID subject</a>. A <a>DID controller</a> is authorized to
-make changes to the respective <a>DID document</a>.
+A <a>DID controller</a> is an entity that is authorized to make changes to
+a <a>DID document</a>. The process of authorizing a <a>DID controller</a> is
+defined by the <a>DID Method</a>.
       </p>
 
       <p>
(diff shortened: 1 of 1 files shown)


rhiaro (approved):

agropper:
Apologies for being late to comment on this issue. I don't object to the proposed changes but wonder if we could find a different word for authorization. Maybe something like 'allowed to change' the DID Document. Authorization is a bit confusing because in OAuth and with respect to some DID service endpoints authorization refers to read as well as write or change access. DID Documents are best understood as public (even though they don't have to be) and using a phrase like 'allowed to change' or 'authorized to change' would be clearer than 'authorization"

msporny:
@agropper wrote:
> Maybe something like 'allowed to change' the DID Document. 

Hmm, the PR does that... see text from the spec below:

> authorized to make changes to ￼a DID document

The change the PR implements is being very clear about the specific type of authorization (when there is something like that that can be identified).

-1 to using another word, as we do mean "authorization" in the instances where we use it in the specification (and are clear about the type of authorization). The group has also gone to some lengths to define the difference between authentication and authorization.

If you have concrete spec text suggestions @agropper, that might help us discuss the changes you'd like to see.

Fak3 (approved):

dlongley (approved):

brentzundel (approved):

OR13 (approved):

msporny:
Editorial, multiple reviews, changes requested (but not made), no objections, merging.

Issue #384, opened 2020-08-31
Normative statements review


rhiaro:
I used a respec extension to pull out all the normative statements from the spec. They are listed here, with a note about their testability (if empty, assume it's automatically testable afaik), and proposals for changes to some.

Changes are to remove redundancy (it's not good to have more than one normative statement for the same feature, especially in different sections of the spec, and especially if the wording is different such that it might introduce ambiguity); to move statements to more appropriate sections; or to remove normative language where it's not needed (eg. doesn't make sense in terms of something for concrete implementations to follow).

Changes are *not* to alter the intent of the wording spec in a normative way that would actually require implementation changes. If there is a normative statement that you disagree with the _content or existence_ (rather than clarity, location or structure) of, please raise that in a new issue (and check for existing issues) rather than commenting here.

I will start a series of PRs for the proposed changes and keep track of them here too.

The main purpose of this is to help with the test suite development.

Something being "not automatically testable" is fine - just means the implementer asserts that it is true. "Not testable" at all, is a problem.

Please comment if there are any further changes you would like to see to normative statements, or if you disagree with any of my proposed changes, or if you notice I've missed one.

| Section | Statement | Testability | Proposed change | PR |
| ---------- | -------------- | ------------- | ------------------------ | -- |
| 1.4 Conformance | A conforming DID is any concrete expression of the rules specified in Section § 3. Identifier and MUST comply with relevant normative statements in that section. | Not testable by itself; depends on certain other tests passing | Remove normative to be purely descriptive? | https://github.com/w3c/did-core/pull/426 |
| 1.4 Conformance | A conforming DID document is any concrete expression of the data model described in this specification and MUST comply with the relevant normative statements in Sections § 4. Data Model and § 5. Core Properties. | Not testable by itself; depends on certain other tests passing | Remove normative to be purely descriptive? | https://github.com/w3c/did-core/pull/426 |
| 1.4 Conformance | A serialization format for the conforming document MUST be deterministic, bi-directional, and lossless as described in Section § 6. Core Representations. |  | Remove as duplicate of same in section 6 | https://github.com/w3c/did-core/pull/426 |
| 1.4 Conformance | A conforming producer MUST NOT produce non-conforming DIDs or DID documents. |  | | |
| 1.4 Conformance | A conforming consumer MUST produce errors when consuming non-conforming DIDs or DID documents. |  | | |
| 3.1 DID Syntax | The DID scheme name MUST be an ASCII lowercase string. |  | | |
[...]

PR #429, opened 2020-10-03
normative statements: Core properties section

rhiaro:
Normative statement tweaks in the core properties section based on feedback from @kdenhartog and @msporny; see #384

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/429.html" title="Last updated on Oct 18, 2020, 3:57 PM UTC (7cf35fc)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/429/c764c67...7cf35fc.html" title="Last updated on Oct 18, 2020, 3:57 PM UTC (7cf35fc)">Diff</a>

PR #526, opened 2021-01-03
Complete top-to-bottom review of testability of normative statements.

msporny:
This PR contains a complete top-to-bottom review of the testability of the normative statements in the DID Core specification and should address issue #384. The goal with this PR is to reduce the workload for the entire WG wrt. the test suite and the tests we're going to have to write in the coming month or two. With that goal in mind - this PR removes or refactors normative statements that are tested elsewhere in the specification, aren't our job to test, or are too vague to be tested (by a human or a machine).

I expect there are a few that reviewers might find concerning, so while you're considering the change, ask yourself the following questions to suggest alternatives:

* Is it the job of the DID Core specification (based on our Charter) to test this particular thing?
* Can this statement be empirically tested by a machine or a human?
* Is this statement not tested anywhere else in the specification?

If the answer to all of those questions is "yes", then there may be an argument to not make a subset of the changes made in this PR.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/526.html" title="Last updated on Jan 3, 2021, 9:20 PM UTC (ce0f0e3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/526/10744a0...ce0f0e3.html" title="Last updated on Jan 3, 2021, 9:20 PM UTC (ce0f0e3)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1290,12 +1290,12 @@ method that ensures that two different representations will be able to work
 together.
         </li>
       <li>
-Representations MAY define other extensibility mechanisms, including ones that do
-not require the use of the DID Specification Registries. Such extension mechanisms SHOULD
-support lossless conversion into any other conformant representation. Extension
-mechanisms for a representation SHOULD define a mapping of all properties and
-representation syntax into the <a href="#data-model">data model</a> and its type
-system.
+Representations MAY define other extensibility mechanisms, including ones that
+do not require the use of the DID Specification Registries. Such extension
+mechanisms SHOULD support lossless conversion into any other conformant
+representation. Extension mechanisms for a representation SHOULD define a
+mapping of all properties and representation syntax into the <a
+href="#data-model">data model</a> and its type system.
       </li>
       </ol>
       <p class="note">
@@ -2189,7 +2189,7 @@ A <a>DID document</a> MAY include a <code><a>service</a></code> property.
         <dd>
           <p>
 If a <a>DID document</a> includes a <code>service</code> property, the value
-of the property SHOULD be an <a data-cite="INFRA#ordered-set">ordered set</a>
+of the property MUST be an <a data-cite="INFRA#ordered-set">ordered set</a>
 of <a>service endpoints</a>, where each service endpoint is described by a set
 of properties. Each <a>service endpoint</a> MUST have <code><a>id</a></code>,
 <code>type</code>, and <code><a>serviceEndpoint</a></code> properties, and MAY
--- a/index.html
+++ b/index.html
@@ -4130,7 +4130,7 @@ accept
 The MIME type the caller prefers for <code>content-stream</code>. The <a>DID
 URL Dereferencing</a> implementation SHOULD use this value to determine the
 representation contained in the returned value if such a representation is
-supported and available. This property is OPTIONAL.
+supported and available.
                 </dd>
             </dl>
         </section>
@@ -4151,8 +4151,8 @@ common properties.
 content-type
                 </dt>
                 <dd>
-The MIME type of the returned <code>content-stream</code>.
-This property is OPTIONAL if dereferencing is successful.
+The MIME type of the returned <code>content-stream</code> SHOULD be expressed
+using this property if dereferencing is successful.
                 </dd>
                 <dt>
 error
--- a/index.html
+++ b/index.html
@@ -3264,7 +3264,7 @@ rules are followed:
 
           <ul>
             <li>
-CBOR tags other than the CID tag (42) MUST NOT be utilized.
+CBOR tags other than the CID tag (42) MUST NOT be used.
             </li>
           </ul>
 
--- a/index.html
+++ b/index.html
@@ -943,7 +943,7 @@ If present, the associated value MUST be the
         </table>
 
         <p>
-Implementers as well as <a>DID method</a> specification authors MAY use
+Implementers as well as <a>DID method</a> specification authors might use
 additional DID parameters that are not listed here. For maximum
 interoperability, it is RECOMMENDED that DID parameters use the official W3C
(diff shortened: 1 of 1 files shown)


rhiaro (approved):
I'm not sure about most of the MAY->mights. I read MAYs to be explicit warnings to implementers to look out and account for these behaviours in other implementations they're trying to interop with. I think 2119-ing the MAYs really makes that clear, and doesn't add a testing burden since we don't test MAYs (without tons of free time - but they're all things that _could_ be tested). (But I don't feel strongly about this.)

msporny:
> I'm not sure about most of the MAY->mights. I read MAYs to be explicit warnings to implementers to look out and account for these behaviours in other implementations they're trying to interop with. I think 2119-ing the MAYs really makes that clear, and doesn't add a testing burden since we don't test MAYs (without tons of free time - but they're all things that _could_ be tested). (But I don't feel strongly about this.)

Agree in general. A decent chunk (but not all) of the OPTIONAL/MAYs that I changed were either contradicting a SHOULD earlier in the document, were asserting things about resolving parties, or wouldn't result in testing anything of importance. I tend to not really like MAY/OPTIONAL language as it is rarely ever tested or enforced... it's kinda nothing language -- sure, be aware of this MAY but we're not expecting anyone to do anything about it. In those cases, why have the normative language at all? I know others don't share this viewpoint so happy to put things back in that folks feel strongly about.

I may have also made a few mistakes and missed the nuance of some of the normative MAY/OPTIONAL language.

Happy to reconsider anything folks feel needs to go back to the way it was.

index.html
```diff
@@ -4070,7 +4058,7 @@ <h2>
       </dl>
 
       <p>
-The output variables of this function MUST be as follows:
+The output variables of this function are as follows:
```
selfissued, line 4061: Why would we not test this normative statement?

selfissued (changes requested):
I mostly agree with this PR but it seems to indicate that we don't intend to test some important normative and testable statements.  Let's correct this before merging this.

index.html
```diff
@@ -4045,7 +4033,7 @@ <h2>
       </code></p>
 
       <p>
-The input variables of this function MUST be as follows:
+The input variables of this function are as follows:
```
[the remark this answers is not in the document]
msporny: Eliminating this normative statement has no effect on the number of tests we write because each item in the list asserts a normative test as well and refers to a data structure for which there are already normative tests.

index.html
msporny: Eliminating this normative statement has no effect on the number of tests we write because each item in the list asserts a normative test as well and refers to a data structure for which there are already normative tests.

index.html
```diff
@@ -4035,7 +4023,7 @@ <h2>
 and fragment. This process depends on <a>DID resolution</a> of the <a>DID</a>
 contained in the <a>DID URL</a>.
 The details of how this process is accomplished are outside the scope of this
-specification, but all conformant implementations MUST implement a function
+specification, but all conformant implementations implement a function
```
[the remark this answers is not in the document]
msporny: How would we write a test to ensure that a conformant implementation implements a function with the following *abstract* form. That is, how do we test to make sure an abstract function is defined... since by it's very nature, the function is abstract and has no testable realization?

index.html
```diff
@@ -1847,9 +1847,9 @@ <h2>Verification Relationships</h2>
       </p>
       <p>
 The <a>verification relationship</a> between the <a>DID subject</a> and the
-<a>verification method</a> MUST be explicit in the <a>DID document</a>.
+<a>verification method</a> is explicit in the <a>DID document</a>.
 <a>Verification methods</a> that are not associated with a particular
-<a>verification relationship</a> MUST NOT be used for that <a>verification
+<a>verification relationship</a> cannot be used for that <a>verification
 relationship</a>. For example, a <a>verification method</a> in the value of
 the <code><a>authentication</a></code> property cannot be used to engage in
```
rhiaro, line 1854: Just noting that I tried to remove this normative language in an earlier PR based on a comment by @kdenhartog, and @msporny [objected on the basis that it is testable](https://github.com/w3c/did-core/pull/429#pullrequestreview-501574027). Given this is a conflict between present manu and past manu, I'm going to suggest manu can make the call :)

index.html
```diff
@@ -2189,7 +2189,7 @@ <h2>Service Endpoints</h2>
         <dd>
           <p>
 If a <a>DID document</a> includes a <code>service</code> property, the value
-of the property SHOULD be an <a data-cite="INFRA#ordered-set">ordered set</a>
+of the property MUST be an <a data-cite="INFRA#ordered-set">ordered set</a>
```
rhiaro, line 2192: was this originally a SHOULD because the value could potentially be an ordered set of just URI strings (which dereference to the rest of the data)? Worth checking if this change affects any implementations?

msporny: I'm not sure we ever had that use case... I may be wrong, but as far as I can remember... service was always supposed to be an array of objects (because you have to give the service an `id`, `type`, and a `serviceEndpoint`)... I don't know of any DID Method that doesn't follow that pattern.

index.html
msporny: *\*lol\** ... Dammit, past manu... 'causing trouble.

Past manu was wrong, @rhiaro and @kdenhartog are right.

The normative-ness of the statement is being removed because it is a test intended for a conformance class of software that we don't define in DID Core. In other words, it has very little to do with production and consumption conformance classes and instead has more to do with software that implements authentication/authorization protocols.

Present manu is displeased that past manu failed to realize that. :P

msporny:
Normative, multiple reviews, changes requested (some made), no objections, merging.

PR #427, opened 2020-10-03
normative statements: Identifier section

rhiaro:
Reorganises normative language in Identifier section, towards #384

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/427.html" title="Last updated on Oct 26, 2020, 3:13 PM UTC (47e023a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/427/2a3ca5d...47e023a.html" title="Last updated on Oct 26, 2020, 3:13 PM UTC (47e023a)">Diff</a>

PR #531, opened 2021-01-05
Swap service and relative-ref parameters in table.

peacekeeper:
This is an editorial PR that simply changes the order of the [DID parameter table](https://w3c.github.io/did-core/#did-parameters) to list `service` before `relative-ref`.

I agree with the re-ordering done by @rhiaro in https://github.com/w3c/did-core/pull/427/commits/74586e8f3f1c7245e9d725c25850d749aa028cb6, which nicely lists required parameters before optional ones, but I think the table becomes more readable if `service` comes before `relative-ref`. After all, the latter mentions the former, and all the examples I've seen also use them in this order.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/531.html" title="Last updated on Jan 5, 2021, 9:50 PM UTC (fbe4639)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/531/10744a0...fbe4639.html" title="Last updated on Jan 5, 2021, 9:50 PM UTC (fbe4639)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -868,6 +868,16 @@ Description
           <tbody>
             <tr>
               <td>
+<code><a>service</a></code>
+              </td>
+              <td>
+Identifies a service from the <a>DID document</a> by service ID. Support for
+this parameter is REQUIRED. The associated value MUST be an <a data-lt="ascii
+string">ASCII string</a>.
+              </td>
+            </tr>
+            <tr>
+              <td>
 <code>relative-ref</code>
               </td>
               <td>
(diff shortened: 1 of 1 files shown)


rhiaro (approved):

brentzundel (approved):

msporny (approved):
We should probably alphabetize all tables... as the tables get larger, people will assume some sort of ordering... typically, alphabetical. I don't feel strongly enough to object to this PR (because @peacekeeper makes a good point)... and we can always alphabetize later.

msporny:

PR #533, opened 2021-01-06
Make DID doc metadata propertly names consistent

rhiaro:
Changes equivalent id and canonical id to use hyphens instead of camelCase because the other DID doc metadata properties use hyphens. If I missed an explicit resolution to do these inconsistently, just close the PR..

I don't think this is technically an editorial change, if there are already implementations of these properties with camelCase.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/533.html" title="Last updated on Jan 12, 2021, 7:06 PM UTC (3d9d375)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/533/c8bc6b1...3d9d375.html" title="Last updated on Jan 12, 2021, 7:06 PM UTC (3d9d375)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -4,7 +4,7 @@
   <title>
     Decentralized Identifiers (DIDs) v1.0
   </title>
-  <meta content='text/html; charset=utf-8' http-equiv='Content-Type'><!--
+  <meta content='text/html; charset=utf-8' http-equiv='contentType'><!--
     === NOTA BENE ===
     For the three scripts below, if your spec resides on dev.w3 you can check
     them out in the same tree and use relative links so that they'll work
@@ -892,7 +892,7 @@ string">ASCII string</a>.
             </tr>
             <tr>
               <td>
-<code>version-id</code>
+<code>versionId</code>
               </td>
               <td>
 Identifies a specific version of a <a>DID document</a> to be resolved (the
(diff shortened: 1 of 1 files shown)


peacekeeper:
Good point, I agree this is currently inconsistent. I guess we could discuss whether we want to use hyphens for all metadata properties, or camelCase for all. Perhaps the latter would be preferable and we should get rid of hyphens altogether, given that the DID document Core Properties also use camelCase?

Same question for the DID parameters (e.g. maybe `version-time` should be `versionTime`?)

msporny:
> Perhaps the latter would be preferable and we should get rid of hyphens altogether, given that the DID document Core Properties also use camelCase?

Agree on camelCase as dashes can't be easily used in many programming languages:

```js
const {versionTime} = resolverResponse.documentMeta; // this feels nice
// vs.
const versionTime = resolverResponse.documentMeta['version-time']; // this is bad mojo
```

msporny (changes requested):
We should be going the other way -- from hyphens to camel-case (for the reasons mentioned above).

rhiaro:
@msporny done!

index.html
```diff
@@ -4,7 +4,7 @@
   <title>
     Decentralized Identifiers (DIDs) v1.0
   </title>
-  <meta content='text/html; charset=utf-8' http-equiv='Content-Type'><!--
+  <meta content='text/html; charset=utf-8' http-equiv='contentType'><!--
```
peacekeeper, line 7:
```suggestion
  <meta content='text/html; charset=utf-8' http-equiv='Content-Type'><!--
```
@rhiaro probably you didn't intend to change this line :)

rhiaro: Thanks!!!

peacekeeper (approved):
Looks good, thanks! I would suggest that we also change the DID parameters to camelCase, either in this same PR or another one.

peacekeeper:
> &lt;marquee&gt;

As a member of the TAG, can I request that you bring back &lt;marquee&gt; into HTML and make all browsers support it? It would make the web a so much better place!

msporny (approved):

msporny:

Issue #399, opened 2020-09-17
DID Fragment semantics cleanup needed


iherman:
(This may or may not be an editorial issue, depending on the alternatives below)

The definition of Fragments in [§3.2.4](https://w3c.github.io/did-core/#fragment) looks incomplete to me.

In an HTTP setting the HTTP protocol means first returning the (complete) resource to a client which then, as a second step, takes a portion of that resource using the fragment ID. If my understanding is correct, this is _not_ exactly what happens in the DID URL world: the [§8.2 DID URL Resolver](https://w3c.github.io/did-core/#did-url-dereferencing) returns the _final_ (whatever it is) resource, i.e., it is up to the DID URL Resolver to take care of the fragment identifier processing result. 

1. Is this correct? If so, this clarification should be, in my view, explicitly stated either in §3.2.4. or in §8.2 (or both)  
2. If this is _not_ the case, then we have an additional problem: if one looks at [§8.2.2](https://w3c.github.io/did-core/#did-url-dereferencing-metadata-properties) the media type is returned (so that the client of the DID URL can handle it) as a dereferencing metadata. However, returning that metadata property is _OPTIONAL_, meaning that, potentially, a resource is returned and the client does not know what to do with the fragment ID. Shouldn't it be the case that, when the DID URL includes a fragment, the media type be _REQUIRED_ in the return metadata?

(I hope the answer to the question is (1) above :-)

peacekeeper:
@iherman yes the answer is 1., i.e. the `dereference()` function returns the final resource identified by the DID URL (including fragment). BTW, in your text above you probably meant to say [DID URL Dereferencer](https://w3c.github.io/did-core/#dfn-did-url-dereferencers) rather than [DID resolver](https://w3c.github.io/did-core/#dfn-did-resolvers) (both terms exist in the terminology).

iherman:
> @iherman yes the answer is 1., i.e. the `dereference()` function returns the final resource identified by the DID URL (including fragment). BTW, in your text above you probably meant to say [DID URL dereferencer](https://w3c.github.io/did-core/#dfn-did-url-dereferencers) rather than [DID resolver](https://w3c.github.io/did-core/#dfn-did-resolvers) (both terms exist in the terminology).

Oops, sorry.

> 
[...]

Issue #539, opened 2021-01-11
Clean up herd privacy issues


jandrieu:
Continuing a conversation from PR #480

Section 10.5 Herd Privacy https://w3c.github.io/did-core/#herd-privacy says 
> When a DID subject is indistinguishable from others in the herd, privacy is available. When the act of engaging privately with another party is by itself a recognizable flag, privacy is greatly diminished. DIDs and DID methods need to work to improve herd privacy, particularly for those who legitimately need it most. Choose technologies and human interfaces that default to preserving anonymity and pseudonymity. To reduce digital fingerprints, share common settings across requesting party implementations, keep negotiated options to a minimum on wire protocols, use encrypted transport layers, and pad messages to standard lengths.

However, there is some debate about what herd privacy means and how to apply it to DIDs.

This issue is for developing a consensus definition, after which one or more PRs are expected to be proposed to provide corrections to the DID specifications.

Herd privacy works exactly to the extent that a given feature applies to all DIDs and DID Subjects. That's the herd. When a feature applies to just certain DID Subjects, that enables privacy penetration that would not be possible if true herd privacy exists. When the herd is separable into distinguishable segments, it is possible to discern unintended details about each segment, causing privacy leaks.

For comparison, consider IP v4 addresses which, have excellent herd privacy. Every public IP address is structured and interpreted the same way and reveal nothing about who owns the machine responding to that IP address, what applications are running on that IP, or what type of hardware or network is running at the endpoint. In fact, you can't even tell if that IP address is just a first hop in a forwarding process that ends up with some other machine—with a different IP address—actually doing the work of composing a response.

The exceptions to herd privacy with IP addresses are informative. 

First, there are a few, special, functionally unique private addresses are which do behave differently. There are the private identifier subnets like 192.168., link-local address like 169.254.*, and localhost 127.0.0.0. If the IP address is one of those exceptions, you know something about the initial destination (it's on a private subnet, its likely on a network without DHCP or an assigned IP, and its running on the same machine), but NOTHING else. You still don't know who owns it, what machine is running there, or what applications might be running on it.

Second, the process of assigning IP addresses has a historical legacy that makes it possible to guess, with some level of accuracy who owns a given IP, or at least who secured that IP address from IANA, and often what region of the world that party is from. This was originally done to simplify IP-based routing and the bureaucratic overhead of issuing IP numbers. These optimizations are not part of the IP spec, but rather a management decision that has ongoing consequences.

[...]

PR #544, opened 2021-01-13
Clarify that DID URL dereferencing returns the final resource.

peacekeeper:
This is an attempt to address #399, in order to clarify that DID URL dereferencing returns the final resource (including when the DID URL contains a fragment). How this is implemented is out of scope for DID Core.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/544.html" title="Last updated on Jan 13, 2021, 4:04 PM UTC (68795c4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/544/c8bc6b1...68795c4.html" title="Last updated on Jan 13, 2021, 4:04 PM UTC (68795c4)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -4021,11 +4021,14 @@ represents a full graph merge because the same DID document describes both the
 DID URL Dereferencing
       </h2>
       <p>
-The DID URL dereferencing function dereferences a <a>DID URL</a> into
+The <a>DID URL dereferencing</a> function dereferences a <a>DID URL</a> into
 a resource with contents depending on the <a>DID URL</a>'s components,
 including the <a>DID method</a>, method-specific identifier, path, query,
 and fragment. This process depends on <a>DID resolution</a> of the <a>DID</a>
-contained in the <a>DID URL</a>.
+contained in the <a>DID URL</a>. <a>DID URL dereferencing</a> might involve
+multiple steps (e.g. when dereferencing the fragment of the DID URL),
+and the function is defined to return the final resource after all
+steps are completed.
 The details of how this process is accomplished are outside the scope of this
 specification, but all conformant implementations implement a function
 which has the following abstract form:
--- a/index.html
+++ b/index.html
@@ -4026,7 +4026,7 @@ a resource with contents depending on the <a>DID URL</a>'s components,
 including the <a>DID method</a>, method-specific identifier, path, query,
 and fragment. This process depends on <a>DID resolution</a> of the <a>DID</a>
 contained in the <a>DID URL</a>. <a>DID URL dereferencing</a> might involve
-multiple steps (e.g. when dereferencing the fragment of the DID URL),
+multiple steps (e.g., when the DID URL being dereferenced includes a fragment),
 and the function is defined to return the final resource after all
 steps are completed.
 The details of how this process is accomplished are outside the scope of this


kdenhartog (approved):

index.html
```diff
@@ -4021,11 +4021,14 @@ <h2>
...
       </h2>
       <p>
-The DID URL dereferencing function dereferences a <a>DID URL</a> into
+The <a>DID URL dereferencing</a> function dereferences a <a>DID URL</a> into
 a resource with contents depending on the <a>DID URL</a>'s components,
 including the <a>DID method</a>, method-specific identifier, path, query,
 and fragment. This process depends on <a>DID resolution</a> of the <a>DID</a>
-contained in the <a>DID URL</a>.
+contained in the <a>DID URL</a>. <a>DID URL dereferencing</a> might involve
+multiple steps (e.g. when dereferencing the fragment of the DID URL),
```
TallTed, line 4029:
```suggestion
multiple steps (e.g., when the DID URL being dereferenced includes a fragment),
```

TallTed:
I'm struggling a bit with this --

`This process [of DID URL dereferencing] depends on <a>DID resolution</a> of the <a>DID</a>
contained in the <a>DID URL</a>.`

(And this struggle may not belong here on #544, and nor perhaps on #399.)

But elsewhere (though perhaps this was changed in the past year, and I have overlooked that change?), we've said that the DID, though it is a URL/URI/IRI, must be resolved in order to *get* the DID URL, which literal is therefore presumably not (at least not always, if ever!) the same literal as the DID.

That means, among other things, that the literal of the DID need not be contained within the literal of the DID URL (though we expect the DID, presumably its literal, to be included within the DID Document retrieved when the DID URL is dereferenced).

More, we've said (elsewhere, at least) that resolution of the DID happens _before_ dereferencing of the DID URL, not as a *part* of that dereferencing.

Indeed, how can a DID be separated from a DID URL in order to be resolved, if that resolution doesn't actually occur *after* the dereferencing delivers the DID Document which *does* contain the DID?

What *is* the canonical order of these operations?

agropper:
This is what I'm getting at in this thread, starting here https://github.com/w3c/did-core/issues/539#issuecomment-759280839

peacekeeper:
> we've said that the DID, though it is a URL/URI/IRI, must be resolved in order to _get_ the DID URL

> the literal of the DID need not be contained within the literal of the DID URL

To be honest, both statements sound strange to me. Can you find a reference when/where we said this?

> we've said (elsewhere, at least) that resolution of the DID happens before dereferencing of the DID URL, not as a part of that dereferencing.

[RFC3986](https://tools.ietf.org/html/rfc3986#section-1.2.2) says _"URI "resolution" is the process of determining an access mechanism and the appropriate parameters necessary to dereference a URI"_. So I think you're right about this. If we say anywhere in the spec that resolution is a part of dereferencing, do you want to create a separate PR to fix that?

TallTed:
I think some things must have changed during my hiatus ... and/or I had internalized something that wasn't actually said.

I see the weird differentiation we now have between DID Resolution (get a DID Document from a DID) and DID URL Dereferencing (get *something* from a DID URL, which might be a DID Document, or a fragment thereof, or something entirely external) ... which I think will confuse many, but I guess it's OK if we're internally consistent.

msporny:
@TallTed wrote:
> which I think will confuse many, but I guess it's OK if we're internally consistent.

Yes, I think the same, Ted. I'm equally concerned of that, but I can't quite put my finger on /why/. 

I mean, if you think of it -- this is exactly what happens w/ HTTP -- first you do DNS Resolution, and then you do URL dereferencing. So why should that confuse many?

I think the issue may be the way we're using "Resolution" in the specification is misleading? It might just be the "get me just the data associated with the fragment" that might throw people... that's supposed to happen client-side, not server side.

We may also want to make the distinction between URI Resolution and DNS Resolution (we mean the latter, but I don't think we ever say it). 

URI Resolution is about getting to an absolute URL (not hitting the DNS system):

https://tools.ietf.org/html/rfc3986#section-5

DNS Resolution is about getting data associated with a name. 

Just some thoughts on our Resolution / Dereferencing sections... I had thought I knew what URI Resolution meant, but upon re-reading the spec, it's very different from DNS resolution.

index.html
```diff
@@ -4021,11 +4021,14 @@ <h2>
...
       </h2>
       <p>
-The DID URL dereferencing function dereferences a <a>DID URL</a> into
+The <a>DID URL dereferencing</a> function dereferences a <a>DID URL</a> into
 a resource with contents depending on the <a>DID URL</a>'s components,
 including the <a>DID method</a>, method-specific identifier, path, query,
 and fragment. This process depends on <a>DID resolution</a> of the <a>DID</a>
-contained in the <a>DID URL</a>.
+contained in the <a>DID URL</a>. <a>DID URL dereferencing</a> might involve
+multiple steps (e.g., when the DID URL being dereferenced includes a fragment),
```
msporny, line 4029: Are there any other dereferencing libraries in existence that do this? This might confuse people that are used to their resolution libraries not doing this part of the process for them.

msporny (approved):
I'm kinda sorta ok with this going in, but I agree with @TallTed -- I think we might confuse people w/o some more clarifying language wrt. Resolution vs. Dereferencing.

index.html
peacekeeper: Nobody is saying that the entire dereferencing process has to be implemented in a single library. As we know, in the case of HTTP URLs, the dereferencing processes involves both server and client components. In the case of DID URLs, those details depend on the method and concrete dereferencer implementation. This PR just clarifies that dereferencing returns the final resource after all the steps (including potential "client"-side fragment dereferencing) are completed.

peacekeeper:
> URI Resolution is about getting to an absolute URL (not hitting the DNS system):
> 
> https://tools.ietf.org/html/rfc3986#section-5

@msporny That section is about "URI reference resolution", which is different from "URI resolution" :) For URI resolution, see https://tools.ietf.org/html/rfc3986#section-1.2.2.

msporny:
> @msporny That section is about "URI reference resolution", which is different from "URI resolution" :) For URI resolution, see https://tools.ietf.org/html/rfc3986#section-1.2.2.

Right, I had completely read past the "reference resolution" bit... and even worse, that makes it even more confusing. I was finally able to find the one sentence in the URI specification that speaks to "URI resolution" instead of "URI reference resolution". In any case, we may want to do better than the URI spec here and really spell out what we mean (and provide an example).

index.html
msporny: Hmm, perhaps we could use a less misleading example then... like: "e.g., when the DID URL being dereferenced requires the resolution process to query a Decentralized Ledger."

msporny: Waiting on @peacekeeper to respond to see if he's ok w/ this change before merging.

peacekeeper: @msporny the intent of this PR was to address https://github.com/w3c/did-core/issues/399, which specifically asked the question whether fragment processing is part of resolving/dereferencing or not.

Maybe @iherman could comment if this PR is helpful, or if we can do something else to address his original issue.

msporny:
Editorial, multiple reviews, changes requested and made, non-blocking request for further improvements by @msporny and @TallTed, no objections, merging.

Issue #545, opened 2021-01-13
DID URL dereferencing term definition is inconsistent with its section


peacekeeper:
The "DID URL dereferencing" [definition in the Terminology](https://w3c.github.io/did-core/#dfn-did-url-dereferencing) says:

> The function that takes as its input a DID URL, a DID document, plus a set of dereferencing options, and returns a resource

But the [DID URL Dereferencing section](https://w3c.github.io/did-core/#did-url-dereferencing) doesn't mention that the input can be a DID document. Therefore, "DID document" as an input should be removed from the Terminology definition.

msporny:
PR #546 has been raised. This issue will be closed once that PR has been merged.

msporny:
PR #546 has been merged, closing.

PR #546, opened 2021-01-13
Remove mention of DID document as input in DID URL dereferencing term definition

peacekeeper:
Addresses #545.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/546.html" title="Last updated on Jan 13, 2021, 1:32 AM UTC (1279415)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/546/c8bc6b1...1279415.html" title="Last updated on Jan 13, 2021, 1:32 AM UTC (1279415)">Diff</a>

### DIFF
--- a/terms.html
+++ b/terms.html
@@ -188,12 +188,12 @@ path</a> (with its leading <code>/</code> character), optional <a>DID query</a>
   <dt><dfn>DID URL dereferencing</dfn></dt>
 
   <dd>
-The function that takes as its input a <a>DID URL</a>, a <a>DID document</a>,
-plus a set of dereferencing options, and returns a <a>resource</a>. This
+The function that takes as its input a <a>DID URL</a>
+and a set of input metadata, and returns a <a>resource</a>. This
 resource might be a <a>DID document</a> plus additional metadata, or it might be a
 secondary resource contained within the <a>DID document</a>, or it might be a
-resource entirely external to the <a>DID document</a>. If the function begins
-with a <a>DID URL</a>, it uses the <a>DID resolution</a> function to fetch a
+resource entirely external to the <a>DID document</a>. The function
+uses the <a>DID resolution</a> function to fetch a
 <a>DID document</a> indicated by the <a>DID</a> contained within the
 <a>DID URL</a>. The dereferencing function then can perform additional
 processing on the <a>DID document</a> to return the dereferenced resource


msporny (approved):

jandrieu:
I'm confused by this change. If you don't pass on the entire result from resolution, then resolver *clients* would have to know how to extract the right metadata to pass it to the dereferencer.

In contrast, if you have a resolver that undersands a given method, all the client has to do is pass on the DID Document.

Seems to be that should be the fix: adding the DID Document to the dereferencing contract.

peacekeeper:
> If you don't pass on the entire result from resolution, ...

@jandrieu the section on [DID URL Dereferencing](https://w3c.github.io/did-core/#did-url-dereferencing) currently says:

_"This process depends on DID resolution of the DID contained in the DID URL."_

So this means that of course the DID document from the DID resolution process is used when you dereference a DID URL (similar to how the result of DNS resolution is used when you dereference an HTTP URL).

However does this mean that the DID document should be considered an **input** for the DID URL dereferencing function? Are DNS records an **input** to the HTTP URL dereferencing function?

I've been thinking that DID resolution happens as part of DID URL dereferencing, just like DNS resolution happens as part of HTTP URL dereferencing.

jandrieu:
@peacekeeper I agree with your outline of the flow and the distinction that DNS records aren't used as I am proposing.

However, with 70+ DID Methods and counting, I think we would be better off if the transfer from resolution to dereferencing is method independent. 

If we don't pass on the DID Document, then we not only need method-specific  resolution and dereferencing, we also need method-specific glue to know which parts of the DID Document to pass to dereferencing.

This would significantly limit the flexibility of our extensibility model.

msporny:
> I think we would be better off if the transfer from resolution to dereferencing is method independent.

I don't see how what @peacekeeper is proposing prevents that from happening? Or to put it another way, my mental model matches @peacekeeper's mental model and this sort of interface (between resolution and dereferencing) isn't something that specifications need to define.

jandrieu:
I guess the previous prose established a mental model for me that the dereferencer would have access to the full DID document. Removing that would, for some DID Methods, mean that you need bespoke logic to tease out the relevant logic from the resolver response and the dereferencer input. Which means the API for resolving does not have a simple pass-through to the API for dereferencing.

As for whether or not specifications normally define this, um... we are literally defining the API for this, so I think it does matter.

I'm speaking up on this because I have been working with folks dealing with NFTs and presented exactly the mental model from the prose: that no matter how bespoke your method resolution and dereferencing, you can always count on your dereferencer getting the entire DID Document. 

With this change, that is no longer true. Whoever is managing the integration of resolvers and dereferencers will have to write bespoke code and check it against every method their resolver and dereferencer supports. This accelerates the power-law dynamics of successful DIDs and will retard innovation.

I actually don't see how the universal resolver can address this problem in its current architecture without passing the entire DID Document, not if resolution and dereferencing are independent functions.

@peacekeeper How do you decide, in your resolver, what parts of the DID Document are presented to the dereferencing function? Maybe I'm misinterpreting the flow of data.

msporny:
> As for whether or not specifications normally define this, um... we are literally defining the API for this, so I think it does matter.

We're defining abstract functions... not a concrete API. The latter has a concrete shape with concrete datatypes and parameters to/from the interface. For example, this is what a concrete API looks like:

https://www.w3.org/TR/webauthn-2/#iface-pkcredential

We're definitely not doing that in our spec.

msporny:
> you can always count on your dereferencer getting the entire DID Document.

Yes, but this isn't enough... you also (possibly) need the DID Document metadata and DID Resolution metadata to be returned. So, either we need to pass everything through that the DID Resolution stage gets back... or we just assume the coupling between the dereferencing code and the resolution code handles that hand off.

jandrieu:
@msporny 
> Yes, but this isn't enough... you also (possibly) need the DID Document metadata and DID Resolution metadata to be returned. So, either we need to pass everything through that the DID Resolution stage gets back... or we just assume the coupling between the dereferencing code and the resolution code handles that hand off.

That's right. Everything from resolution should probably be passed to dereferencing, along with the option to add additional input meta-data that did not come from resolution.

FWIW, I'm not saying we are defining a concrete API--whatever distinction that has. I'm saying we are specifying a function signature and when I understood (from the prose) that the DID Document is passed to the dereference function, it afforded a full path without custom code for each different DID Method. Yes, custom code for resolution and dereferencing, but not for the handoff between the two.

In other words, if you, the DID Method, define resolution with these parameters and results and dereferencing with these parameters and results, then you'll have met the contract as specified in the DID Core spec--and no one has to write special code to use your method. Perhaps more importantly, if the DID Resolution spec defines an API with those parameters and results, then it will be conformant with the DID Core spec.

All I'm saying is that there is value from always passing the DID Document that made sense to me and to several others as I've explained how DID extensibility works. Removing the DID Document from that flow means that every DID consumer must be prepared to handle custom logic for each DID Method rather than simply using the API defined by contract in DID Core.

jandrieu:
Either way, I'm good with your decision. I do think we're losing something with this edit, but it's not a major issue.

dlongley (approved):

msporny:
Editorial, multiple reviews, changes suggested but not made, no objections, merging.

Issue #386, opened 2020-09-02
need to clarify revocation vs. rotation


dhh1128:
Creating a new issue to track a tangent to #382.

There appears to be a misalignment in the community about the semantics of changing a key associated with `assertionMethod` in a DID doc. The relevant comments from #382:

https://github.com/w3c/did-core/issues/382#issuecomment-685248497
https://github.com/w3c/did-core/issues/382#issuecomment-685629401

Tagging @jandrieu and @SmithSamuelM who have already commented on the topic.

dhh1128:
@SmithSamuelM already commented on this, but I want to directly answer @jandrieu's question.

The whole point of digital signatures is that they are non-repudiable; they commit the signer to a historical action in a way that they can't deny. This is the basis for holding a party accountable. It's the non-repudiation that makes them useful.

If I use `assertionMethod` to sign a mortgage, I cannot say that when I rotate my key, it invalidates my commitment to pay the mortgage. The bank gave me money, and I incurred a debt. I can certainly sign a mortgage and then rotate my key; this prevents my old key from being used to incur new debts, but it doesn't cancel my old one. Rotation cannot be retroactive, because it would give the signer unilateral power to reinterpret history.

Revocation is different. It presupposes a shared understanding between multiple parties that unambiguous history needs to be given new semantics based on future developments. With that understanding in place, it's possible to construct mechanisms that allow either party on their own, both parties together, or even third parties to apply new semantics to a historical event, no matter how strongly attested. "Yes, Fred really did sign this mortgage. No question about it. But Fred wasn't legally competent when he did so. Therefore the mortgage is invalid -- not because we deny existence of the signature, but because we are viewing the event with different semantics."

We have to be able to tell the difference between rotation (unilaterally change rules for the future) and revocation (joint agreement about changing semantics we apply to the past).

If an issuers uses `assertionMethod` to sign a VC, the issuer cannot say that when they rotate the key, it cancels their commitment to the assertions they once stood behind. That's because an unknown number of verifiers have already made decisions based on the reputational capital the issuer staked against their assertions; allowing the issuer to back out is like letting a lender off the hook for a mortgage without any loan repayment. There exists no shared understanding that the issuer has this unilateral ability to escape accountability and reputational consequences for their actions.

Besides, if rotating a key cancelled all previous assertions, how would we do the other operation (the one that merely changes future possibilities, without attempting to apply a new lens to the past)?

[...]

PR #548, opened 2021-01-16
Split up rotation, revocation and recovery

OR13:
This PR attempts to address the mega thread on https://github.com/w3c/did-core/issues/386

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/548.html" title="Last updated on Jan 23, 2021, 3:07 PM UTC (2494681)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/548/c8bc6b1...2494681.html" title="Last updated on Jan 23, 2021, 3:07 PM UTC (2494681)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -4864,11 +4864,6 @@ and/or verify the integrity of external contexts against the cryptographic
 hash for the context as registered in the DID specification registries
 (see the registration process for more detail) [[?DID-SPEC-REGISTRIES]].
       </p>
-      <p>
-Implementers can also consider other means to protect links to external
-content as best suited to their application. Examples of URL schemes which
-enforce content integrity are [[HASHLINK]] and [[IPFS]].
-      </p>
     </section>
   </section>
 
--- a/index.html
+++ b/index.html
@@ -2917,25 +2917,10 @@ interoperable across representations.
 
             <p>
 It is RECOMMENDED that dereferencing each <a>URI</a> value of the
-<code>@context</code> property results in a document containing machine-readable
-information about the context. These URIs SHOULD be associated with a
-cryptographic hash of the content of the JSON-LD Context as part of registration
-in the DID Specification Registries [[?DID-SPEC-REGISTRIES]].
-            </p>
-            <p class="advisement">
-If included, the cryptographic hash of the content of the JSON-LD Context MUST
-be computed in a manner equivalent to the following:
-            </p>
-            <pre class="example">
-curl -s https://www.w3.org/ns/did/v1 | openssl sha256
-            </pre>
-            <p>
-The command above will result in the following hash, expressed in hexadecimal
-notation:
-            </p>
-            <pre class="example">
-910cd6648f6f7b72a7896a8da83e63460eb8355a0af4b56b699c9281452ac8bb
-            </pre>
+<code>@context</code> property results in a document containing
+machine-readable information about the context. Note that further expectations
+of additional JSON-LD contexts are described as part of the DID specification
+registries registration process.
             </p>
           </dd>
         </dl>
@@ -4862,6 +4847,28 @@ an attack vector that will circumvent verification unless the DID document is
 re-verified.
       </p>
 
+    </section>
+
+    <section>
+      <h2>
+Content Integrity Protection
+      </h2>
+      <p>
+DID documents which include external JSON-LD contexts (see
+<a href="#json-ld"></a>) or any other links to external machine-readable
+content are vulnerable to tampering.
+      </p>
+      <p>
+DID document consumers can cache local static copies of JSON-LD contexts
+and/or verify the integrity of external contexts against the cryptographic
+hash for the context as registered in the DID specification registries
+(see the registration process for more detail) [[?DID-SPEC-REGISTRIES]].
+      </p>
+      <p>
+Implementers can also consider other means to protect links to external
+content as best suited to their application. Examples of URL schemes which
+enforce content integrity are [[HASHLINK]] and [[IPFS]].
+      </p>
     </section>
   </section>
 
--- a/index.html
+++ b/index.html
@@ -3982,7 +3982,11 @@ and without sub-second decimal precision. For example:
 DID document metadata SHOULD include an <code>updated</code> property to
 indicate the timestamp of the last <a href="#update">Update operation</a> for
 the document version which was resolved. The value of the property MUST follow
-the same formatting rules as the <code>created</code> property.
+the same formatting rules as the <code>created</code> property. The updated property
+is omitted if an Update operation has never been performed on the DID document.
+If an <code>updated</code> property exists, it may be the same value as the
+<code>created</code> property when the difference between the two timestamps
+is less than one second.
                 </p>
               </section>
 
--- a/index.html
+++ b/index.html
@@ -3982,7 +3982,7 @@ and without sub-second decimal precision. For example:
 DID document metadata SHOULD include an <code>updated</code> property to
 indicate the timestamp of the last <a href="#update">Update operation</a> for
 the document version which was resolved. The value of the property MUST follow
-the same formatting rules as the <code>created</code> property. The updated property
+the same formatting rules as the <code>created</code> property. The <code>updated</code> property
 is omitted if an Update operation has never been performed on the DID document.
 If an <code>updated</code> property exists, it can be the same value as the
 <code>created</code> property when the difference between the two timestamps
--- a/index.html
+++ b/index.html
@@ -1358,7 +1358,7 @@ property.
               </td>
             </tr>
             <tr>
-              <td><code><a>alsoKnownAs</a</code></td>
+              <td><code><a>alsoKnownAs</a></code></td>
               <td>OPTIONAL</td>
               <td>
                 An <a data-cite="INFRA#ordered-set">ordered set</a> of <a data-cite="INFRA#string">strings</a> that conform to the rules of [[RFC3986]] for <a>URIs</a>.
@@ -3072,7 +3072,7 @@ choosing the shortest byte representation
               </td>
               <td>
 <a data-cite="RFC7049#section-2.1">CBOR floating-point number</a> (major type
-7)</a>. All floating point values MUST be encoded as 64-bits (additional type
+7). All floating point values MUST be encoded as 64-bits (additional type
 value 27), even for integral values.
               </td>
             </tr>
@@ -3255,7 +3255,7 @@ choosing the shortest byte representation
             <tr>
               <td>
 <a data-cite="RFC7049#section-2.1">CBOR floating-point number</a> (major type
-7)</a>.
+7).
               </td>
               <td>
 <a>double</a>
@@ -4353,7 +4353,7 @@ scope of this specification.
         <p>
 The following example demonstrates a JSON-encoded metadata structure that
 might be used as <a href="#did-resolution-input-metadata-properties">DID
-resolution input metadata</a></a>.
+resolution input metadata</a>.
         </p>
 
         <pre class="example"
--- a/index.html
+++ b/index.html
@@ -868,6 +868,16 @@ Description
           <tbody>
             <tr>
               <td>
+<code><a>service</a></code>
+              </td>
+              <td>
+Identifies a service from the <a>DID document</a> by service ID. Support for
+this parameter is REQUIRED. The associated value MUST be an <a data-lt="ascii
+string">ASCII string</a>.
+              </td>
+            </tr>
+            <tr>
+              <td>
 <code>relative-ref</code>
               </td>
               <td>
@@ -878,16 +888,6 @@ a <a>DID document</a> by using the <code>service</code> parameter. Support for
 this parameter is REQUIRED. The associated value MUST be an <a data-lt="ascii
 string">ASCII string</a> and MUST use percent-encoding for certain characters
 as specified in <a data-cite="RFC3986#section-2.1">RFC3986 Section 2.1</a>.
-              </td>
-            </tr>
-            <tr>
-              <td>
-<code><a>service</a></code>
-              </td>
-              <td>
-Identifies a service from the <a>DID document</a> by service ID. Support for
-this parameter is REQUIRED. The associated value MUST be an <a data-lt="ascii
-string">ASCII string</a>.
               </td>
             </tr>
             <tr>
--- a/index.html
+++ b/index.html
@@ -4873,14 +4873,15 @@ Keep Personally-Identifiable Information (PII) Private
       </h2>
 
       <p>
-If a <a>DID method</a> specification is written for a public <a>verifiable data
-registry</a> where all <a>DIDs</a> and <a>DID documents</a> are publicly
+If a <a>DID method</a> specification is written for a public <a>verifiable
+data registry</a> where all <a>DIDs</a> and <a>DID documents</a> are publicly
 available, it is <em>critical</em> that <a>DID documents</a> contain no
 personally-identifiable information.
       </p>
       <p>
-Due
-diligence should be taken around the use of URLs in service endpoints
+Personally-identifiable information can instead be placed behind <a>service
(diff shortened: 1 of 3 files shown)


dhh1128 (changes requested):
I cannot sign off on a characterization of revocation as undoing the past. Revocation means stopping the use of something in the future, and cannot be used to repudiate past actions.

It is FINE for DID methods to not support revocation. It is not FINE for us to define out of existence a feature that's important for some of them.

index.html
```diff
@@ -4545,28 +4545,172 @@ <h2>
...
+    Section <a href="#method-operations"></a> specifies the <a>DID</a> operations to
+    be supported by a <a>DID method</a> specification, 
+    including <a href="#update">update</a>
+    which MUST be used to perform a verification method rotation.
+  </p>
+  
+  <p>
+    A controller performs a rotation when they change 
+    <code>publicKeyJwk</code> or <code>publicKeyBase58</code>, or other verification material
+    without changing the <code>id</code> of the verification method.
```
dhh1128, line 4578: We don't constrain an `id` to always point to an item of the same type, let along the same verification method. That is, `#key1` could be a service endpoint in one version of a DID doc, a signing key in another version, and an encryption key in another version. What you mean here, I think, is that we change *only* the value of a key, but none of its associated semantics/role/purpose.

index.html
```diff
@@ -4545,28 +4545,172 @@ <h2>
...
+  </p>
+  
+  <p>
+    Unlike revocation, rotation allows for new verifications associated with the <code>id</code>, 
+    to be valid, while invalidating old verifications that were produced with the older material.
```
dhh1128, line 4583: I don't agree that key rotation invalidates old verifications. That is not the case with passwords or SSH keys; why should we claim it is true of DID keys? You are wanting to use verification methods to produce ephemeral tokens and are claiming that is a general property of the entire DID ecosystem. I disagree.

index.html
```diff
@@ -4545,28 +4545,172 @@ <h2>
...
+</h2>
+
+<p>
+  Verification method revocation is a reactive security measure.
+</p>
+
+<p>
+  If a verification method is no longer exclusivley accessible to the 
+  controller or parties trusted to act on behalf of the controller, 
+  it should be revoked immediatly to reduce the risk of masquerading, theft and fraud.
```
dhh1128, line 4615: "immediatly" --> "immediately"

index.html
```diff
@@ -4545,28 +4545,172 @@ <h2>
...
+</p>
+
+<p class="advisement">
+  It is considered a best practice to support key revocation.
+</p>
+
+<p class="advisement">
+A controller should revoke any verification method that is believed to be compromised immediatly.
```
dhh1128, line 4623: "immediatly" --> "immediately"

index.html
```diff
@@ -4545,28 +4545,172 @@ <h2>
...
+</p>
+
+<p class="advisement">
+Cryptographic verifications associated with revoked verification methods should be considered invalid
```
dhh1128, line 4627: I could possibly live with the following rewording:

Cryptographic verifications associated with a revoked verification method are not valid. However, knowing whether a verification was made with a revoked verification method is trickier than it may seem.

Some DID methods provide the ability to look back at the state of a DID at a point in time, or at a particular version of the DID document. When such a feature is combined with a reliable way to determin the time or DID version that existed when a cryptographically verifiable statement was made, then revocation does not undo that statement. This can be the basis for using DIDs to make binding commitments (e.g., to sign a mortgage). If these conditions are met, revocation is not retroactive; it only nullifies future use of the method.

But in order for such semantics to be safe, the second condition -- an ability to know what the state of the DID document was at the time the assertion was made -- MUST apply. Without that guarantee, someone could discover a revoked key and use it to make cryptographically verifiable statements with a simulated date in the past.

Some DID methods only allow the retrieval of the current state of a DID. When this is true, or when the state of a DID at the time of cryptographically verifiable statement cannot be reliably determined, then the only safe interpretation of revocation is to make it apply both forward and backward in time. DID ecosystems that take this approach essentially provide cryptographically verifiable statements as ephemeral tokens that can be invalidated at any time by the DID controller.

index.html
```diff
@@ -4545,28 +4545,172 @@ <h2>
...
+Even if a verification method is present in a DID Document, 
+additional information such as a public key revocation certificate,
+or external allow or deny list might be used to determine if a verification method has been revoked.
+</p>
+
+
+
+<p class="note">
+An attacker can trivially back date signatures to a period before the verification method was revoked, 
+this is the reason that revocation applies to verifications for all time,
```
dhh1128, line 4652: I suggest removing this paragraph in favor of the wording I provided in a previous comment.

dhh1128 (changes requested):

index.html
dlongley: @dhh1128, I agree with your reword. I think it's a little strange to talk about the "state of a DID" (which is just an identifier), so we should perhaps instead talk about DID Documents or resolution. That's just an editorial quibble though. A slightly more important suggestion, I think, would be such that we don't so strongly state that if a DID method only allows retrieval of the current state of a DID that there are no alternative solutions to preventing undesirable "retroactive revocation". We shouldn't be so prescriptive -- instead of saying:

> Some DID methods only allow the retrieval of the current state of a DID.

We can say:

> In some cases, it may only be possible to obtain a current DID Document.

That way, future clever solutions that layer on top of DID methods that do not support versioning directly are not needlessly discouraged.

dhh1128: Good point about "state of a DID." I once poked a colleague about that very error, so I'm being inconsistent to get sloppy now. :-)

I agree that the rewording you suggested improves the text.

index.html
TallTed, line 4623:
```suggestion
A controller should immediately revoke any verification method that is believed to be compromised.
```

index.html
```diff
@@ -4545,28 +4545,172 @@ <h2>
...
+  <p>
+    Section <a href="#method-operations"></a> specifies the <a>DID</a> operations to
+    be supported by a <a>DID method</a> specification, 
+    including <a href="#update">update</a>
+    which MUST be used to perform a verification method rotation.
+  </p>
+  
+  <p>
+    A controller performs a rotation when they change 
+    <code>publicKeyJwk</code> or <code>publicKeyBase58</code>, or other verification material
```
TallTed, line 4577:
```suggestion
    <code>publicKeyJwk</code>, <code>publicKeyBase58</code>, or other verification material
```

index.html
```diff
@@ -4545,28 +4545,172 @@ <h2>
...
+    reduces the value of a compromised verification method to an attacker.
+  </p>
+  
+  <p class="note">
+    Higher security environments tend to employ more frequent verification method rotation.
+  </p>
+  
+  <p class="note">
+    Frequent rotation of a verification method might be frustrating 
+    for parties that are forced to continiously renew or refresh associated credentials.
```
TallTed, line 4598:
```suggestion
    for parties that are forced to continuously renew or refresh associated credentials.
```

index.html
TallTed, line 4615:
```suggestion
  it should be revoked immediately to reduce the risk of masquerading, theft and fraud.
```

index.html
```diff
@@ -4545,28 +4545,172 @@ <h2>
...
-<a href="#did-controller"></a>. Not all <a>DID method</a>
+be supported by a <a>DID method</a> specification, 
+including <a href="#update">update</a> and <a href="#deactivate">deactivate</a> 
+which MAY be used to remove verification method from a DID Document.
+</p>
+
+<p>
+Even if a verification method is present in a DID Document, 
+additional information such as a public key revocation certificate,
+or external allow or deny list might be used to determine if a verification method has been revoked.
```
TallTed, line 4645:
```suggestion
or external allow or deny list, might be used to determine if a
verification method has been revoked.
```

index.html
dlongley: Yeah, I'm not a fan of this approach either -- and it's not really what I consider "rotation" either. Rather, I'd expect rotation to involve *adding* new verification methods to the DID Document without removing existing ones (and thus, their proofs continue to be valid). Internally/privately, the private key material/access to it is changed such that it is either made inactive or destroyed (but this is the private aspect of rotation, having nothing to do with the DID Document). I don't think we should encourage swapping key materials while keeping IDs the same -- I think that's a recipe for confusion.

OR13:
@dhh1128 thanks for your comments, I intend to address them, I am hoping to see more comments from others before a try though.... one area of concern I have is aligning 'revocation' in did core, with TLS and PGP revocation... because if we do things differently I worry we may make things more confusing (its possible my PR is making this worse).

- https://www.digicert.com/kb/util/utility-test-ocsp-and-crl-access-from-a-server.htm
- http://www.spywarewarrior.com/uiuc/ss/revoke/pgp-revoke.htm

^ these are from google, I would welcome better links to add to did core for informative references.

As I understand TLS and PGP, revocation destroys verification. I think you have said you don't want that to be true of DIDs... This is the primary issue we need to address in this PR.

index.html
OR13: I agree with both your comments, perhaps I should remove this section on rotation... its telling that the word "rotation" does not appear once here: https://cheatsheetseries.owasp.org/cheatsheets/Key_Management_Cheat_Sheet.html

OR13: However what I wrote is in line with https://developer.okta.com/docs/concepts/key-rotation/

index.html
OR13: removing an ssh key or a jwk from the associated collection does cause a verify to reject them... perhaps you can propose some alternate language that applies equally to SSH / GPG / OIDC?

GitHub will not accept commits from "rotated" ssh keys.

index.html
OR13: I don't believe the proposed language is an improvement.

However, I hope I have captured its spirit in my latest change set, by stating that revocation and rotation only apply to the latest version of a did document.... and clarifying how revoked material is commonly treated by other software systems.

index.html
OR13: I have made changes attempting to address your concern with this paragraph.

OR13:
since I am. not a git wizard, and this branch has git issued beyond my skill I will be reimplementing it.

OR13:
Originally this branch was meant to replace:

```
 <section>
      <h2>
Key Revocation and Recovery
      </h2>

      <p>
It is up to <a>DID methods</a> to define how revocation of cryptographic keys
might occur. <a>DID method</a> specifications are also expected
to enable support for a quorum of trusted parties to enable key recovery. Some
of the facilities to do so are suggested in Section
<a href="#did-controller"></a>. Not all <a>DID method</a>
specifications will recognize control from <a>DIDs</a> registered using other
<a>DID methods</a> and they might restrict third-party control to <a>DIDs</a>
that use the same method. Access control and key recovery in a
<a>DID method</a> specification can also include a time lock feature to
protect against key compromise by maintaining a second track of control for
recovery.
</p>
    </section>
```

with: 


```
 <section>
      <h2>Verification Method Rotation</h2>
  
      <p>Verification method rotation is a proactive security measure.</p>
  
      <p>
        Verification method rotation applies only to the current or latest
        verison of a DID Document.
      </p>
  
      <p>
        When a verification method has been active for a long time, or used
        for many operations, a controller might wish to perform a rotation.
      </p>
  
      <p class="advisement">
        It is considered a best practice to perform verification method
        rotation on a regular basis.
      </p>

      <p class="advisement">
        Proofs or signatures that rely on verification methods that are not
        present in the latest version of a DID Document are not impacted by
        rotation, and may require additional information to mitigate compromise.
      </p>
  
      <p>
        Section <a href="#method-operations"></a> specifies the
        <a>DID</a> operations to be supported by a
        <a>DID method</a> specification, including
        <a href="#update">update</a>
        which MUST be used to perform a verification method rotation.
      </p>
  
      <p>
        A controller performs a rotation when they add a new verification
        method that is meant to replace an existing verification method after
        some time.
      </p>
  
      <p>Not all DID Methods support verification method rotation.</p>
  
      <p>
        Unlike revocation, rotation allows for new verifications associated
        with the <code>id</code>, to be valid, while invalidating older
        verifications once the older verification method is removed.
      </p>
  
      <p>
        Rotation is a useful mechanism for protecting against verification
        method compromise, since frequent rotation of a verification method by
        the controller, reduces the value of a compromised verification method
        to an attacker.
      </p>
  
      <p class="note">
        Higher security environments tend to employ more frequent verification
        method rotation.
      </p>
  
      <p class="note">
        Frequent rotation of a verification method might be frustrating for
        parties that are forced to continiously renew or refresh associated
        credentials.
      </p>
    </section>
  
    <section>
      <h2>Verification Method Revocation</h2>
  
      <p>Verification method revocation is a reactive security measure.</p>

      <p>
        Verification method revocation applies only to the current or latest
        verison of a DID Document.
      </p>
  
      <p>
        If a verification method is no longer exclusivley accessible to the
        controller or parties trusted to act on behalf of the controller, it
        should be revoked immediatly to reduce the risk of masquerading, theft
        and fraud.
      </p>
  
      <p class="advisement">
        It is considered a best practice to support key revocation.
      </p>
  
      <p class="advisement">
        A controller should revoke any verification method that is believed to
        be compromised immediatly.
      </p>
  
      <p class="advisement">
        Cryptographic verifications associated with revoked verification
        methods should be considered invalid regardless of when revocation
        occurs.
      </p>
  
      <p class="note">
        As described in Section <a href="#verification-method-types"></a>,
        absence of a verification method is the only form of revocation that
        applies to all DID Methods.
      </p>
  
      <p>
        Section <a href="#method-operations"></a> specifies the
        <a>DID</a> operations to be supported by a
        <a>DID method</a> specification, including
        <a href="#update">update</a> and <a href="#deactivate">deactivate</a>
        which MAY be used to remove verification method from a DID Document.
      </p>

      <p>Not all DID Methods support verification method revocation.</p>
  
      <p>
        Even if a verification method is present in a DID Document, additional
        information such as a public key revocation certificate, or external
        allow or deny list might be used to determine if a verification method
        has been revoked.
      </p>
  
      <p class="note">
        Compromise of a verification method allows the attacker access to the verification 
        relationship of the controller, for example authentication.
        The attacker is indistinguishable from the legitimate controller from the 
        time the compromised verification method was registered, to the time it was revoked.
      </p>

      <p class="note">
        The day-to-day operation of any software relying on a compromised verification method 
        might be impacted by a user’s operating system, antivirus, or endpoint protection software 
        when the verification method is publicly revoked.
      </p>

    </section>
  
    <section>
      <h2>DID Recovery</h2>
  
      <p>
        Recovery is a reactive security measure, whereby a controller is able
        to regain the ability to perform DID operations.
      </p>
  
      <p class="advisement">
        Recovery is advised when a controller or services trusted to act on
        their behalf no longer has the exclusive ability to perform DID
        operations as described in <a href="#method-operations"></a>.
      </p>
  
      <p class="advisement">
        It is considered a best practice to never reuse a verification method
        or key material associated with recovery for any other purposes.
      </p>
  
      <p>
        Recovery is commonly performed in conjunction with
        <a href="#verification-method-rotation"
          >verification method rotation</a
        >
        and
        <a href="#verification-method-revocation"
          >verification method revocation</a
        >.
      </p>
  
      <p>
        There are no common recovery mechanisms that apply to all DID Methods.
      </p>
  
      <p>
        <a>DID method</a> specifications might choose to enable support for a
        quorum of trusted parties to faciliate recovery. Some of the
        facilities to do so are suggested in Section
        <a href="#did-controller"></a>.
      </p>
  
      <p>
        Not all <a>DID method</a> specifications will recognize control from
        <a>DIDs</a> registered using other <a>DID methods</a> and they might
        restrict third-party control to <a>DIDs</a> that use the same method.
      </p>
  
      <p>
        Access control and recovery in a <a>DID method</a>
        specification can also include a time lock feature to protect against
        key compromise by maintaining a second track of control for recovery.
      </p>
  
      <p class="note">
        Performing recovery proactively on an infrequent but regular basis,
        can help to ensure that control has not been lost.
      </p>
    </section>
  
  ```

PR #553, opened 2021-01-18
Change DID parameter names to camelCase.

peacekeeper:
In https://github.com/w3c/did-core/pull/533, metadata property names were changed from using hyphens to camelCase (e.g. `content-type` -> `contentType`). That same PR also changed both the `version-id` metadata property and the `version-id` DID parameter to `versionId`.

This PR here changes the remaining two concerned DID parameters `relative-ref` and `version-id` to be consistent with the others.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/553.html" title="Last updated on Jan 18, 2021, 10:21 AM UTC (935f753)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/553/af9c06b...935f753.html" title="Last updated on Jan 18, 2021, 10:21 AM UTC (935f753)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -878,7 +878,7 @@ string">ASCII string</a>.
             </tr>
             <tr>
               <td>
-<code>relative-ref</code>
+<code>relativeRef</code>
               </td>
               <td>
 A relative URI reference according to <a
@@ -903,7 +903,7 @@ data-lt="ascii string">ASCII string</a>.
             </tr>
             <tr>
               <td>
-<code>version-time</code>
+<code>versionTime</code>
               </td>
               <td>
 Identifies a certain version timestamp of a <a>DID document</a> to be resolved.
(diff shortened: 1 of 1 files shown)


brentzundel (approved):

kdenhartog (approved):

OR13:
I find pascal case harder to read, and therefore a poor choice for URIs.... what we are doing is establishing a convention here. I am fine with pascal case is the convention... I would prefer some explanation for why we are encouraging this, over just making the change... future readers will have questions.

OR13 (changes requested):
I would prefer to see a `note` or `advisement` regarding this change, and noting that it is a convention, not a requirement.

peacekeeper:
> I am fine with pascal case as the convention

@OR13 "pascal case" I think would be "RelativeRef" or "VersionId"? I don't think anyone has ever suggested or even mentioned that... Are you proposing we switch to that?

msporny (approved):

OR13:
yes, sorry, I always confuse PascalCase, TitleCase and camelCase.

PR #554, opened 2021-01-18
Adding the `verificationMethod` as an optional property for service endpoints' table

iherman:
[Example 22](https://w3c.github.io/did-core/#example-22-various-service-endpoints) suggests that it is possible to use the `verificationMethod` as part of a service endpoint, although that is not explicitly listed in [§5.2.5](https://w3c.github.io/did-core/#service-endpoints). That being said, it looks like an omission; using a verification method defined elsewhere in the DID document makes perfect sense.

This PR adds the entry to the table in [§5.1.3](https://w3c.github.io/did-core/#core-properties-for-a-service-endpoint). Note that, in contrast to the behaviour of the property when used for a DID Document, its value, in this case, is a single URL value, i.e., not a full Verification Method instance (I presume all verification methods SHOULD be defined on the DID Document level). 

Also, I have not defined it as an array either; maybe it does make sense to have an array of verification references, maybe not, I am not sure.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/554.html" title="Last updated on Jan 18, 2021, 10:55 AM UTC (a7d4e82)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/554/af9c06b...a7d4e82.html" title="Last updated on Jan 18, 2021, 10:55 AM UTC (a7d4e82)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1500,6 +1500,13 @@ property.
               </td>
             </tr>
             <tr>
+              <td><code><a>verificationMethod</a></code></td>
+              <td>OPTIONAL</td>
+              <td>
+                A <a data-cite="INFRA#string">string</a> that conforms to the rules of [[RFC3986]] for <a>URIs</a>. The <a>URI</a> value SHOULD also conform to the rules in Section <a href="#did-url-syntax"></a>.
+              </td>
+            </tr>
+            <tr>
               <td><code><a>serviceEndpoint</a></code></td>
               <td>OPTIONAL</td>
               <td>


msporny:
Hrm, I think that was a old, completely made up example -- and as far as I know, there are zero implementations that do that today. :) -- so we might want to go the other way -- remove that example from the spec.

iherman:
> Hrm, I think that was a old, completely made up example -- and as far as I know, there are zero implementations that do that today. :) -- so we might want to go the other way -- remove that example from the spec.

@msporny I can go either way. My only concern, raising this PR, is to ensure the consistency of the spec. But I think this should be decided prior to CR.

iherman:
For the records: this is linked to #404

kdenhartog:
If I remember correctly, the intent behind this was that a specific verification method was intended to represent the identity hub service. Given there's been significant changes to this work (now takes the form of the confidential storage work), I think it's fine for us to remove this. Let's get @csuwildcat to confirm if he still thinks this type of approach is needed. As a side note, it appears when @csuwildcat originally added this example he used the property name `publicKey` and @rhiaro changed it to `verificationMethod` as we were transitioning to that new vocabulary.

csuwildcat:
I'd remove this, as it's an outdated example.

peacekeeper:
If anything, wouldn't we want a verification relationship such as `authentication` rather than the `verificationMethod` property on the service level?

Either way, my recollection is that early on we wanted to express per-service keys, but I don't think this has really come up again in the last year or so. So I don't have a strong preference here, but my sense is also that we can just remove this example and not merge this PR.

iherman:
I have just created #565 as an alternative to this PR: removal of the example with a verification method in a service. If that PR is merged, this one should be closed without any further action; vice versa: if this PR is merged than #565 should be closed without any further actions.

Pinging @msporny to take the appropriate actions.

msporny:
> Pinging @msporny to take the appropriate actions.

Closing this PR in preference of PR #565, which seems like it has a better chance of reaching consensus. Folks can request that this PR is reopened if they disagree.

Issue #452, opened 2020-10-29
Compound key is not clear and points to wikipedia not a normative definition


jonnycrunch:
In an attempt at solidifying concrete representations of key formats I found that `compound key` description of the `id` property in the [Verification Methods](https://www.w3.org/TR/did-core/#verification-methods) section lacking in substance.  In points to [Composite_key](  https://en.wikipedia.org/wiki/Composite_key) on wikipedia.  I need something more explicit and unambiguous to complete the CDDL representations and I find the current MAY, MUST, RECOMMENDED insufficient to implement the representations in CDDL.  

The text currently states MUST be URI, but seems that this really should be a MUST be a [DID-URL](https://www.w3.org/TR/did-core/#dfn-did-urls)

OR13:
agree, perhaps we should just say a DID URL with a fragment.

OR13:
by compound key, we mean: "did:example:123#kid` per https://tools.ietf.org/html/rfc7638

jonnycrunch:
@OR13, yes In general that is what belong on the right-hand side of the `#`, but the concept of string concatenation of `did:example:123` with '#` and the fingerprint with what need to be normatively capture for the abstract data model.  


here is what I think would be the best `rule`:

https://tools.ietf.org/html/rfc5234#section-3.1


and for representation specific implementations in CDDL: 

https://tools.ietf.org/id/draft-bormann-cbor-cddl-control-02.html#section-2.1

OR13:
Really, we are saying that `id` is an IRI that can contain a fragment, where the fragment is determined according to another RFC, I welcome any PRs that better express this.

msporny:
PR #571 addresses this issue, which will be closed when that PR is merged.

msporny:
PR #571 has been merged, closing.

PR #571, opened 2021-01-23
Clean up Core Properties

rhiaro:
Following @iherman's addition of a summary table to Core Properties, I did a full review of the section to remove redundant language, make sure there is no duplicate normative language and no inconsistencies, and adjust the structure a little to improve readability.

I marked the summary table as explicitly non-normative. All normative language remains in the subsequent sections which contain the full property definitions and longer descriptions of what the properties are for.

https://github.com/w3c/did-core/commit/ffb6c2b056189e8af310c8bc4d2885e4e994a16a replaces the algorithm for getting verification methods by reference with a couple of sentences instead. I can roll this back if necessary, but I find it less jarring to read in the context of the rest of the section.

I touched on #452 in the process, possibly enough to close it if this is merged?

I should not have changed any normative statements, except where they conflicted directly with other normative statements and I knew (99%) them to be wrong.

The first commit is fixing line breaks, so it looks like every single line has changed. You may wish to review commit by commit, rather than all at once to see changes more easily.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/571.html" title="Last updated on Jan 31, 2021, 7:23 PM UTC (7d81c2c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/571/b8f8696...7d81c2c.html" title="Last updated on Jan 31, 2021, 7:23 PM UTC (7d81c2c)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1320,19 +1320,44 @@ describe relationships between the <a>DID subject</a> and the value of the
 property.
     </p>
 
-
-    <p>This document defines three classes of core properties:</p>
+    <p>
+This document defines three classes of core properties:
+    </p>
 
     <ol>
-      <li>properties defined for a <a>DID Document</a>, serializing the topmost <a data-cite="INFRA#ordered-map">map</a> in the  <a href="#data-model">data model</a>;</li>
-      <li>properties defined for a <a>Verification Method</a>, serializing a concept represented by a <a data-cite="INFRA#ordered-map">map</a> in the <a href="#data-model">data model</a>, and specified in Section <a href="#verification-methods"></a>; </li>
-      <li>properties defined for a <a>Service Endpoint</a>, serializing a concept represented by a <a data-cite="INFRA#ordered-map">map</a> in the <a href="#data-model">data model</a>, and specified in Section <a href="#service-endpoints"></a>.</li>
+      <li>
+properties defined for a <a>DID Document</a>, serializing the topmost
+<a data-cite="INFRA#ordered-map">map</a> in the <a href="#data-model">data
+model</a>;
+      </li>
+      <li>
+properties defined for a <a>Verification Method</a>, serializing a concept
+represented by a <a data-cite="INFRA#ordered-map">map</a> in the
+<a href="#data-model">data model</a>, and specified in
+Section <a href="#verification-methods"></a>;
+      </li>
+      <li>
+properties defined for a <a>Service Endpoint</a>, serializing a concept
+represented by a <a data-cite="INFRA#ordered-map">map</a> in the
+<a href="#data-model">data model</a>, and specified in Section
+<a href="#service-endpoints"></a>.
+      </li>
     </ol>
 
-    <p class="note" title="Repeated properties">Some of the properties (e.g., <code>id</code> or <code>type</code>) are present in different classes but with possible differences in the associated constraints. Hence their repeated presence in the tables in Section <a href="#property-summary"></a>.</p>
+    <p class="note" title="Repeated properties">
+Some of the properties (e.g., <code>id</code> or <code>type</code>) are
+present in different classes but with possible differences in the associated
+constraints. Hence their repeated presence in the tables in Section
+<a href="#property-summary"></a>.</p>
 
     <p class="note" title="Ordering of values">
-      As a result of the <a href="#data-model">data model</a> being defined using terminology from [[INFRA]], property values which can contain more than one item, such as <a data-cite="INFRA#ordered-map">maps</a> and <a data-cite="INFRA#ordered-set">sets</a>, are explicitly ordered. For the purposes of this specification, unless otherwise stated, ordering is not important and implementations are not expected to produce or consume deterministically ordered values.
+As a result of the <a href="#data-model">data model</a> being defined using
+terminology from [[INFRA]], property values which can contain more than one
+item, such as <a data-cite="INFRA#ordered-map">maps</a> and
+<a data-cite="INFRA#ordered-set">sets</a>, are explicitly ordered. For the
+purposes of this specification, unless otherwise stated, ordering is not
+important and implementations are not expected to produce or consume
+deterministically ordered values.
     </p>
 
     <section>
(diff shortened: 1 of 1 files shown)


OR13 (approved):

index.html
```diff
@@ -1320,190 +1320,247 @@ <h1>Core Properties</h1>
...
               </td>
             </tr>
             <tr>
               <td><code><a>verificationMethod</a></code></td>
-              <td>OPTIONAL</td>
-              <td>
-                An <a data-cite="INFRA#ordered-set">ordered set</a> of either <a>Verification Method</a> <a data-cite="INFRA#ordered-map">maps</a> (see <a href="#core-properties-for-a-verification-method"></a>) or <a data-cite="INFRA#string">strings</a> that conform to the rules of [[RFC3986]] for <a>URIs</a>. <br/>The <a>URI</a> values SHOULD also conform to the rules in Section <a href="#did-url-syntax"></a>.
+              <td>no</td>
+              <td rowspan="6">
+An <a data-cite="INFRA#ordered-set">ordered set</a> of either <a>Verification
```
iherman, line 1422: The core text §5.4 says "If present, the value MUST be an ordered set of verification methods, where each verification method is described by a map. ". The possibility to use, instead, a DID-URL is only allowed for the other properties like `keyAgreement` and the like.

(The table was wrong before, because I did not realize this subtle difference. This means that JSON Schema must be updated, too, to reflect this difference. This difference cannot be expressed in RDFS (the verification methods themselves always have a URI, ie, no blank nodes are allowed) and I will have to check the SHACL and SHEX cases.)

index.html
```diff
@@ -1703,65 +1748,51 @@ <h2>Verification Methods</h2>
...
+registered in the DID Specification Registries [[?DID-SPEC-REGISTRIES]].
+          </p>
 
           <p>
-A verification method MUST contain verification material, such as
-<a>publicKeyJwk</a> or <a>publicKeyBase58</a>.
+A verification method MUST contain a property to express verification
+material. The property used is determined by the value of the
+<code>type</code> property. Examples of such properties are
+<code><a>publicKeyJwk</a></code> or <code><a>publicKeyBase58</a></code>.
```
iherman, line 1795: The terminology "such properties are ..." made clear a different error in the table before (and in the SHACL and SHEX constraints as well as the JSON Schema). Indeed, in all those cases the constraints say that _either_ a `publicKeyJwk` _or_ a `publicKeyBase58`, but one of the two must be present.

To avoid such a misunderstanding, it may be worth adding a reference to [§4.4. of the registry document](https://www.w3.org/TR/did-spec-registries/#verification-method-properties) to refer to other possible properties that can be used instead.

index.html
rhiaro: I wasn't 100% sure about this, so I'm glad you picked up on it. I want to double check with the editors that values of  `verificationMethod` can never contain only a reference and must always be embedded as you say, in which case I'll fix the table. However, the algorithm for embedded vs referenced verification method objects that has been present since the `publicKey` days implies that the value of `verificationMethod` could potentially reference a verification method elsewhere, eg. in another DID document. In which case the table stays the same and the normative language in the `verificationMethod` definition needs updating. But it is ambiguous.

iherman: From a purely JSON-LD point of view (I know that should not necessarily influence the model) there is no way of checking this, unless the pure JSON structure is also imposed. So I would prefer to allow the usage of a URL as part of the possible values of a `verificationMethod` to avoid this discrepancy on the JSON-LD level.

rhiaro: Confirmed with @msporny that the value of `verificationMethod` should NOT be a URL, only a fully qualified verification method, so I will update the table accordingly.

iherman: @msporny, are you sure?

Consider these two JSON-LD fragments:

- https://tinyurl.com/y4xdj83l (on JSON-LD Playground): taking an example from the spec
- https://tinyurl.com/yyavt36k (on JSON-LD Playground): same example but reformulated in a (clearly uglier) JSON-LD

From a JSON-LD point of view, _these two JSON-LD fragments are identical_, as you can see the N-Quads versions which are literally identical. This means that the JSON-LD serialization of the model is restricted to a particular "profile" of JSON-LD (which, I guess, can be expressed through a JSON Schema). 

If there is a good reason for this I will not object. But let us make it sure that we go into something not really clean with an open eye.

cc @dlongley

iherman: On more thought on https://github.com/w3c/did-core/pull/571#discussion_r563848624 (almost a day later:-): 

I see now that the text says, in [§6.3.1](https://w3c.github.io/did-core/#production-0) that

> The DID document MUST be serialized according to the production rules for JSON, with one additional requirement: The DID document MUST include the @context property.

which means, indeed, that the [second playground example](https://tinyurl.com/yyavt36k) would not be an acceptable DID Document in JSON-LD. I still do not understand why this restriction is necessary, but I accept that the spec is at least consistent.

But then, I believe, we should add a note that emphasizes that the conformance of the DID Document, when serialized in JSON-LD, is _not_ based on the semantic mapping of JSON-LD on Linked Data. From a Linked Data point of view, making this clear is important.

I also wonder whether it is possible to publish a [JSON-LD Frame document](https://www.w3.org/TR/json-ld11-framing/) that could be used to take any arbitrary, but semantically equivalent JSON-LD (like the aforementioned second playground example). We could then also say that any JSON-LD document, that expresses a DID Document, when transformed using that Frame, should abide to the constraints in the spec. I have no experience with JSON-LD Framing, but we may get some help from some people who have (@dlongley? @davidlehn?)

index.html
```diff
@@ -1320,190 +1320,252 @@ <h1>Core Properties</h1>
...
+represented by a <a data-cite="INFRA#ordered-map">map</a> in the
+<a href="#data-model">data model</a>, and specified in Section
+<a href="#service-endpoints"></a>.
+      </li>
     </ol>
 
-    <p class="note" title="Repeated properties">Some of the properties (e.g., <code>id</code> or <code>type</code>) are present in different classes but with possible differences in the associated constraints. Hence their repeated presence in the tables in Section <a href="#property-summary"></a>.</p>
+    <p>
+The core properties are summarized informatively in Section
+<a href="#summary"></a>, and described in detail and specificed
```
msporny, line 1349:
```suggestion
<a href="#summary"></a>, and described in detail and specified
```

index.html
```diff
@@ -1320,190 +1320,252 @@ <h1>Core Properties</h1>
...
+      <h2>Summary</h2>
+
+      <p>
+This section contains an informative reference for the core properties defined
+by this specification, with expected values and whether or not they are
+required. The property names in the tables are linked to the normative
+definitions and more detailed descriptions of each property.
+      </p>
+      <p class="note" title="Property names used in multiple classes">
+  The property names <code>id</code>, <code>type</code> and
```
msporny, line 1374:
```suggestion
  The property names <code>id</code>, <code>type</code>, and
```

index.html
```diff
@@ -2018,6 +2045,31 @@ <h2>Authentication</h2>
...
+                      containing three verification methods">
+{
+  "@context": "https://www.w3.org/ns/did/v1",
+  "id": "did:example:123456789abcdefghi",
+  <span class="comment">...</span>
+  "authentication": [
+    <span class="comment">// this method can be used to authenticate as did:...fghi</span>
+    "did:example:123456789abcdefghi#keys-1",
+    <span class="comment">// this method can be used to authenticate as did:...fghi</span>
+    "did:example:123456789abcdefghi#biometric-1",
```
msporny, line 2058:
```suggestion
    <span class="comment">// this method can be used to authenticate as did:...fghi</span>
    "did:example:123456789abcdefghi#biometric-1",
```
I'm fairly certain the group would rail against the notion of embedding a biometric in the DID Document at this point... removing it from the example.

msporny (approved):
Minor nits, which I'll merge, otherwise LGTM.

msporny:
Merge conflicts @rhiaro, please fix.

rhiaro:
@msporny merge conflicts resolved

index.html
rhiaro: The constraints about top level `id` and and internally nested verification method objects apply to the abstract data model, so I would argue that serializing them to JSON-LD that doesn't exactly match that description but that is functionally the same (as in the second playground example) would be acceptable. If we were hoping to convert between plain JSON and JSON-LD directly it might be a problem, but since we're supposed to go via the ADM in the middle, maybe that's okay? The main thing with the nested verification methods is that you shouldn't be going out to another document to fetch them. In the case of the second playground example, that's still fine.

msporny:

msporny:
Apologies, @rhiaro -- I had to squash and merge -- couldn't rebase due to conflicts. Doing git merge doesn't always allow future rebasing... always do `git pull --rebase` -- can explain more next time we sync.

PR #581, opened 2021-01-26
Deactivated flag moved to DID Document Metadata

csuwildcat:
***
<a href="https://pr-preview.s3.amazonaws.com/csuwildcat/did-core/pull/581.html" title="Last updated on Feb 4, 2021, 2:18 PM UTC (355b2d7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/581/c8bc6b1...csuwildcat:355b2d7.html" title="Last updated on Feb 4, 2021, 2:18 PM UTC (355b2d7)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -3818,13 +3818,6 @@ This error code is returned if the representation requested via the
 <code>accept</code> input metadata property is not supported by the DID method
 and/or DID resolver implementation.
 </dd>
-                        <dt>
-deactivated
-                        </dt>
-                        <dd>
-The <a>DID</a> supplied to the <a>DID resolution</a> function has been
-deactivated. (See <a href="#deactivate"></a>.)
-                        </dd>
                     </dl>
                 </dd>
             </dl>
@@ -3859,6 +3852,15 @@ DID document metadata SHOULD include an <code>updated</code> property to
 indicate the timestamp of the last <a href="#update">Update operation</a> for
 the document version which was resolved. The value of the property MUST follow
 the same formatting rules as the <code>created</code> property.
+                </p>
+              </section>
+
+          <section>
+            <h4>deactivated</h4>
+            <p>
+This property MUST be populated with a <a data-cite="INFRA#boolean">boolean</a> 
+`true` value if the DID Method determines that the <a>DID</a> supplied has been 
+deactivated. (See <a href="#deactivate"></a>.)
                 </p>
               </section>
 
@@ -4182,13 +4184,6 @@ not-found
 The <a>DID URL dereferencer</a> was unable to find the
 <code>content-stream</code> resulting from this dereferencing request.
                         </dd>
-                        <dt>
-deactivated
-                        </dt>
-                        <dd>
-The <a>DID</a> in the <a>DID URL</a> supplied to the <a>DID URL
-dereferencing</a> function has been deactivated. (See
-<a href="#deactivate"></a>.)
                     </dl>
                 </dd>
             </dl>
--- a/index.html
+++ b/index.html
@@ -3858,7 +3858,8 @@ the same formatting rules as the <code>created</code> property.
           <section>
             <h4>deactivated</h4>
             <p>
-This property MUST be populated with a <a data-cite="INFRA#boolean">boolean</a> 
+This property MUST be populated by a <a data-cite="INFRA#boolean">boolean</a> 
+value that is true if the DID has been deactivated, and false if the DID is still active.
 `true` value if the DID Method determines that the <a>DID</a> supplied has been 
 deactivated. (See <a href="#deactivate"></a>.)
                 </p>


index.html
```diff
@@ -3862,6 +3855,15 @@ <h4>updated</h4>
                 </p>
               </section>
 
+          <section>
+            <h4>deactivated</h4>
+            <p>
+This property MUST be populated with a <a data-cite="INFRA#boolean">boolean</a>
```
msporny, line 3861:
```suggestion
This property is populated with a <a data-cite="INFRA#boolean">boolean</a> 
```

msporny (changes requested):
Minor nit, don't think we need to make this statement testable.

index.html
```diff
@@ -3862,6 +3855,15 @@ <h4>updated</h4>
                 </p>
               </section>
 
+          <section>
+            <h4>deactivated</h4>
+            <p>
+This property MUST be populated with a <a data-cite="INFRA#boolean">boolean</a> 
+`true` value if the DID Method determines that the <a>DID</a> supplied has been
```
peacekeeper, line 3862:
```suggestion
`true` value if the supplied <a>DID</a> has been 
```

I'd prefer to simplify this sentence a bit, but I'd also be fine with the PR as-is.

csuwildcat:
@msporny don't you think we should state that the DID Method MUST populate this with a true value if the DID has been deactivated? Given we tell them every method must take care to sort out their deactivated story, how then is it not logical to tell the constructor of the resolution response to mark this as true if the method resolution determines it is deactivated?

csuwildcat:
I guess I feel like we're taking the whole testability thing a tad too far, and these kinds of clear, logical directives that naturally follow from the other directives to DID Methods elsewhere in the spec are being watered down as a result.

msporny:
> @msporny don't you think we should state that the DID Method MUST populate this with a true value if the DID has been deactivated? Given we tell them every method must take care to sort out their deactivated story, how then is it not logical to tell the constructor of the resolution response to mark this as true if the method resolution determines it is deactivated?

That would be a DID Method-specific test... how do we test this without driving a specific DID Method implementation to register a DID and then deactivate the DID and then use a resolver to check that the DID is deactivated? That's the only way that I know of to test the statement. Our charter also lists DID Method specifications as out of scope: 

https://www.w3.org/2019/09/did-wg-charter

So, if we were to start testing DID Method implementations, a W3C Member could assert that we're violating our charter.

> I guess I feel like we're taking the whole testability thing a tad too far, and these kinds of clear, logical directives that naturally follow from the other directives to DID Methods elsewhere in the spec are being watered down as a result.

It it's a clear logical directive that flows naturally, it doesn't need to be stated at all. :)

Another way to put this is -- the intent of the wording is clear without the normative requirement. What are we afraid of... that implementations will set deactivated to 'false' when documents have been deactivated? I'm pretty sure no one would use that DID Method or resolver if that were the case.

msporny:
Waiting on feedback from @csuwildcat on response to his questions. Waiting on others from the WG to break the normative language tie.

index.html
jricher: I prefer to keep this normative instead of declarative.

msporny: If we keep this normative, how do we test this without wiring up a DID Method, creating a DID, deactivating it, and then querying the DID Method through the resolver interface to check this property? In other words, that sounds like a lot of effort for the WG to check a feature that we are not expecting people to mis-implement (and which should probably be in the DID Resolution test suite), no?

OR13: The same way we test numbers and booleans today :)

https://did.actor/mike/did.json

OR13: https://github.com/w3c/did-test-suite/pull/20/files#diff-7828e42130aa848d05937900dd5476d3978c431142ee878fe4388b87b92d129fR18

csuwildcat:
I'll do whatever folks want here, I just thought it was important to make this a directive to get compliance from Method authors.

index.html
msporny: It's not the "check to make sure it's a boolean" part of the normative state that I'm concerned about. It's the "if the DID Method determines that the DID supplied has been deactivated." part of the normative statement that's an issue. The test would require you to query a live system where a DID isn't deactivated... then deactivate it... then test to see if the value you get back would be set to "deactivated===true".

OR13: The entire class of tests associated with sequential did operations should be considered out of scope... instead we should provide examples of resolution results for after: "created", "updated", "deactivated"... this is how we establish conformance in sidetree across languages and implementations... even @csuwildcat has agreed to those tests :)

peacekeeper:
Can we make it normative in DID Core and then test it elsewhere, e.g. in the CCG where we ARE allowed to work on DID resolution and concrete DID methods?

msporny:
> Can we make it normative in DID Core and then test it elsewhere, e.g. in the CCG where we ARE allowed to work on DID resolution and concrete DID methods?

We are not supposed to delay testing or put the burden of testing on an external group. If we want a normative statement that can be tested by a machine (which is ideally all the normative statements in the specification), we are expected to write a test for it. The normative language in the specification is expected to be self-contained (or rely on other existing standards that have an official standing).

index.html
msporny: >  instead we should provide examples of resolution results for after: "created", "updated", "deactivated"...

Then the normative language would need to change to not include the decision process of a DID Method. We could reword this statement to:

> This property MUST be populated by a boolean value that is expected to be `true` if the DID has been deactivated, and `false` if the DID is still active.

... but that's just saying, the value MUST be a boolean value. It allows us to write two tests -- one where it's true and another where it's false. I question the value of those tests, but if folks really want to have normative language around the range of values and when they're expected to be those values, the language above would do that.

OR13:
Related https://github.com/decentralized-identity/sidetree/issues/1015#issuecomment-771998522

peacekeeper:
> Does this look right?

Looks great to me!

[Off-topic] Also, perhaps we should standardize and register the "method" metadata property so that any method can use it for putting their own method-specific things in there.

index.html
msporny: The WG proposed the following text in a special topic call, there were no objections: This property MUST be populated by a boolean value that is true if the DID has been deactivated, and false if the DID is still active.

I'm going to hold the merge until we hear from @csuwildcat to make sure he's ok w/ the language.

@csuwildcat, are you ok with the proposed language above?

index.html
msporny: The WG proposed the following text in a special topic call, there were no objections: This property MUST be populated by a boolean value that is true if the DID has been deactivated, and false if the DID is still active.

Since you were on that call, I'm going to go with that language.

index.html
peacekeeper: I agree with this language. Unless @csuwildcat has any objections, we should accept this suggestion and then merge the PR.

csuwildcat:
Sounds good to me

peacekeeper (approved):
+1 to merging after applying the suggestion https://github.com/w3c/did-core/pull/581/commits/74f2d40e276e023b44942ab965ba7765bff7d8cf#r564900912.

msporny (approved):

msporny:

Issue #549, opened 2021-01-17
Concerns regarding testability of DID Resolution and Dereferencing


msporny:
A few items of concern:

* We currently have 153 normative MUST statements in the specification. That's a lot of tests that the group will have to write over the next month or two. 
* 43 of those tests (or roughly 1/3rd of the normative statements in the specification) come from the Resolution and Dereferencing section.
* We have no conformance class for DID resolution.
* We say the following in the Resolution section: "This section defines the inputs and outputs of DID resolution and DID URL dereferencing. These functions are defined in an abstract way. Their exact implementation is out of scope for this specification, but some considerations for implementors are discussed in [DID-RESOLUTION]."
* We have these resolutions:
  * Resolution #1 on 2020-04-21: It is in scope for the DID WG to normatively define the parameters of a concrete set of processes that take a DID as input and provides a DID Document as output.
  * Resolution #2 on 2020-04-21: It is in scope to make these processes take in options and provide back a document along with different classes of metadata (e.g., document metadata and resolution metadata).
  * Resolution #3 on 2020-04-21: It is in scope to add tests to the test suite that exercise these processes to test the inputs (DIDs, DID URLs, and input metadata) as well as the outputs (DID Documents, resources, and output metadata).
  * Resolution #4 on 2020-04-21: It is out of scope to normatively define how any specific DID method implements method-specific details of DID resolution.
  * Resolution #5 on 2020-04-21: it is out of scope to define whether protocols are used for DID Method implementations or to define those protocols.
  * Resolution #6 on 2020-04-21: It is out of scope to test concrete DID Resolution protocols and data formats beyond the necessary process to demonstrate interoperability between the test suite and an implementation.

At this point, we need one or more people to step forward to write the tests for the DID Resolution and Dereferencing section and we need at least two companies that will commit to implement DID resolvers that will pass the tests. If we can't get that sort of commitment (I don't even know how we'd write the tests with our current test suite -- maybe @OR13 has some suggestions), we should mark the section at risk and say that the Resolution and Dereferencing section may be moved into a separate NOTE published by the group.

My recollection is that the group desired to know how the interfaces for resolution and dereferencing would work. I believe the group has achieved that with the current specification and we could safely move Resolution and Dereferencing into a NOTE for publication. This would set the stage for a future DID Resolution WG without having any substantive negative impact on the understanding of how Resolution and Dereferencing are meant to work. Publication as a NOTE would have the added benefit of relieving a good chunk of CR implementation pressure.

[...]

PR #587, opened 2021-02-01
Make DID Resolution section concrete and testable (WebIDL).

msporny:
This PR is intended to address issue #549 by converting the abstract functions discussed in the DID Resolution and DID URL Dereferencing sections into concrete, testable functions. The PR uses a very light amount of WebIDL while keeping much of what we had the same as before. It also tries to strike a balance by enabling at least one concrete realization of the interface without preventing other concrete interfaces from being realized.

There are a LOT of reformatting changes in this PR, it might be best to <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/587.html#resolution">look at the rendered preview for this PR</a>.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/587.html" title="Last updated on Feb 2, 2021, 3:51 PM UTC (51dc50b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/587/518362a...51dc50b.html" title="Last updated on Feb 2, 2021, 3:51 PM UTC (51dc50b)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -3710,8 +3710,8 @@ The <a>DID resolution</a> function resolves a <a>DID</a> into a <a>DID
 document</a> by using the "Read" operation of the applicable <a>DID method</a>
 as described in <a href="#read-verify"></a>. The details of how this process is
 accomplished are outside the scope of this specification, but all conforming
-DID Resolver implementations realize at least the following interface (or one
-that is functionally equivalent):
+<a>DID resolver</a> implementations realize at least the following interface
+(or one that is functionally equivalent):
         </p>
 
         <pre class="idl">
@@ -3720,11 +3720,6 @@ interface mixin DidResolver {
 };
         </pre>
 
-        <p>
-The {{DidResolver/resolve}} function returns a byte stream of the
-<a>DID Document</a> formatted in the corresponding representation,
-DID Resolution Metadata, and DID Document Metadata.
-        </p>
         <p>
 The input variables of the {{DidResolver/resolve}} function are as follows:
         </p>
@@ -3780,14 +3775,12 @@ an <code>error</code> property describing the error.
 didDocument
             </dt>
             <dd>
-If the resolution is successful, and if the <code>resolveRepresentation</code>
-function was called, this MUST be a byte stream of the resolved <a>DID
-document</a> in one of the conformant
+If the resolution is successful this MUST be a byte stream of the resolved
+<a>DID document</a> in one of the conformant
 <a href="#representations">representations</a>. The byte stream might then be
-parsed by the caller of the <code>resolveRepresentation</code> function into a
-<a href="#data-model">data model</a>, which can in turn be validated and
-processed. If the resolution is unsuccessful, this value MUST be an empty
-stream.
+parsed by the caller into a <a href="#data-model">data model</a>, which can in
+turn be validated and processed. If the resolution is unsuccessful, this value
+MUST be an empty stream.
             </dd>
             <dt>
 didDocumentMetadata
@@ -3796,13 +3789,13 @@ didDocumentMetadata
 If the resolution is successful, this MUST be a <a
 href="#metadata-structure">metadata structure</a>. This structure contains
 metadata about the <a>DID document</a> contained in the
-<code>didDocument</code>. This metadata
+<code>didDocument</code> property. This metadata
 typically does not change between invocations of the {{DidResolver/resolve}}
 function unless the <a>DID document</a> changes, as it represents data about
 the <a>DID document</a>. If the resolution is unsuccessful, this output MUST
 be an empty <a href="#metadata-structure">metadata structure</a>. Properties
 defined by this specification are in <a
-href="#did-document-metadata-properties"></a>.
+href="#did-document-metadata"></a>.
             </dd>
         </dl>
 
@@ -3842,7 +3835,7 @@ available. This property is OPTIONAL.
         </section>
 
         <section>
-            <h3>DID Resolution Metadata Properties</h3>
+            <h3>DID Resolution Metadata</h3>
 
             <p>
 The possible properties within this structure and their possible values are
@@ -3910,7 +3903,7 @@ deactivated. (See <a href="#deactivate"></a>.)
         </section>
 
         <section>
-          <h3>DID Document Metadata Properties</h3>
+          <h3>DID Document Metadata</h3>
 
           <p>
 The possible properties within this structure and their possible values SHOULD
@@ -3918,21 +3911,19 @@ be registered in the DID Specification Registries [[?DID-SPEC-REGISTRIES]].
 This specification defines the following common properties.
           </p>
 
-          <section>
-            <h4>created</h4>
-            <p>
+          <dl>
+            <dt><dfn>created</dfn></dt>
+            <dd>
 <a>DID document</a> metadata SHOULD include a <code>created</code> property to
 indicate the timestamp of the <a href="#create">Create operation</a>. The value
 of the property MUST be a <a data-cite="INFRA#string">string</a> formatted as an
 <a data-cite="XMLSCHEMA11-2#dateTime">XML Datetime</a> normalized to UTC
 00:00:00 and without sub-second decimal precision. For example:
 <code>2020-12-20T19:17:47Z</code>.
-            </p>
-          </section>
+            </dd>
 
-          <section>
-            <h4>updated</h4>
-            <p>
+            <dt><dfn>updated</dfn></dt>
+            <dd>
 <a>DID document</a> metadata SHOULD include an <code>updated</code> property to
 indicate the timestamp of the last <a href="#update">Update operation</a> for
 the document version which was resolved. The value of the property MUST follow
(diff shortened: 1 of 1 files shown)


peacekeeper:
The idea behind the abstract functions was to make it crystal clear that DID resolution is not a single protocol (as in DNS), but that invocation of the resolution function can come in many forms (command line, SDK call, HTTP API, DIDComm, ...). I think your PR still allows this by saying _"DID resolver implementations realize at least the following interface (or one that is functionally equivalent)"_.

However I don't really understand how adding one concrete form (WebIDL) makes the section more testable? I thought WebIDL is intended for browser APIs, is anyone working on DID resolution in browsers?

The biggest change seems to be that it removes `resolveRepresentation` (and at the same time changes `resolve` to be like `resolveRepresentation`). I don't think this is a good idea, we had many meetings/issues/PRs that led to this design. DID resolution can happen without ever dealing with any concrete representation.

Some smaller comments
- I like the reformatting efforts, including camelCase and cleaning up section headings and property names
- I could live with the rename from "did-resolution-input-metadata" to "resolutionOptions" is okay, but would slightly prefer to continue using the term "metadata" also for inputs, to point out that input metadata and output metadata follow the same data model (similar to HTTP request/response headers).

iherman (changes requested):
We should **_NOT_** use WebIDL. We would get into a long, pointless, and religious debate on what WebIDL is for, whether it is a formalism that can be used outside of browsers or not, whether the inclusion of WebIDL in the standard means that browsers are required to implement it or not.

The JSON-LD 1.1 Recommendation vote was delayed exactly on those grounds: there were formal objections from some AC members to publish the Recommendation. We did find some sort of compromise WebIDL syntax solution at the end, but it was not pleasant. We would incur the danger of finding ourselves in the same discussions. It is not worth it.

OR13:
> The biggest change seems to be that it removes `resolveRepresentation` (and at the same time changes `resolve` to be like `resolveRepresentation`). I don't think this is a good idea, we had many meetings/issues/PRs that led to this design. DID resolution can happen without ever dealing with any concrete representation.

agree with @peacekeeper ... there was specific language about changing the function signature... however, I think its better to have 1 concrete function signature than 2 abstract ones... so I am supportive of the change.

I have always been confused by how resolve is different from resolve representation, here is how we implemented it:

https://github.com/transmute-industries/did-key.js/blob/master/packages/did-key-common/src/getResolver.ts#L90

We build a resolver of did key, from its Key Pair class.. ad we only support "resolveRepresentation" but we call it "resolve".

msporny:
> The JSON-LD 1.1 Recommendation vote was delayed exactly on those grounds: there were formal objections from some AC members to publish the Recommendation. 

I'll note that the JSON-LD API specification /did/ end up using WebIDL: https://www.w3.org/TR/json-ld-api/#the-jsonldprocessor-interface

I can remove the letters "WebIDL"... but the result is going to effectively be WebIDL. If folks feel like removing those letters from the specification is going to make CR/PR easier, then that's fine, I can do it. It is a bit ridiculous of a few AC Members to object -- on what grounds did they object?

Keep in mind that I tried to say: "This is one way to do it, it's not the only way" to prevent these sorts of objections. We're effectively saying... "if you're implementing in Javascript, here's a concrete way to do it... but it's not the only way, you can do this same sort of thing as an HTTP API, a gRPC API, or anything else."

msporny:
> I think your PR still allows this by saying _"DID resolver implementations realize at least the following interface (or one that is functionally equivalent)"_.

Yes, exactly... that's the intent. "This is just one concrete way to realize resolution/dereferencing, but there are other ways and those are fine too as long as they're functionally equivalent."

> However I don't really understand how adding one concrete form (WebIDL) makes the section more testable? I thought WebIDL is intended for browser APIs, is anyone implementing a DID resolution browser API?

WebIDL is meant for Javascript APIs, and yes, there are multiple companies implementing resolvers using Javascript. It doesn't mean that's the /only/ way you can have a conformant DID resolver implementation... it just says that that is /one/ path to conformance that's testable.

> The biggest change seems to be that it removes `resolveRepresentation` (and at the same time changes `resolve` to be like `resolveRepresentation`). I don't think this is a good idea, we had many meetings/issues/PRs that led to this design. DID resolution can happen without ever dealing with any concrete representation.

If there is no concrete representation, there is nothing to test. That's the point -- we can't deal in the abstract when we go to write the test suite tests.

iherman:
> 
> I can remove the letters "WebIDL"... but the result is going to effectively be WebIDL. If folks feel like removing those letters from the specification is going to make CR/PR easier, then that's fine, I can do it. It is a bit ridiculous of a few AC Members to object -- on what grounds did they object?

That, by using WebIDL, we say that the feature must (also) be implemented in a browser. Do we want to get to that discussion?

> 
> Keep in mind that I tried to say: "This is one way to do it, it's not the only way" to prevent these sorts of objections. We're effectively saying... "if you're implementing in Javascript, here's a concrete way to do it... but it's not the only way, you can do this same sort of thing as an HTTP API, a gRPC API, or anything else."

Believe me, I understand. One way to do this, maybe, is to move the WebIDL part into an (explicitly non-normative!) appendix in the spec to avoid unnecessary time spent on this later.

peacekeeper:
All the sources I find about WebIDL say that it's for implementation in web browsers, not for generic JavaScript APIs.

And even if it was, I don't understand how this makes the section more testable? What about non-JavaScript implementations of DID Resolution, how would they benefit from the switch to WebIDL, and how would they be tested?

msporny:
@iherman wrote:
> That, by using WebIDL, we say that the feature must (also) be implemented in a browser. Do we want to get to that discussion?

... but that objection is ridiculous! :P That's not only what WebIDL is for... sure, we want to avoid ridiculous objections... but the JSON-LD API is an example that those ridiculous objections are often stayed and more sensible folks are allowed to proceed.

@peacekeeper wrote:
> All the sources I find about WebIDL say that it's for implementation in web browsers, not for generic JavaScript APIs.

The specification for WebIDL defines what WebIDL is for -- it's for ECMAScript (aka: JavaScript):

https://heycam.github.io/webidl/#introduction

> And even if it was, I don't understand how this makes the section more testable? 

Because at least a JavaScript developer can take a WebIDL interface and write code against it. What we had in the specification before could be misinterpreted in a variety of different ways. Can I use a dictionary? A map? A Map (which is not a map)? Can I expose this over an HTTP interface? The answer to all those questions is "yes" -- and that's a good indication that the thing isn't easily testable and if we were to write tests, people would disagree with the tests because they interpreted the language incorrectly.

> What about non-JavaScript implementations of DID Resolution, how would they benefit from the switch to WebIDL, and how would they be tested?

I have no idea how we'd test those things... and that's the problem... it was not clear how any of this would be tested because the interface that we'd have to use was purely abstract. It's like asking people to test the DID specification using purely the Abstract Data Model without resorting to using a representation. You can't test things that are abstract because they are... abstract! :yum:

msporny:
@iherman and @peacekeeper -- trying to figure out how to change this PR so it would address each of your concerns. If I were to remove the "WebIDL" text and replace it with "an example function signature in JavaScript" as a non-normative example... would that address both of your concerns?

jricher (changes requested):
This is not a simple swap out of one description language for another.

This change set removes the `resolveRepresentation` function and collapsed a number of previously separated bits of functionality into a single catch all-function, effectively undoing months of work and consensus within the working group.

This PR should be rejected.

msporny:
@jricher wrote:
> This change set removes the resolveRepresentation function and collapsed a number of previously separated bits of functionality into a single catch all-function, effectively undoing months of work and consensus within the working group.

It only thing that was removed was the purely abstract part that was not testable. It would be fairly trivial to add that one abstract function back in if that would remove your objection. The core of the spec changes for this PR took about an hour to make, which demonstrates that there really weren't a ton of functional changes (most of the work was editorial reformatting changes -- renaming variables, changing deeply nested sections into definition lists that match the rest of the specification, etc).

The biggest change was renaming resolveRepresentation -> resolve and the WebIDL rework. What other changes do you think happened?

OR13 (approved):

OR13:
I like the web idl definitions, and the fact that the awkward double resolution definition is gone, I think this PR is a step in the right direction.

I could live without the WebIDL and instead see the function defined in INFRA, for the sake of consistency that might be a better path to take.

iherman:
> @iherman and @peacekeeper -- trying to figure out how to change this PR so it would address each of your concerns. If I were to remove the "WebIDL" text and replace it with "an example function signature in JavaScript" as a non-normative example... would that address both of your concerns?

Yes, that could work. Actually, I would probably prefer TypeScript to JavaScript, because we can then provide type annotations/definitions in the interface. Typescript is close enough to JavaScript that vanilla JavaScript users would also understand it.

peacekeeper:
> > And even if it was, I don't understand how this makes the section more testable?
> 
> Because at least a JavaScript developer can take a WebIDL interface and write code against it. What we had in the specification before could be misinterpreted in a variety of different ways. Can I use a dictionary? A map?

@msporny I still don't get it. The functions currently specify inputs and outputs. Those inputs and outputs are well-defined using INFRA. I'm perfectly able to write code that takes the inputs and returns the outputs of those functions, and I don't really need any extra spec text for that.

Is there an "official INFRA way" to define functions? That might be ideal.

> If I were to remove the "WebIDL" text and replace it with "an example function signature in JavaScript" as a non-normative example... would that address both of your concerns?

I'd be fine with leaving the functions as-is and adding a JavaScript/TypeScript example. In this case I would suggest also adding the [HTTP(S) binding](https://w3c-ccg.github.io/did-resolution/#bindings-https) from the DID Resolution spec as a second concrete example of how those functions can be implemented, even if that can only be a non-normative reference of course.

peacekeeper:
> This change set removes the `resolveRepresentation` function and collapsed a number of previously separated bits of functionality into a single catch all-function, effectively undoing months of work and consensus within the working group.

I agree with @jricher. Personally, I would also prefer a single function rather than two. I raised several PRs myself which tried to do it with a single function, but they were all rejected, and the WG then arrived at the current design.

For further information, see the following 13 (!) PRs related to this topic with >200 comments and links to multiple special topic calls:

https://github.com/w3c/did-core/pulls?q=is%3Apr+label%3Acontract+

msporny:
@peacekeeper @iherman, I have put forward an alternate PR that removes all WebIDL and uses a Typescript interface definition as an example, does that work for both of you? You can review the changes in PR #591.

peacekeeper:
> "Now that we know that removing the abstract resolve() function doesn't impact implementations, do we still want to keep it, if so, why?"

I don't think we know that. There are implementations that implement resolve(), i.e. they take a DID as input and return a DID document data model that is not in any concrete representation.

OR13:
> > "Now that we know that removing the abstract resolve() function doesn't impact implementations, do we still want to keep it, if so, why?"
> 
> I don't think we know that. There are implementations that implement `resolve()`, i.e. they take a DID as input and return a DID document data model that is not in any concrete representation.

ahh... that is not a thing we do anywhere... but maybe there are folks that are returning Class or Objects that represent a DID Document in the ADM... in which case this would impact those...

index.html
```diff
@@ -3703,35 +3709,25 @@ <h1>Resolution</h1>
...
-        <p><code>
-resolve ( did, did-resolution-input-metadata ) <br>
-&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;-&gt; ( did-resolution-metadata, did-document, did-document-metadata )
-        </code></p>
-
-        <p><code>
-resolveRepresentation ( did, did-resolution-input-metadata ) <br>
-&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;-&gt; ( did-resolution-metadata, did-document-stream, did-document-metadata )
-        </code></p>
+        <pre class="idl">
```
OR13, line 3723: sad web idl

msporny:
Closing in favor of PR #592.

PR #591, opened 2021-02-02
Make DID Resolution section concrete and testable (Typescript). 

msporny:
This PR is intended to address issue #549 by converting the abstract functions discussed in the DID Resolution and DID URL Dereferencing sections into concrete, testable functions. The PR uses an Typescript interface to provide an example while keeping much of what we had the same as before. It also tries to strike a balance by providing at least one concrete realization of the interface without preventing other concrete interfaces from being realized.

There are a LOT of reformatting changes in this PR, it might be best to <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/591.html#resolution">look at the rendered preview for this PR</a>.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/591.html" title="Last updated on Feb 2, 2021, 9:27 PM UTC (f8a8d97)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/591/518362a...f8a8d97.html" title="Last updated on Feb 2, 2021, 9:27 PM UTC (f8a8d97)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -3710,8 +3710,8 @@ The <a>DID resolution</a> function resolves a <a>DID</a> into a <a>DID
 document</a> by using the "Read" operation of the applicable <a>DID method</a>
 as described in <a href="#read-verify"></a>. The details of how this process is
 accomplished are outside the scope of this specification, but all conforming
-DID Resolver implementations realize at least the following interface (or one
-that is functionally equivalent):
+<a>DID resolver</a> implementations realize at least the following interface
+(or one that is functionally equivalent):
         </p>
 
         <pre class="idl">
@@ -3720,11 +3720,6 @@ interface mixin DidResolver {
 };
         </pre>
 
-        <p>
-The {{DidResolver/resolve}} function returns a byte stream of the
-<a>DID Document</a> formatted in the corresponding representation,
-DID Resolution Metadata, and DID Document Metadata.
-        </p>
         <p>
 The input variables of the {{DidResolver/resolve}} function are as follows:
         </p>
@@ -3780,14 +3775,12 @@ an <code>error</code> property describing the error.
 didDocument
             </dt>
             <dd>
-If the resolution is successful, and if the <code>resolveRepresentation</code>
-function was called, this MUST be a byte stream of the resolved <a>DID
-document</a> in one of the conformant
+If the resolution is successful this MUST be a byte stream of the resolved
+<a>DID document</a> in one of the conformant
 <a href="#representations">representations</a>. The byte stream might then be
-parsed by the caller of the <code>resolveRepresentation</code> function into a
-<a href="#data-model">data model</a>, which can in turn be validated and
-processed. If the resolution is unsuccessful, this value MUST be an empty
-stream.
+parsed by the caller into a <a href="#data-model">data model</a>, which can in
+turn be validated and processed. If the resolution is unsuccessful, this value
+MUST be an empty stream.
             </dd>
             <dt>
 didDocumentMetadata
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -3957,270 +3933,222 @@ <h4>updated</h4>
...
-      <p><code>
-dereference ( did-url, did-url-dereferencing-input-metadata ) <br>
-&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;-&gt; ( did-url-dereferencing-metadata, content-stream, content-metadata )
-      </code></p>
+      <pre class="example" title="Conforming dereferencer interface in Typescript">
+interface DidDereferencer {
+  dereference(USVString did,
+              DereferencingOptions dereferencingOptions): DereferencingResponse;
+};
+      </pre>
```
iherman, line 4084: That does not seem like a correct Typescript. Should it be

```
did: UVString, dereferencingOptions: DereferencingOptions
```

iherman (changes requested):

peacekeeper (changes requested):
This ignores the feedback given in the other PR by @jricher and myself (see https://github.com/w3c/did-core/pull/587#issuecomment-770701841 and https://github.com/w3c/did-core/pull/587#pullrequestreview-580758550).

Adding JavaScript/TypeScript is a good idea, but only as an example, not to replace the abstract function definitions. Also, the resolve() and resolveRepresentation() functions should both be kept.

OR13:
I feel that "resolve() and resolveRepresentation()" should NOT both be kept, especially if we can get consensus on "resolveRepresentation" and its interface definition, since it is a superset.

index.html
```diff
@@ -3703,35 +3709,26 @@ <h1>Resolution</h1>
...
-        <p><code>
-resolve ( did, did-resolution-input-metadata ) <br>
-&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;-&gt; ( did-resolution-metadata, did-document, did-document-metadata )
-        </code></p>
-
-        <p><code>
-resolveRepresentation ( did, did-resolution-input-metadata ) <br>
-&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;-&gt; ( did-resolution-metadata, did-document-stream, did-document-metadata )
-        </code></p>
+        <pre class="example" title="A conforming resolver interface in Typescript ">
```
OR13, line 3723: As @peacekeeper mentioned in his review, I think we should keep the abstract function definition in infra AND keep the typescript example provided here.

peacekeeper:
> since it is a superset.

No it's not a superset. It is possible to implement `resolve()` in a way that internally calls `resolveRepresentation()`, and it's also possible to implement `resolveRepresentation()` in a way that internally calls `resolve()`.

The long discussion process in the WG has shown that there are use cases for both functions.

OR13 (changes requested):
We should keep a single abstract function definition using infra for `resolve` which should map to what we have previously called `resolveRepresentation`.

OR13:
I think practically the abstract definition for `resolve` is useless and confusing, whereas `resolveRepresentation` is useful, and should be kept, after being renamed to `resolve`.

peacekeeper:
@OR13 I understand your point of view if you are working mostly in a JavaScript environment, where the native data model = JSON, which is at the same time also a representation. In this environment there's no real difference between the abstract data model and the representation, therefore you also don't see a need for two separate functions.

But once you start thinking outside of a JavaScript environment, in a generic and implementation-agnostic way (which is what the spec should do), then there's a big difference between a function that returns a data model, and a function that returns a concrete representation.

OR13:
@peacekeeper the abstract function definition is what prevents us from having a language / representation specific view of resolution, I am supportive of preserving that definition in INFRA.

I am also supportive of providing helpful examples.

In summary I agree we should preserve the abstract definitions, and we should provide concrete examples, in as many representations / languages as we can.

index.html
msporny: You are correct, fixed in latest push.

jricher (changes requested):
This PR removes functionality agreed upon by the WG after many hours of discussion. 

This also ignores feedback provided in #587 to the same.

This PR should be rejected.

msporny:
Closing in favor of PR #592.

PR #597, opened 2021-02-03
Add diagram "resolve() and resolveRepresentation()".

peacekeeper:
For the "DID Resolution" section:

_Diagram illustrating how resolve() returns the DID document data model in its abstract form and resolveRepresenation() returns it in one of the conformant representations; conversion is possible using production and consumption rules._

diagram-resolve-resolverepresentation

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/597.html" title="Last updated on Mar 1, 2021, 8:25 PM UTC (2ee68b1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/597/4f4a7b1...2ee68b1.html" title="Last updated on Mar 1, 2021, 8:25 PM UTC (2ee68b1)">Diff</a>


iherman:
I wonder whether it is possible to create an SVG version and use that in the document (with a PNG fallback). Also, the ECHIDNA file must be updated, too, otherwise the automatic publication will not work!

(Obviously, this comment is also relevant for #595 and #596, too.)

peacekeeper:
> I wonder whether it is possible to create an SVG version

I made them in LibreOffice Draw/Impress; there's an SVG export function, I will try that!

iherman:
> > I wonder whether it is possible to create an SVG version
> 
> I made them in LibreOffice Draw/Impress; there's an SVG export function, I will try that!

At least on the LibreOffice on the Mac there is an export menu item listing all kinds of formats, including SVG. I had a cursory look at the generated SVG, and it looks fine to me.

TallTed:
As with https://github.com/w3c/did-core/pull/595#issuecomment-772845778...

Needs much more detailed `longdesc` description

shigeya:
Looks good. One note, it's confusing to add parenthesis to `consume()` and `produce()` unless there is a discussion about these *concrete* functions.

OR13 (changes requested):
Same comment as https://github.com/w3c/did-core/pull/596

I like the diagram, but not the visualization of the ADM. 

You don't need to actually show any part of the ADM, its internal representation is 100% not relevant to did core... all we need to know is how to produce and consume concrete representations... representing the ADM as JSON is misleading.

See 

Note that it says the DID Document is an INFRA map.... whether `@context` shows up in YAML / CBOR is up the did method... but its clearly an entry in the infra map for the purposes of JSON-LD production.

Perhaps we can use the diagram above (that makes the distinction between property types, instead of the one included here).

peacekeeper:
> representing the ADM as JSON is misleading.

@OR13 the diagram doesn't show the ADM as JSON. It shows the ADM in INFRA notation, see here: https://infra.spec.whatwg.org/

Since we use INFRA in the spec to define the data model, I thought it's appropriate to show INFRA notation also here in the diagram.

Does this clarify the diagram, or do you still have concerns?

peacekeeper:
I just pushed a small fix that addresses a JSON-LD bug, and also avoids the double-directional arrow as suggested by @shigeya .

diagram-resolve-resolverepresentation

OR13:
@peacekeeper the diagram is not accurate, in order produce JSON-LD from the ADM, we must follow the JSON-LD production rules.... which require `@context` in the ADM DID Document infra map.

You cannot produce JSON-LD without holding some representation of the `@context` in the ADM... it might be a buffer, or a file, or strings, or whatever... but before you produce JSON-LD you need to account for that.... your picture above, is perpetuating our disagreement, and disregarding the consensus we had regarding this on the multiple special topic calls on the subject....

See 


Note that `@context` is present in "Entries in the DID Document INFRA map"...  this is correct.... whether this value shows up in other representations is not the question... when representing JSON-LD, you need to include this value. I will not approve the PR until the picture is updated to reflect: 

See

jricher:
@OR13 You don't need to have a `context` value in the ADM to produce JSON-LD, though. If my software can map all properties in the ADM using contexts that it knows about, then it can do so without storing the context as part of the ADM map. It's allowed to store it there, but there's no restriction that it has to (nor should there be such a restriction).

iherman:
> @OR13 You don't need to have a `context` value in the ADM to produce JSON-LD, though. If my software can map all properties in the ADM using contexts that it knows about, then it can do so without storing the context as part of the ADM map. It's allowed to store it there, but there's no restriction that it has to (nor should there be such a restriction).

+1 

It is a bit problematic that the same problem is discussed at two places. I said the same thing in https://github.com/w3c/did-core/pull/596#issuecomment-775283616 although, as usual, I am unnecessarily more verbose:-)

OR13:
> @OR13 You don't need to have a `context` value in the ADM to produce JSON-LD, though. If my software can map all properties in the ADM using contexts that it knows about, then it can do so without storing the context as part of the ADM map. It's allowed to store it there, but there's no restriction that it has to (nor should there be such a restriction).

@jricher I believe your perspective is not reflected in the current spec language.... see:

https://w3c.github.io/did-core/#production-0

> The DID document MUST be serialized according to the production rules for JSON, with one additional requirement: The DID document MUST include the @context entry.

The linked term for "DID Document" leads to: https://w3c.github.io/did-core/#dfn-did-documents

> A set of data describing the DID subject, including mechanisms, such as public keys and pseudonymous biometrics, that the DID subject or a DID delegate can use to authenticate itself and prove its association with the DID. A DID document might also contain other attributes or claims describing the DID subject. A DID document might have one or more different representations as defined in § 6. Representations or in the W3C DID Specification Registries [DID-SPEC-REGISTRIES].

I assume when we say "DID Document" we mean an INFRA Map, such as what is described in:

See 

Hence the picture is not accurate, and actually violates the JSON-LD Production rules.

As I mentioned elsewhere... there are 3 solutions to this;

1. Split the production consumption diagrams, and include `@context` in the ADM for JSON-LD Production, but not for the other ones.
2. Include `@context` in the ADM, (and keep it excluded from `did+json` / `did+cbor`.
3. Remove the erroneous description of JSON-LD production from the diagram.

jricher:
The production language for JSON-LD is sloppy. In the second space it should say "the representation of the DID Document" as it's clearly talking about the result of the production not the source.

OR13:
> The production language for JSON-LD is sloppy. In the second space it should say "the representation of the DID Document" as it's clearly talking about the result of the production not the source.

I don't agree.... production is defined generally as conversion from an INFRA map, to a concrete representation such as `application/did+json`... in the case of vanilla JSON... we have consensus that all properties are preserved and not dropped... We then further clarified that additional properties may be sanitized and the discretion of the method implementer (allowing for defense against prototype pollution for example)....

Its important that when we describe the `DID Document: INFRA<Map>` we are consistent for both production and consumption... 

> The JSON-LD specification defines values that are valid for this entry. This entry contains representation-specific syntax and therefore could be present in the data model to aid in lossless conversion. ***If the entry is present in the data model, it MUST be used during production unless either an alternative @context value is explicitly provided to the producer or if the value from the data model is not valid according to the consumption rules.***

We require the bolded language, and will object to any attempt to eliminate it. 

There are other DID Methods that also require this language, including `did:web`... This language is the result of many months of consensus gathering, and we should be careful not to construct pictures summarizing positions for which there is no consensus.

msporny:
> We require the bolded language, and will object to any attempt to eliminate it.

Agree with @OR13 -- the diagram does not represent what's in the spec today. There are no rules for just magic'ing a `@context` into existence.

msporny (changes requested):
@OR13 wrote:
> We require the bolded language, and will object to any attempt to eliminate it.

Agree with @OR13 -- the diagram does not represent what's in the spec today. There are no rules for magic'ing a `@context` into existence... it has to come from the ADM.

peacekeeper:
> You cannot produce JSON-LD without holding some representation of the `@context` in the ADM

@OR13 how is this statement compatible with the spec text you just quoted in your comment just above this one??

I.e.:

> This entry contains representation-specific syntax and therefore **could be** present in the data model [..] **If the entry is present** in the data model,

I think this makes it clear that `@context` and other representation-specific entries can be preserved in the data model, but it certainly doesn't have to be there...

msporny:
> I think this makes it clear that @context and other representation-specific entries can be preserved in the data model, but it certainly doesn't have to be there in order for production to work...

If `@context` isn't in the ADM... then how does production inject it? We had discussed that you can't safely inject `@context` because of non-determinism problems. I think there was an assumption that it /would/ be preserved and used. Here's the meeting where we had multiple discussions about this feature:

https://www.w3.org/2019/did-wg/Meetings/Minutes/2020-11-05-did#section2

msporny:
@peacekeeper now that we have the concept of a `representation-specific entries map` can you please update the diagrams to include that (noting that `@context` comes from that map for JSON-LD production). That should allow us to get these diagrams integrated. I will also note that the diagram in the data model section is now wrong (because it contemplates `@context` in the DID Document data model, which is now incorrect).

iherman:
> I will also note that the diagram in the data model section is now wrong (because it contemplates `@context` in the DID Document data model, which is now incorrect).

See https://github.com/w3c/did-core/pull/596#issuecomment-783989033 : I will take care of that one.

iherman:
> > I will also note that the diagram in the data model section is now wrong (because it contemplates `@context` in the DID Document data model, which is now incorrect).
> 
> See [#596 (comment)](https://github.com/w3c/did-core/pull/596#issuecomment-783989033) : I will take care of that one.

See https://github.com/w3c/did-core/pull/692.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2019/did-wg/Meetings/Minutes/2021-02-25-did-topic#section1) on 2021-02-25

- no resolutions were taken

### 1. Diagrams 

_See github pull request [#596](https://github.com/w3c/did-core/pull/596), [#597](https://github.com/w3c/did-core/pull/597)._




**Markus Sabadello:** there are two open PRs with diagrams that I was working on  
… one is [#596](https://github.com/w3c/did-core/pull/596)  
… on production and consumption  
… one on resolve and resolveRepresentation  
… 597, I have not worked on updating that yet  
… I have tried to update the production and consumption one  
… I came up with this version  
… latest attempt maybe not perfect, but how close is it to what people want to see?  
… tries to implement this separation of the maps that we're saying we're passing into the production function and getting out of the consumption function  
… and the did document data model and representation specific entries as a separate map  

**Manu Sporny:** this is not necessarily aligned with where I thought we were going  
… its' confusing because the bottom one looks like it's just using the representation specific entries  
… some changes  
… if you take did document data model at the top and figure out how to move it into that dashed box  
… and representation specific entries, strike json-ld, and move into the dashed box below  
… and do a larger dashed box around both  
… and have produce and consume go from that would solve the issue for me  

**Markus Sabadello:** *live edits diagram*  

**Manu Sporny:** it does dodge some of the questions, but addresses my concern, if everyone agrees  

**Markus Sabadello:** I can live with it  

> *Ivan Herman:* +1 with these changes

**Orie Steele:** what was proposed is consistent with the current diagram that has been merged into did core  
… the upper left hand box with dashed lines is already represented in did core under the data model section  
… i'd be fine if we used the existing diagram and didn't add another diagram that in some ways conflicts  
… the existing diagram we have in did core is acceptable for me  
… as a result of consumption  
… i'm struggling why we should show different pictures of the same data model  
… almost there, stop representing the data model differently in different diagrams  
… use the same diagram for each of them  

> *Shigeya Suzuki:* While I almost agree to the diagram... +1 to Orie's opinion

> *Shigeya Suzuki:* (reuse same diagram on same concept)

**Ted Thibodeau Jr.:** i'm concerned that surrounding dashed box suggests that those representation specific entries are part of all of the productions  

**Ivan Herman:** orie, when you talk about the diagram in the spec right now, are you talking about that one or the change that is part of another PR?  

**Manu Sporny:** merged few minutes ago  

**Ivan Herman:** you make our life difficult :-)  

> *Orie Steele:* I mean use this diagram: [https://w3c.github.io/did-core/#data-model](https://w3c.github.io/did-core/#data-model)

> *Orie Steele:* :)

**Manu Sporny:** I like where orie is going with this, we do have something already, can we reuse that  
… that would be better I think  

**Ivan Herman:** there is a difference, what markus is doing is with code extract which here is not the case  
… reusing the same style and colours would make sense  
… but it's not the same diagram  
… here we list some of the core properties, not all  

> *Shigeya Suzuki:* +1 to Ivan

> *Justin Richer:* +1 this is not the same diagram, this is how we get TO that diagram (and from it)

> *Ted Thibodeau Jr.:* +1 these are different diagrams for different purposes

> *Brent Zundel:* +1

**Markus Sabadello:** also some similarities and differences  
… both show the data model, but this one here that ivan made is intended to be more generic and higher level, whereas the other has a concrete example  
… the other difference is that this is showing the concepts of core things and registered extensions and unregistered extensions, which are irrelevant in the other diagram which is about production and consumption  

> *Drummond Reed:* +1 to the diagram we are seeing now on "production and consumption"

**Markus Sabadello:** using similar colours and style, I'd be happy to make an attempt to do that  

**Manu Sporny:** orie, would you be okay with that? I buy the argument that they're slightly different.. would you be okay that kind of set up?  

**Orie Steele:** we have a resolution that hasn't made it into the spec yet around ignoring and consuming properties  
… somewhat related to what we're talking about here  
… this diagram that's showing concrete production and consumption for three representations  
… aligning the colours is a good idea  
… dangerously to represent the adm abstractly and concretely in slightly different ways  
… an issue of colouring and consistency  
… and boxes coloured one way in the abstract sense, should show how properties are preserved during consumption should show up in those boxes  
… when you consume your properties are going into one of many different coloured boxes, and when you produce they're coming out of the coloured boxes and going into one representation  
… I like the image in did core right now, and this one doesn't look like that  
… the subtle differences are harmful  
… we should use the same labels  
… representation specific entries, core representations specific entries  
… all of the term labels should be the same  

**Manu Sporny:** I agree with what orie is saying in general  
… is markus willing to give that a shot?  
… Ted raised that it make sit look like you use all those things to produce and consume  
… that's why I was saying there's a bit of a handwave that is okay  
… you can  

> *Orie Steele:* For the record, I am fine with the "content" of the diagram, just concerned about colors / labels

**Manu Sporny:** nothing to say that an application can't go into these boxes and modify them between production and consumption  
… I might consume json-ld, and if I know i'm going to produce json and my application feels strongly about removing @context it has the choice of doing that  
… we don't have anything in the spec that says you can't  
… we suggest that you should preserve everything  
… but apps can process these boxes for their own reasons  
… in consumption we say you have to preserve it  
… but we don't say you have to all the way through to production  
… you can modify them between those two  
… thats why we're handwaving a bit over this  
… I think that's okay because we need to get something that we're more or less okay with, otherwise this is going to keep getting blocked  

> *Orie Steele:* For example, you are not required to preserve prototype pollution or sql injection attacks :)

**Markus Sabadello:** agree that what we have now would allow both  
… it would allow having @context int he representations specific entries then producing plain json, and including `@context` there though personally I don't like it I think it's allowed and possible with the spec language  
… what I could also do is what we discussed before to split this up into multiple diagrams where we show different flows or different ways of using this  
… first consumption, then production  
… sometimes `@context` is there, sometimes it's not  
… I'd rather have one diagram  

**Brent Zundel:** feels like we have consensus on direction  
… I'd also rather have one diagram  
… opposition to moving on to testing?  

> *Orie Steele:* Also for the record, when you advocate for producing application/did+json without an @context, you are advocating for incompatibility / non-interoperability with JSON-LD verifiable credentials.

</details>

msporny:
I'm running out of time to merge this PR before freeze this coming Monday COB ET. I need an update @peacekeeper -- ETA? If we can't get this in before CR, it's not a big deal, it's editorial, it can go in after we're in CR.

peacekeeper:
Here's an update to the diagram, analogous to https://github.com/w3c/did-core/pull/596:

diagram-resolve-resolverepresentation

iherman:
mini-minor knit: the spacing in the middle box for the JSON code (in application/did+json) does not align... Otherwise +1 for me.

OR13 (approved):

OR13:
@msporny approved

peacekeeper:
> mini-minor knit: the spacing in the middle box for the JSON code (in application/did+json) does not align... Otherwise +1 for me.

Sorry, fixed now:

diagram-resolve-resolverepresentation

msporny (approved):
We'll need to fix the alt text.

PR #600, opened 2021-02-04
Adding a detailed overview diagram to appendix

shigeya:
**Experimental**
Keeping old figure for review.

Related to #453.  Trying to include all of the relationships of the entities described in 1.3. Architecture Overview.

This diagram does not display DID methods, DID resolvers, and DID URL dereferences in this picture since their position in this level of abstraction is unclear.

***
<a href="https://pr-preview.s3.amazonaws.com/shigeya/did-core/pull/600.html" title="Last updated on Feb 10, 2021, 4:34 AM UTC (36b8395)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/600/59687a4...shigeya:36b8395.html" title="Last updated on Feb 10, 2021, 4:34 AM UTC (36b8395)">Diff</a>

PR #592, opened 2021-02-02
Make DID Resolution section concrete and testable (abstract + Typescript). 

msporny:
This PR is intended to address issue #549 by adding to the abstract functions discussed in the DID Resolution and DID URL Dereferencing sections and describing potentially conforming expressions in Typescript as concrete, testable functions. The PR uses an Typescript interface to provide an example while keeping all of what we had before. It also tries to strike a balance by providing at least one concrete realization of the interface without preventing other concrete interfaces from being realized.

There are a LOT of reformatting changes in this PR, it might be best to <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/592.html#resolution">look at the rendered preview for this PR</a>.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/592.html" title="Last updated on Feb 6, 2021, 5:18 PM UTC (028302a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/592/518362a...028302a.html" title="Last updated on Feb 6, 2021, 5:18 PM UTC (028302a)">Diff</a>

PR #601, opened 2021-02-06
Clean up DID Resolution section; enable testability via proof by construction

msporny:
This PR is intended to address issue #549 by adding to the abstract functions discussed in the DID Resolution and DID URL Dereferencing sections and ensuring that all statements are testable using a "proof by construction". That is, all statements that return abstract data models are testable due to additional normative statements that effectively state: "anything that returns an abstract data model MUST be serializable in one of the representations". Previously, that statement didn't apply to `didDocument`, the metadata structure, some of the dereferencing values, etc. This PR fixes that.

The PR also adds at-risk markers to some of the newer features requested by individuals (but not necessarily broadly requested by the WG -- such as nextUpdate, nextVersionId, equivalentId, and canonicalId. There are also a few bugs fixed in this version (such as stating that a contentType must be specified for the resolve call -- that returns an abstract didDocument... which is clearly wrong).

There are a LOT of reformatting changes in this PR, it might be best to <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/601.html#resolution">look at the rendered preview for this PR</a>.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/601.html" title="Last updated on Feb 10, 2021, 9:57 PM UTC (a58ae7f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/601/3858d12...a58ae7f.html" title="Last updated on Feb 10, 2021, 9:57 PM UTC (a58ae7f)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -4149,8 +4149,8 @@ to the <code>dereference</code> function specified here.
 The possible properties within this structure and their possible values SHOULD
 be registered in the DID Specification Registries [[?DID-SPEC-REGISTRIES]].
 This specification defines the following common properties. This
-specification defines the following common properties for a
-<dfn>DereferencingOptions</dfn> <a data-cite="INFRA#maps">map</a>:
+specification defines the following common properties for dereferencing
+options:
             </p>
 
             <dl>
--- a/index.html
+++ b/index.html
@@ -4139,7 +4139,7 @@ contentStream
         <dd>
 If the <code>dereferencing</code> function was called and successful, this MUST
 contain a <a>resource</a> corresponding to the <a>DID URL</a>. The
-<code>content-stream</code> SHOULD be a <a>DID document</a> in one of the
+<code>contentStream</code> SHOULD be a <a>DID document</a> in one of the
 conformant <a>representations</a> obtained through the resolution process.<br><br>
 
 If the dereferencing is unsuccessful, this value MUST be empty.
@@ -4208,7 +4208,7 @@ specification defines the following common properties.
 contentType
                 </dt>
                 <dd>
-The MIME type of the returned <code>content-stream</code> SHOULD be expressed
+The MIME type of the returned <code>contentStream</code> SHOULD be expressed
 using this property if dereferencing is successful.
                 </dd>
                 <dt>
@@ -4233,7 +4233,7 @@ notFound
                         </dt>
                         <dd>
 The <a>DID URL dereferencer</a> was unable to find the
-<code>content-stream</code> resulting from this dereferencing request.
+<code>contentStream</code> resulting from this dereferencing request.
                         </dd>
                     </dl>
                 </dd>
--- a/index.html
+++ b/index.html
@@ -3682,13 +3682,16 @@ The <a>DID resolution</a> function resolves a <a>DID</a> into a <a>DID
 document</a> by using the "Read" operation of the applicable <a>DID method</a>
 as described in <a href="#read-verify"></a>. The details of how this process is
 accomplished are outside the scope of this specification, but all conforming
-<a>DID resolvers</a> implement two functions which have the following abstract
-forms:
+<a>DID resolvers</a> implement at least one of the functions below which
+have the following abstract forms:
         </p>
 
         <pre title="Abstract functions for DID Resolution">
-resolve(did, options) -> Resolution
-resolveRepresentation(did, options) -> ResolutionRepresentation
+resolve(did, options) →
+   « didResolutionMetadata, didDocument, didDocumentMetadata »
+
+resolveRepresentation(did, options) →
+   « didResolutionMetadata, didDocumentStream, didDocumentMetadata »
         </pre>
 
         <p>
@@ -4071,7 +4074,8 @@ the following function which has the following abstract form:
       </p>
 
       <pre title="Abstract functions for DID Dereferencing">
-dereference(didUrl, options) -> DereferencingResponse
+dereference(didUrl, options) →
+   « didUrlDereferencingMetadata, contentStream, contentMetadata »
       </pre>
 
       <p>
--- a/index.html
+++ b/index.html
@@ -3944,6 +3944,11 @@ document. It indicates the version of the next
 
           <dt><dfn>equivalentId</dfn></dt>
           <dd>
+            <p class="issue atrisk">
+The DID Working Group is seeking implementer feedback on this feature. If there
+is not enough implementation experience with this feature at the end of the
+Candidate Recommendation period, it will be removed from the specification.
+            </p>
             <p>
 A <a>DID Method</a> can define different forms of a <a>DID</a> that are
 logically equivalent. An example is when a <a>DID</a> takes one form prior to
@@ -3995,48 +4000,56 @@ directives related to this metadata property.
             </dd>
             <dt><dfn>canonicalId</dfn></dt>
             <dd>
+              <p class="issue atrisk">
+The DID Working Group is seeking implementer feedback on this feature. If there
+is not enough implementation experience with this feature at the end of the
+Candidate Recommendation period, it will be removed from the specification.
+              </p>
+              <p>
 The <code><a>canonicalId</a></code> property is identical to the
-<code><a>equivalentId</a></code> property except: a) it accepts only a single
-value rather than a list, and b) that <a>DID</a> is defined to be the canonical ID for
-the <a>DID subject</a> within the scope of the containing <a>DID document</a>.<br><br>
-
+<code><a>equivalentId</a></code> property except: a) it is associated with a
+single value rather than a ordered set, and b) the <a>DID</a> is defined to be
+the canonical ID for the <a>DID subject</a> within the scope of the containing
+<a>DID document</a>.
+              </p>
+              <p>
 The value of <code><a>canonicalId</a></code> MUST be a <a
 data-cite="INFRA#string">string</a> that conforms to the rules in Section <a
-href="#did-syntax"></a>.<br><br>
-
-The relationship is a statement that the <code><a>canonicalId</a></code> value
-is logically equivalent to the <code>id</code> property value and that the
-<code><a>canonicalId</a></code> value is defined by the <a>DID Method</a> to be
-the canonical ID for the <a>DID subject</a> in the scope of the containing <a>DID
-document</a>.<br><br>
-
-A <code><a>canonicalId</a></code> value MUST be produced by, and a form of, the
-same <a>DID Method</a> as the <code>id</code> property value. (e.g.,
-<code>did:example:abc</code> == <code>did:example:ABC</code>)
-                  </dd>
-                  <dd>
-
-A conforming <a>DID Method</a> specification MUST guarantee that the
+href="#did-syntax"></a>. The relationship is a statement that the
 <code><a>canonicalId</a></code> value is logically equivalent to the
-<code>id</code> property value.<br><br>
-
+<code>id</code> property value and that the <code><a>canonicalId</a></code>
+value is defined by the <a>DID Method</a> to be the canonical ID for the <a>DID
+subject</a> in the scope of the containing <a>DID document</a>. A
+<code><a>canonicalId</a></code> value MUST be produced by, and a form of, the
+same <a>DID Method</a> as the <code>id</code> property value. (e.g.,
+<code>did:example:abc</code> == <code>did:example:ABC</code>). A conforming
+<a>DID Method</a> specification MUST guarantee that the
+<code><a>canonicalId</a></code> value is logically equivalent to the
+<code>id</code> property value.
+              </p>
+              <p>
 A requesting party is expected to use the <code><a>canonicalId</a></code> value
-as its primary ID value for the <a>DID subject</a> and treat all other equivalent
-values as secondary aliases. (e.g. update corresponding primary references in
-their systems to reflect the new canonical ID directive).<br><br>
-
+as its primary ID value for the <a>DID subject</a> and treat all other
+equivalent values as secondary aliases (e.g. update corresponding primary
+references in their systems to reflect the new canonical ID directive).
+              </p>
+              <p>
 <code><a>canonicalId</a></code> is the same statement of equivalence as
 <code><a>equivalentId</a></code> except it is constrained to a single value that
-is defined to be canonical for the <a>DID subject</a> in the scope of the <a>DID document</a>.
-Like <code><a>equivalentId</a></code>, <code><a>canonicalId</a></code>
-represents a full graph merge because the same <a>DID document</a> describes both the
-<code><a>canonicalId</a></code> DID and the <code>id</code> property <a>DID</a>.<br><br>
-
-If a resolving party does not use the <code><a>canonicalId</a></code> value
-as its primary ID value for the DID subject and treat all other equivalent
-values as secondary aliases, there might be negative or unexpected issues that
-arise related to user experience. Implementers are strongly advised to observe the
-directives related to this metadata property.<br><br>
+is defined to be canonical for the <a>DID subject</a> in the scope of the <a>DID
+document</a>. Like <code><a>equivalentId</a></code>,
+<code><a>canonicalId</a></code> represents a full graph merge because the same
+<a>DID document</a> describes both the <code><a>canonicalId</a></code> DID and
+the <code>id</code> property <a>DID</a>.
+              </p>
+              <p>
+If a resolving party does not use the <code><a>canonicalId</a></code> value as
+its primary ID value for the DID subject and treat all other equivalent values
+as secondary aliases, there might be negative or unexpected issues that arise
+related to user experience. Implementers are strongly advised to observe the
+directives related to this metadata property.
+              </p>
+              <p>
           </dd>
         </dl>
       </section>
--- a/index.html
+++ b/index.html
@@ -3736,10 +3736,10 @@ The return value of <code>resolve</code> is the
             <dd>
 A <a href="#metadata-structure">metadata structure</a> consisting of values
 relating to the results of the <a>DID resolution</a> process. This structure is
-REQUIRED and MUST NOT be empty. This metadata is defined by
-<a href="#did-resolution-metadata"></a>. If the resolution is successful this
-structure MUST contain a <code>contentType</code> property containing the
-mime-type of the <code>didDocument</code> in this result. If the resolution is
+REQUIRED and MUST NOT be empty. This metadata is defined by <a
+href="#did-resolution-metadata"></a>. If <code>resolveRepresentation</code> was
+called, this structure MUST contain a <code>contentType</code> property
+containing the MIME type of the <code>didDocument</code>. If the resolution is
 not successful, this structure MUST contain an <code>error</code> property
 describing the error.
             </dd>
@@ -4142,7 +4142,7 @@ contain a <a>resource</a> corresponding to the <a>DID URL</a>. The
 <code>contentStream</code> MAY be a <a>resource</a> such as a <a href="">DID
 Document</a>,  <a href="#verification-methods">Verification Method</a>,  or <a
 href="#services">service</a> that is serializable in one of the conformant
-<a>representations</a> obtained through the resolution process 
+<a>representations</a> obtained through the resolution process
 or any other resource format that can be identified via a MIME type. If the
 dereferencing is unsuccessful, this value MUST be empty.
         </dd>
--- a/index.html
+++ b/index.html
@@ -4047,7 +4047,7 @@ realize an interfaces that is functionally equivalent to the following example:
 
       <pre class="example" title="Conforming dereferencer interface in Typescript">
 interface DidDereferencer {
-  dereference(USVString did,
+  dereference(string did,
               DereferencingOptions dereferencingOptions): DereferencingResponse;
 };
       </pre>
--- a/index.html
+++ b/index.html
@@ -3709,8 +3709,8 @@ interface DidResolver {
         </pre>
 
         <p>
-The input variables of the <code>resolve()</code> and
-<code>resolveRepresentation()</code> functions are as follows:
+The input variables of the <code>resolve</code> and
+<code>resolveRepresentation</code> functions are as follows:
         </p>
 
         <dl>
@@ -3732,11 +3732,11 @@ REQUIRED, but the structure MAY be empty.
         </dl>
 
         <p>
-The return value of <code>resolve()</code> is <dfn>Resolution</dfn> which is
+The return value of <code>resolve</code> is <dfn>Resolution</dfn> which is
 a <a data-cite="INFRA#maps">map</a> that contains the
 <a>didResolutionMetadata</a>, <a>didDocument</a>, and
 <a>didDocumentMetadata</a> properties. The return value of
-<code>resolveRepresentation()</code> is <dfn>ResolutionRepresentation</dfn>
+<code>resolveRepresentation</code> is <dfn>ResolutionRepresentation</dfn>
 which is a <a data-cite="INFRA#maps">map</a> that contains
 <a>didResolutionMetadata</a>, <a>didDocumentStream</a>, and
 <a>didDocumentMetadata</a> properties. These properties are described below:
@@ -3785,7 +3785,7 @@ If the resolution is successful, this MUST be a <a
 href="#metadata-structure">metadata structure</a>. This structure contains
 metadata about the <a>DID document</a> contained in the
 <code>didDocument</code> property. This metadata
-typically does not change between invocations of the <code>resolveRepresentation()</code>
+typically does not change between invocations of the <code>resolveRepresentation</code>
 function unless the <a>DID document</a> changes, as it represents data about
 the <a>DID document</a>. If the resolution is unsuccessful, this output MUST
 be an empty <a href="#metadata-structure">metadata structure</a>. Properties
@@ -3801,7 +3801,7 @@ this function to a
 method-specific internal function to perform the actual <a>DID resolution</a>
 process. <a>DID resolver</a> implementations might implement and expose
 additional functions with different signatures in addition to the
-<code>resolveRepresentation()</code> function specified here.
+<code>resolveRepresentation</code> function specified here.
         </p>
 
         <section>
@@ -3845,10 +3845,10 @@ contentType
                 <dd>
 The MIME type of the returned <code>didDocument</code>. This property is
 REQUIRED if resolution is successful. This property MUST NOT
-be present if the <code>resolveRepresentation()</code> function was called. The value of this
+be present if the <code>resolveRepresentation</code> function was called. The value of this
 property MUST be an <a data-lt="ascii string">ASCII string</a> that is the MIME
 type of the conformant <a>representations</a>. The
-caller of <code>resolveRepresentation()</code> function MUST use this value
+caller of <code>resolveRepresentation</code> function MUST use this value
 when determining how to parse and process the <code>didDocument</code>
 returned by this function into the <a href="#data-model">data model</a>.
                 </dd>
--- a/index.html
+++ b/index.html
@@ -3676,8 +3676,8 @@ The <a>DID resolution</a> function resolves a <a>DID</a> into a <a>DID
 document</a> by using the "Read" operation of the applicable <a>DID method</a>
 as described in <a href="#read-verify"></a>. The details of how this process is
 accomplished are outside the scope of this specification, but all conforming
-DID Resolver implementations realize at least the following interface (or one
-that is functionally equivalent):
+<a>DID resolver</a> implementations realize at least the following interface
+(or one that is functionally equivalent):
         </p>
 
         <pre class="idl">
@@ -3686,11 +3686,6 @@ interface mixin DidResolver {
 };
         </pre>
 
-        <p>
-The {{DidResolver/resolve}} function returns a byte stream of the
-<a>DID Document</a> formatted in the corresponding representation,
-DID Resolution Metadata, and DID Document Metadata.
-        </p>
         <p>
 The input variables of the {{DidResolver/resolve}} function are as follows:
         </p>
@@ -3746,14 +3741,12 @@ an <code>error</code> property describing the error.
 didDocument
             </dt>
             <dd>
-If the resolution is successful, and if the <code>resolveRepresentation</code>
-function was called, this MUST be a byte stream of the resolved <a>DID
-document</a> in one of the conformant
+If the resolution is successful this MUST be a byte stream of the resolved
+<a>DID document</a> in one of the conformant
 <a href="#representations">representations</a>. The byte stream might then be
-parsed by the caller of the <code>resolveRepresentation</code> function into a
-<a href="#data-model">data model</a>, which can in turn be validated and
-processed. If the resolution is unsuccessful, this value MUST be an empty
-stream.
+parsed by the caller into a <a href="#data-model">data model</a>, which can in
+turn be validated and processed. If the resolution is unsuccessful, this value
+MUST be an empty stream.
             </dd>
             <dt>
 didDocumentMetadata
@@ -3762,13 +3755,13 @@ didDocumentMetadata
 If the resolution is successful, this MUST be a <a
 href="#metadata-structure">metadata structure</a>. This structure contains
 metadata about the <a>DID document</a> contained in the
-<code>didDocument</code>. This metadata
+<code>didDocument</code> property. This metadata
 typically does not change between invocations of the {{DidResolver/resolve}}
 function unless the <a>DID document</a> changes, as it represents data about
 the <a>DID document</a>. If the resolution is unsuccessful, this output MUST
 be an empty <a href="#metadata-structure">metadata structure</a>. Properties
 defined by this specification are in <a
-href="#did-document-metadata-properties"></a>.
+href="#did-document-metadata"></a>.
             </dd>
         </dl>
 
@@ -3807,7 +3800,7 @@ available. This property is OPTIONAL.
         </section>
 
         <section>
-            <h3>DID Resolution Metadata Properties</h3>
+            <h3>DID Resolution Metadata</h3>
 
             <p>
 The possible properties within this structure and their possible values are
@@ -3876,7 +3869,7 @@ deactivated. (See <a href="#deactivate"></a>.)
         </section>
 
         <section>
-          <h3>DID Document Metadata Properties</h3>
+          <h3>DID Document Metadata</h3>
 
           <p>
 The possible properties within this structure and their possible values SHOULD
@@ -3884,21 +3877,19 @@ be registered in the DID Specification Registries [[?DID-SPEC-REGISTRIES]].
 This specification defines the following common properties.
           </p>
 
-          <section>
-            <h4>created</h4>
-            <p>
+          <dl>
+            <dt><dfn>created</dfn></dt>
+            <dd>
 <a>DID document</a> metadata SHOULD include a <code>created</code> property to
 indicate the timestamp of the <a href="#create">Create operation</a>. The value
 of the property MUST be a <a data-cite="INFRA#string">string</a> formatted as an
 <a data-cite="XMLSCHEMA11-2#dateTime">XML Datetime</a> normalized to UTC
 00:00:00 and without sub-second decimal precision. For example:
 <code>2020-12-20T19:17:47Z</code>.
-            </p>
-          </section>
+            </dd>
 
-          <section>
-            <h4>updated</h4>
-            <p>
+            <dt><dfn>updated</dfn></dt>
+            <dd>
 <a>DID document</a> metadata SHOULD include an <code>updated</code> property to
 indicate the timestamp of the last <a href="#update">Update operation</a> for
 the document version which was resolved. The value of the property MUST follow
@@ -3907,83 +3898,53 @@ the same formatting rules as the <code>created</code> property. The
 performed on the <a>DID document</a>. If an <code>updated</code> property exists, it
 can be the same value as the <code>created</code> property when the difference
 between the two timestamps is less than one second.
-                </p>
-              </section>
+            </dd>
 
-          <section>
-            <h4>deactivated</h4>
-            <p>
-This property MUST be populated by a <a data-cite="INFRA#boolean">boolean</a>
-value that is true if the DID has been deactivated, and false if the DID is still active.
-`true` value if the DID Method determines that the <a>DID</a> supplied has been
-deactivated. (See <a href="#deactivate"></a>.)
-                </p>
-              </section>
-
-            <section>
-                <h4>
-                  nextUpdate
-                </h4>
-                <p>
+            <dt><dfn>nextUpdate</dfn></dt>
+            <dd>
 <a>DID document</a> metadata MAY include a <code>nextUpdate</code> property
 if the resolved document version is not the latest version of the document. It
 indicates the timestamp of the next <a href="#update">Update operation</a>. The
 value of the property MUST follow the same formatting rules as the
 <code>created</code> property.
-            </p>
-          </section>
+            </dd>
 
-          <section>
-              <h4>
-                versionId
-              </h4>
-              <p>
+            <dt><dfn>versionId</dfn></dt>
+            <dd>
 <a>DID document</a> metadata SHOULD include a <code>versionId</code> property to
 indicate the version of the last <a href="#update">Update operation</a> for the
 document version which was resolved. The value of the property MUST be an <a
 data-lt="ascii string">ASCII string</a>.
-              </p>
-            </section>
+            </dd>
 
-          <section>
-              <h4>
-                nextVersionId
-              </h4>
-              <p>
+            <dt><dfn>nextVersionId</dfn></dt>
+            <dd>
 <a>DID document</a> metadata MAY include a <code>nextVersionId</code>
 property if the resolved document version is not the latest version of the
 document. It indicates the version of the next
 <a href="#update">Update operation</a>. The value of the property MUST be an
 <a data-lt="ascii string">ASCII string</a>.
-              </p>
-            </section>
+            </dd>
 
-          <section>
-            <h4>equivalentId</h4>
-                <p>
+          <dt><dfn>equivalentId</dfn></dt>
+          <dd>
 A <a>DID Method</a> can define different forms of a <a>DID</a> that are
 logically equivalent. An example is when a <a>DID</a> takes one form prior to
 registration in a <a>verifiable data registry</a> and another form after such
 registration. In this case, the <a>DID Method</a> specification may need to
 express one or more <a>DIDs</a> that are logically equivalent to the resolved
 <a>DID</a> as a property of the <a>DID document</a>. This is the purpose of the
-<code><a>equivalentId</a></code>  property.
-                </p>
+<code><a>equivalentId</a></code>  property. <br><br>
 
-                <dl>
-                  <dt><dfn>equivalentId</dfn></dt>
-                  <dd>
 The value of <code>equivalentId</code> MUST be an
 <a data-cite="INFRA#ordered-set">ordered set</a> where each item in the list is
 a <a data-cite="INFRA#string">string</a> that conforms to the rules in Section
-<a href="#did-syntax"></a>.
-                  </dd>
-                  <dd>
+<a href="#did-syntax"></a>.<br><br>
+
 The relationship is a statement that each <code><a>equivalentId</a></code> value
 is logically equivalent to the <code>id</code> property value and thus
-identifies the same <a>DID subject</a>.
-                  </dd>
-                  <dd>
+identifies the same <a>DID subject</a>.<br><br>
+
 Each <code><a>equivalentId</a></code> DID value MUST be produced by, and a form
 of, the same <a>DID Method</a> as the <code>id</code> property value. (e.g.,
 <code>did:example:abc</code> == <code>did:example:ABC</code>)
@@ -3991,9 +3952,8 @@ of, the same <a>DID Method</a> as the <code>id</code> property value. (e.g.,
                   <dd>
 A conforming <a>DID Method</a> specification MUST guarantee that each
 <code><a>equivalentId</a></code> value is logically equivalent to the
-<code>id</code> property value.
-                  </dd>
-                  <dd>
+<code>id</code> property value.<br><br>
+
 A requesting party is expected to retain the values from the <code>id</code> and
 <code><a>equivalentId</a></code> properties to ensure any subsequent
 interactions with any of the values they contain are correctly handled as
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -3686,46 +3692,42 @@ <h1>Resolution</h1>
...
-        </code></p>
+        <pre title="Abstract functions for DID Resolution">
+resolve(did, options) →
+   « didResolutionMetadata, didDocument, didDocumentMetadata »
 
-        <p><code>
-resolveRepresentation ( did, did-resolution-input-metadata ) <br>
-&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;-&gt; ( did-resolution-metadata, did-document-stream, did-document-metadata )
-        </code></p>
+resolveRepresentation(did, options) →
```
OR13, line 3721: do we mean DID or DID URL?

Would prefer to see a note that clarifies the type... and explicitly states if `path`, `query` and `fragment` are allowed.

index.html
```diff
@@ -3686,46 +3692,42 @@ <h1>Resolution</h1>
...
+<a>DID resolvers</a> implement the functions below which have the following
+abstract forms:
         </p>
 
-        <p><code>
-resolve ( did, did-resolution-input-metadata ) <br>
-&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;-&gt; ( did-resolution-metadata, did-document, did-document-metadata )
-        </code></p>
+        <pre title="Abstract functions for DID Resolution">
+resolve(did, options) →
```
OR13, line 3718: do we mean DID or DID URL?

Would prefer to see a note that clarifies the type... and explicitly states if `path`, `query` and `fragment` are allowed.

OR13: if it's possible to hyperlink to definitions here, that might help avoid any confusion.

index.html
```diff
@@ -3798,60 +3795,59 @@ <h2>DID Resolution</h2>
...
-determine the <a>representation</a> contained in the returned
-<code>did-document-stream</code> if such a <a>representation</a> is supported and
-available. This property is OPTIONAL. It is only used if the
-<code>resolveRepresentation</code> function is called and MUST be ignored if the
-<code>resolve</code> function is called.
+The MIME type of the caller's preferred <a>representation</a> of the <a>DID
+document</a>. The MIME type MUST be expressed as an  <a data-lt="ascii
+string">ASCII string</a>. The <a>DID resolver</a> implementation SHOULD use this
+value to determine the <a>representation</a> contained in the returned
+<code>didDocumentStream</code> if such a <a>representation</a> is supported and
```
OR13, line 3843: What is its not supported... what if I ask for:

```
accept: application/did+dag+cbor, application/did+json
``

Prefer to see a note encouraging or discouraging alignment with https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/406

File: index.html [hunk 4]
```diff
@@ -3944,283 +3945,242 @@ <h4>updated</h4>
...
+of this specification, but all conforming <a>DID resolvers</a> implement
+the following function which has the following abstract form:
       </p>
 
-      <p><code>
-dereference ( did-url, did-url-dereferencing-input-metadata ) <br>
-&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;-&gt; ( did-url-dereferencing-metadata, content-stream, content-metadata )
-      </code></p>
+      <pre title="Abstract functions for DID Dereferencing">
+dereference(didUrl, options) →
```
OR13, line 4115: hmmm, this answers my question about... but it feels odd that this abstract function is defined so far away from the other ones... wonder if we can define them all together in one place, and then address them one at a time like this PR is doing.

index.html
```diff
@@ -3944,283 +3945,242 @@ <h4>updated</h4>
...
 
       <p>
-The input variables of this function are as follows:
+The input variables of the <code>dereference</code> function are as follows:
       </p>
 
       <dl>
         <dt>
-did-url
+didUrl
```
OR13, line 4125: here would be a good place to warn about `query` and `path` and their relatively poor utilization... or add some detail on why we kept them in the spec.

index.html
```diff
@@ -3944,283 +3945,242 @@ <h4>updated</h4>
...
-The <code>content-stream</code> SHOULD be a <a>DID document</a> in one of the
-conformant <a>representations</a> obtained through
-the resolution process.
-
-If the dereferencing is unsuccessful, this value MUST be empty.
+contain a <a>resource</a> corresponding to the <a>DID URL</a>. The
+<code>contentStream</code> SHOULD be a <a>resource</a> such as a <a href="">DID
+Document</a>,  <a href="#verification-methods">Verification Method</a>,  or <a
+href="#services">service</a> that is serializable in one of the conformant
+<a>representations</a> obtained through the resolution process. If the
```
OR13, line 4171: hmm, is this true of dereferencing an image with location transparency?

For example:

```
did:example:123?service=images&relativeRef=/favicon.png
```

I expect the content type to be `image/png` not `application/did+json`....

index.html
```diff
@@ -3944,283 +3945,242 @@ <h4>updated</h4>
...
 metadata structure</a>, but the structure MAY be empty. This structure contains
-metadata about the <code>content-stream</code>.
-If the <code>content-stream</code> is a <a>DID document</a>, this MUST be a
-<code>did-document-metadata</code> structure as described in <a>DID
-Resolution</a>.
-
-If the dereferencing is unsuccessful, this output MUST be an empty <a
-href="metadata-structure">metadata structure</a>.
+metadata about the <code>contentStream</code>. If the <code>contentStream</code>
+is a <a>DID document</a>, this MUST be a <a>didDocumentMetadata</a> structure as
```
OR13, line 4181: implies that contentStream can be something other than a registered did document representation mime type... see https://github.com/w3c/did-core/pull/601/files#r571649879

index.html
```diff
@@ -4235,29 +4195,32 @@ <h2>DID URL Dereferencing</h2>
...
-URL dereferencing</a> implementation SHOULD use this value to determine the
-<a>representation</a> contained in the returned value if such a <a>representation</a> is
-supported and available.
-                </dd>
+              </dt>
+              <dd>
+The MIME type that the caller prefers for <code>contentStream</code>. The  MIME
+type MUST be expressed as an <a data-lt="ascii string">ASCII string</a>.
+The <a>DID URL dereferencing</a> implementation SHOULD use this value to
+determine the <a>representation</a> contained in the returned value if such a
```
OR13, line 4216: I think we may mean `contentType` of the `resource`... and not want to use `representation` here.

OR13 (changes requested):
Overall this feels like its heading in the correct direction.

Blocking feedback:

I am confused over the relationship between `accept`, `contentType` and `representation` and `resource`. I think we mean for DID URLs to be capable of dereferencing to resources other than DID Documents, yet we are not clearly saying that here... but we sorta allude to it.

I think the solution is to define the abstract functions all together:

`resolve`, `resolveRepresentation` and `dereference`

And describe the domain of `accept` and `contentType`.

for each, if they are different (I am asserting they are, for example `resolve` does not produce a `contentType`... `resolveRepresentation` produces `contentType` of DID Document Representation, and `dereference` produces `contentType` of IANA Registered MIME type (which includes did documents and images).

Nonblocking feedback:

I think it would help to hyperlink the terms to their definition, and be more excplit about where `path`, `query` and `fragment` are allowed, required, or ignored.

OR13:
We may find it helpful to review / agree to https://github.com/w3c/did-core/pull/600 before attempting further changes here.

index.html
```diff
@@ -3944,283 +3945,242 @@ <h4>updated</h4>
...
-statements related to requesting parties may be downgraded in the future from a
-MUST to a SHOULD/MAY or similar language.</span>
-                  </dd>
-                </dl>
-
-                <div class="note" title="Equivalence and equivalentId">
-                  <p>
+with any one maps to the same underlying account).
+            </p>
+            <p>
```
rhiaro, line 4025:
```suggestion
            <p class="note" title="Stronger equivalence">
```

Suggest making this paragraph a note, and moving it to the end (to be after the next paragraph).

index.html
```diff
@@ -3944,283 +3945,242 @@ <h4>updated</h4>
...
-                  </dd>
-                </dl>
-
-                <div class="note" title="Equivalence and canonicalId">
-                  <p>
+as its primary ID value for the <a>DID subject</a> and treat all other
+equivalent values as secondary aliases (e.g. update corresponding primary
+references in their systems to reflect the new canonical ID directive).
+              </p>
+              <p>
```
rhiaro, line 4077:
```suggestion
              <p class="note" title="Canonical equivalence">
```

And move to after the next paragraph

index.html
```diff
@@ -4318,24 +4283,26 @@ <h2>Metadata Structure</h2>
 set</a>, <a data-cite="INFRA#boolean">boolean</a>, or
 <a data-cite="INFRA#null">null</a>. The values within any complex data
 structures such as maps and lists MUST be one of these data types as well.
-All metadata property definitions MUST define the value type, including any
+All metadata property definitions registered in the DID Specification
+Registries [[DID-SPEC-REGISTRIES] MUST define the value type, including any
```
rhiaro, line 4287:
```suggestion
Registries [[?DID-SPEC-REGISTRIES]] MUST define the value type, including any
```

index.html
peacekeeper: resolve() takes a DID, not DID URL

index.html
peacekeeper: resolveRepresentation() takes a DID, not DID URL

index.html
peacekeeper: There's a "representationNotSupported" error defined in the spec. Also note that DID Resolution has no dependency on HTTP.

peacekeeper:
> There are also a few bugs fixed in this version (such as stating that a contentType must be specified for the resolve call

@msporny Where does the current text state that?

index.html
```diff
@@ -3737,55 +3739,50 @@ <h2>DID Resolution</h2>
...
-If the resolution is successful, and if the <code>resolveRepresentation</code>
-function was called, this structure MUST contain a <code>contentType</code>
-property containing the mime-type of the <code>did-document-stream</code> in
-this result. If the resolution is not successful, this structure MUST contain
-an <code>error</code> property describing the error.
-              </p>
+relating to the results of the <a>DID resolution</a> process. This structure is
+REQUIRED and MUST NOT be empty. This metadata is defined by
+<a href="#did-resolution-metadata"></a>. If the resolution is successful this
+structure MUST contain a <code>contentType</code> property containing the
```
peacekeeper, line 3769: Only if resolveRepresentation() was called, not if resolve() was called. Why this change?

index.html
```diff
@@ -3737,55 +3739,50 @@ <h2>DID Resolution</h2>
...
             <dt>
-did-document
+<dfn>didDocument</dfn>
             </dt>
             <dd>
 If the resolution is successful, and if the <code>resolve</code> function was
-called, this MUST be a <a>conforming DID document</a>. If the resolution is
+called, this MUST be a <a>DID document</a> (abstract data model) as described in
+<a href="#data-model"></a> that is capable of being transformed
+into a <a>conforming DID Document</a> (representation). If the resolution is
```
peacekeeper, line 3781:
```suggestion
into a <a>conforming DID Document</a> (representation), using the production rules specified by the representation. If the resolution is
```

+1 to your change, I like it a lot and think it helps with testability. Here is just a small additional suggestion to explicitly reference "production".

index.html
```diff
@@ -3798,60 +3795,59 @@ <h2>DID Resolution</h2>
...
+<dfn>didDocumentMetadata</dfn>
             </dt>
             <dd>
 If the resolution is successful, this MUST be a <a
 href="#metadata-structure">metadata structure</a>. This structure contains
 metadata about the <a>DID document</a> contained in the
-<code>did-document</code> or <code>did-document-stream</code>. This metadata
-typically does not change between invocations of the <code>resolve</code>
+<code>didDocument</code> property. This metadata
+typically does not change between invocations of the <code>resolveRepresentation</code>
```
peacekeeper, line 3805: Why this change? `didDocumentMetadata` is returned both for the `didDocument` and for the `didDocumentStream`, returned by `resolve()` and `resolveRepresentation()` respectively.

index.html
```diff
@@ -3737,55 +3739,50 @@ <h2>DID Resolution</h2>
...
             <dd>
-              <p>
 A <a href="#metadata-structure">metadata structure</a> consisting of values
-relating to the results of the <a>DID resolution</a> process. This structure
-is REQUIRED and MUST NOT be empty.
-              </p>
-              <p>
-This metadata typically changes between invocations of the
-<code>resolve</code> and <code>resolveRepresentation</code> functions as it
-represents data about the resolution process itself. Properties defined by
```
peacekeeper, line 3768: Why delete this? There is complementary language to this below about `didResolutionMetadata` which apparently you didn't delete?

index.html
```diff
@@ -3798,60 +3795,59 @@ <h2>DID Resolution</h2>
...
+href="#did-document-metadata"></a>.
             </dd>
         </dl>
 
         <p>
 Conforming <a>DID resolver</a> implementations do not alter the signature of
-these functions in any way. <a>DID resolver</a> implementations might map the
-<code>resolve</code> and <code>resolveRepresentation</code> functions to a
+these functions in any way. <a>DID resolver</a> implementations might map
+this function to a
```
peacekeeper, line 3817: Why this change?

index.html
```diff
@@ -3864,14 +3860,13 @@ <h3>DID Resolution Metadata Properties</h3>
...
-REQUIRED if resolution is successful and if the
-<code>resolveRepresentation</code> function was called. This property MUST NOT
+The MIME type of the returned <code>didDocumentStream</code>. This property is
+REQUIRED if resolution is successful. This property MUST NOT
 be present if the <code>resolve</code> function was called. The value of this
 property MUST be an <a data-lt="ascii string">ASCII string</a> that is the MIME
 type of the conformant <a>representations</a>. The
-caller of the <code>resolveRepresentation</code> function MUST use this value
-when determining how to parse and process the <code>did-document-stream</code>
+caller of <code>resolveRepresentation</code> function MUST use this value
```
peacekeeper, line 3868:
```suggestion
caller of the <code>resolveRepresentation</code> function MUST use this value
```

index.html
peacekeeper: If we want such a warning, I think it should go into the DID URL Syntax section, not here.

index.html
peacekeeper: This sentence should say MAY, it was incorrectly changed to SHOULD in https://github.com/w3c/did-core/commit/cbbca3b3.

peacekeeper (changes requested):

peacekeeper:
@msporny thanks for this new PR, I agree with a lot of changes, but am also confused by some others. Maybe it would be simpler to break this up into multiple PRs so they are easier to review?

- I agree with the clarification that says the didDocument (abstract data model) returned by resolve() can be transformed into a conforming DID document (representation). This solves the testability problem, right?
- I agree with the change from hyphens (did-document) to camelCase (didDocument).
- I agree with the at-risk markers
- I probably agree with renaming input metadata to options (but this could be done in a separate PR)
- I don't understand which bugs you are referring to, and I don't understand several smaller changes in the PR (see inline comments).

index.html
jricher: Both the `resolve*` functions take a DID, the `dereference` function takes a DID URL. This is one of the most important differences between these functions.

index.html
jricher: Both the `resolve*` functions take a DID, the `dereference` function takes a DID URL. This is one of the most important differences between these functions.

index.html
jricher: They're all in the same overall section, but resolution and dereferencing are two very distinct operations and so they're defined separately.

index.html
msporny: Agreed, this isn't going to change in this PR.

index.html
msporny: Same as above ... `did` is correct here.

index.html
msporny: It was removed when condensed down to one URL. I can add it back in since we're back to two functions now.

index.html
msporny: Yep, this is a bug, will fix.

index.html
msporny: This is a bug, will fix.

index.html
msporny: I'll put it back, this is a hold over from when I condensed down into a single function.

index.html
msporny: @OR13 does @peacekeeper's response answer your question?

index.html
msporny: Agreed with @jricher -- we're going to keep these two sections separate since they do two very different things.

With my Editor hat off -- I continue to assert that we don't need to define this behavior in a normative way -- the whole dereferencing section is weird -- it's the purview of software libraries that we don't need standards for. We don't standardize the software library interface that an XML parser should use to extract a subsection of an XML document (same for HTML documents). I don't know why we continue to insist that we do this for DID Documents.

index.html
msporny: @OR13, I'm taking that as non-blocking feedback that is editorial and that can be added at a future point in time (during CR).

index.html
```diff
@@ -3944,283 +3945,242 @@ <h4>updated</h4>
...
         <dd>
 If the <code>dereferencing</code> function was called and successful, this MUST
-contain a <a>resource</a> corresponding to the <a>DID URL</a>.
-The <code>content-stream</code> SHOULD be a <a>DID document</a> in one of the
-conformant <a>representations</a> obtained through
-the resolution process.
-
-If the dereferencing is unsuccessful, this value MUST be empty.
+contain a <a>resource</a> corresponding to the <a>DID URL</a>. The
+<code>contentStream</code> SHOULD be a <a>resource</a> such as a <a href="">DID
```
msporny, line 4168:
```suggestion
<code>contentStream</code> MAY be a <a>resource</a> such as a <a href="">DID
```

index.html
msporny, line 4171:
```suggestion
<a>representations</a> obtained through the resolution process 
or any other resource format that can be identified via a MIME type. If the
```

index.html
msporny, line 4216:
```suggestion
determine the <code>contentType</code> of the <a>representation</a> contained in the returned value if such a
```

index.html
```diff
@@ -3686,46 +3692,42 @@ <h1>Resolution</h1>
...
-document</a> by using the "Read" operation of the applicable <a>DID method</a>.
-(See <a href="#read-verify"></a>.) The details of how this process is
-accomplished are outside the scope of this specification, but all conformant
-implementations implement two functions which have the following abstract
-forms:
+The <a>DID resolution</a> function resolves a <a>DID</a> into a <a>DID
+document</a> by using the "Read" operation of the applicable <a>DID method</a>
+as described in <a href="#read-verify"></a>. The details of how this process is
+accomplished are outside the scope of this specification, but all conforming
+<a>DID resolvers</a> implement the functions below which have the following
```
TallTed, line 3713:
```suggestion
<a>DID resolvers</a> implement the functions below, which have the following
```

index.html
```diff
@@ -3686,46 +3692,42 @@ <h1>Resolution</h1>
...
-abstract form.
-The <code>resolveRepresentation</code> function returns a byte stream of the
-<a>DID Document</a> formatted in the corresponding representation.
-        </p>
-        <p>
-The input variables of these functions are as follows:
+abstract form. The <code>resolveRepresentation</code> function returns a byte
+stream of the <a>DID Document</a> formatted in the corresponding representation.
+The input variables of the <code>resolve</code> and
+<code>resolveRepresentation</code> functions are as follows:
```
TallTed, line 3730:
```suggestion
<code>resolveRepresentation</code> functions are as follow:
```
"{singular} is as follows" or "{plural} are as follow"

index.html
```diff
@@ -3737,55 +3739,50 @@ <h2>DID Resolution</h2>
...
             <dt>
-did-document
+<dfn>didDocument</dfn>
             </dt>
             <dd>
 If the resolution is successful, and if the <code>resolve</code> function was
-called, this MUST be a <a>conforming DID document</a>. If the resolution is
+called, this MUST be a <a>DID document</a> (abstract data model) as described in
+<a href="#data-model"></a> that is capable of being transformed
+into a <a>conforming DID Document</a> (representation), using the production rules specified by the representation. If the resolution is
```
TallTed, line 3781:
```suggestion
into a <a>conforming DID Document</a> (representation), using the production
rules specified by the representation. If the resolution is
```

index.html
jricher: This is not correct, the correct form is "as follows" regardless of the number of elements.

TallTed: Huh... The gurus of the web concur with you, and much younger me. I wish I'd had Google when in school...

index.html
```diff
@@ -4235,29 +4196,32 @@ <h2>DID URL Dereferencing</h2>
...
-        <h3>DID URL Dereferencing Input Metadata Properties</h3>
+        <h3>DID URL Dereferencing Options</h3>
 
         <p>
 The possible properties within this structure and their possible values SHOULD
 be registered in the DID Specification Registries [[?DID-SPEC-REGISTRIES]].
-This specification defines the following common properties.
+This specification defines the following common properties. This
+specification defines the following common properties for dereferencing
+options:
```
TallTed, line 4206:
```suggestion
This specification defines the following common properties for
dereferencing options:
```

index.html
```diff
@@ -4270,18 +4234,20 @@ <h3>DID URL Dereferencing Metadata Properties</h3>
...
 error
                 </dt>
                 <dd>
 The error code from the dereferencing process. This property is REQUIRED when
-there is an error in the dereferencing process. The value of this property is
-a single keyword string. The possible property values of this field SHOULD be
-registered in the DID Specification Registries [DID-SPEC-REGISTRIES]]. This
-specification defines the following error values:
+there is an error in the dereferencing process. The value of this property
+MUST be a single keyword expressed as an  <a data-lt="ascii string">ASCII
```
TallTed, line 4247:
```suggestion
MUST be a single keyword expressed as an <a data-lt="ascii string">ASCII
```

TallTed (changes requested):
LGTM after changes suggested above

index.html
msporny: Fixed in af3a4fdf150a5a1d636d23a82133553e287d581e.

index.html
msporny: Fixed in 5d96fb87147d2921ce5079c00433f6bfcdd7335e.

index.html
msporny: Fixed in ef9a57dd48a270306652ad8aa7beaaf92d3996aa.

index.html
msporny: Fixed in 8d9fe3b189d88915780fe8535fb82d7de441ff78.

index.html
```diff
@@ -3763,13 +3763,15 @@ <h2>DID Resolution</h2>
...
-REQUIRED and MUST NOT be empty. This metadata is defined by <a
-href="#did-resolution-metadata"></a>. If <code>resolveRepresentation</code> was
-called, this structure MUST contain a <code>contentType</code> property
-containing the MIME type of the <code>didDocument</code>. If the resolution is
-not successful, this structure MUST contain an <code>error</code> property
-describing the error.
+relating to the results of the <a>DID resolution</a> process and typically
+changes between invocations of the <code>resolve</code> and
+<code>resolveRepresentation</code> functions as it represents data about the
+resolution process itself. This structure is REQUIRED and MUST NOT be empty.
```
TallTed, line 3769:
```suggestion
relating to the results of the <a>DID resolution</a> process which typically
changes between invocations of the <code>resolve</code> and
<code>resolveRepresentation</code> functions, as it represents data about the
resolution process itself. This structure is REQUIRED, and MUST NOT be empty.
```

index.html
```diff
@@ -3737,55 +3739,53 @@ <h2>DID Resolution</h2>
...
         </dl>
 
         <p>
-The output variables of these functions are as follows:
+The return value of <code>resolve</code> is the
+<a>didResolutionMetadata</a>, <a>didDocument</a>, and
+<a>didDocumentMetadata</a> properties. The return value of
+<code>resolveRepresentation</code> is the
+<a>didResolutionMetadata</a>, <a>didDocumentStream</a>, and
+<a>didDocumentMetadata</a> properties. These properties are described below:
```
jricher, line 3757: These are not properties, which implies that they're part of a structure, they are separate return values. Trying to straighten that out here:

```suggestion
These functions each return multiple values, and no limitations
are placed on how these values are returned together.
The return values of <code>resolve</code> are
<a>didResolutionMetadata</a>, <a>didDocument</a>, and
<a>didDocumentMetadata</a>. The return values of
<code>resolveRepresentation</code> are
<a>didResolutionMetadata</a>, <a>didDocumentStream</a>, and
<a>didDocumentMetadata</a> properties. These values are described below:
```

index.html
```diff
@@ -3737,55 +3739,53 @@ <h2>DID Resolution</h2>
...
-an <code>error</code> property describing the error.
-              </p>
+relating to the results of the <a>DID resolution</a> process which typically
+changes between invocations of the <code>resolve</code> and
+<code>resolveRepresentation</code> functions, as it represents data about the
+resolution process itself. This structure is REQUIRED, and MUST NOT be empty.
+This metadata is defined by <a href="#did-resolution-metadata"></a>. If
+<code>resolveRepresentation</code> was called, this structure MUST contain a
+<code>contentType</code> property containing the MIME type of the
+<code>didDocument</code>. If the resolution is not successful, this structure
```
jricher, line 3773:
```suggestion
representation found in the <code>didDocumentStream</code>. If the resolution is not successful, this structure
```

index.html
```diff
@@ -3737,55 +3739,53 @@ <h2>DID Resolution</h2>
...
+MUST contain an <code>error</code> property describing the error.
             </dd>
             <dt>
-did-document
+<dfn>didDocument</dfn>
             </dt>
             <dd>
 If the resolution is successful, and if the <code>resolve</code> function was
-called, this MUST be a <a>conforming DID document</a>. If the resolution is
+called, this MUST be a <a>DID document</a> (abstract data model) as described in
```
jricher, line 3781:
```suggestion
called, this MUST be a <a>DID document</a> abstract data model (an INFRA map) as described in
```

index.html
```diff
@@ -3737,55 +3739,53 @@ <h2>DID Resolution</h2>
...
+<a href="#data-model"></a> that is capable of being transformed
+into a <a>conforming DID Document</a> (representation), using the production
+rules specified by the representation. If the resolution is
```
jricher, line 3784: I don't understand why this is requirement was added here. If I can get it back into the ADM, then the resolve is done and whether I can serialize it back out is irrelevant to this function.

```suggestion
<a href="#data-model"></a>. If the resolution is
```

index.html
```diff
@@ -3821,37 +3820,36 @@ <h2>DID Resolution</h2>
 method-specific internal function to perform the actual <a>DID resolution</a>
 process. <a>DID resolver</a> implementations might implement and expose
 additional functions with different signatures in addition to the
-<code>resolve</code> function specified here.
+<code>resolveRepresentation</code> function specified here.
```
jricher, line 3823:
```suggestion
<code>resolve</code> and <code>resolveRepresentation</code> functions specified here.
```

index.html
```diff
@@ -3821,37 +3820,36 @@ <h2>DID Resolution</h2>
...
-<code>did-document-stream</code> if such a <a>representation</a> is supported and
-available. This property is OPTIONAL. It is only used if the
-<code>resolveRepresentation</code> function is called and MUST be ignored if the
-<code>resolve</code> function is called.
+The MIME type of the caller's preferred <a>representation</a> of the <a>DID
+document</a>. The MIME type MUST be expressed as an  <a data-lt="ascii
+string">ASCII string</a>. The <a>DID resolver</a> implementation SHOULD use this
+value to determine the <a>representation</a> contained in the returned
+<code>didDocumentStream</code> if such a <a>representation</a> is supported and
+available. This property is OPTIONAL.
```
jricher, line 3846: Requesting a mime type makes no sense for an ADM resolution, which is why it was previously prohibited.

```suggestion
available. This property is OPTIONAL for the <code>resolveResolution</code> function
and MUST NOT be used with the <code>resolve</code> function.
```

index.html
```diff
@@ -3944,283 +3947,243 @@ <h4>updated</h4>
...
+of this specification, but all conforming <a>DID resolvers</a> implement
+the following function which has the following abstract form:
       </p>
 
-      <p><code>
-dereference ( did-url, did-url-dereferencing-input-metadata ) <br>
-&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;-&gt; ( did-url-dereferencing-metadata, content-stream, content-metadata )
-      </code></p>
+      <pre title="Abstract functions for DID Dereferencing">
+dereference(didUrl, options) →
```
jricher, line 4117:
```suggestion
dereference(didUrl, dereferenceOptions) →
```

index.html
```diff
@@ -3686,46 +3692,42 @@ <h1>Resolution</h1>
...
-        </code></p>
+        <pre title="Abstract functions for DID Resolution">
+resolve(did, options) →
+   « didResolutionMetadata, didDocument, didDocumentMetadata »
 
-        <p><code>
-resolveRepresentation ( did, did-resolution-input-metadata ) <br>
-&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;-&gt; ( did-resolution-metadata, did-document-stream, did-document-metadata )
-        </code></p>
+resolveRepresentation(did, options) →
```
jricher, line 3721:
```suggestion
resolveRepresentation(did, resolutionOptions) →
```

index.html
```diff
@@ -3686,46 +3692,42 @@ <h1>Resolution</h1>
...
+<a>DID resolvers</a> implement the functions below, which have the following
+abstract forms:
         </p>
 
-        <p><code>
-resolve ( did, did-resolution-input-metadata ) <br>
-&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;-&gt; ( did-resolution-metadata, did-document, did-document-metadata )
-        </code></p>
+        <pre title="Abstract functions for DID Resolution">
+resolve(did, options) →
```
jricher, line 3718:
```suggestion
resolve(did, resolutionOptions) →
```

index.html
```diff
@@ -3944,283 +3947,243 @@ <h4>updated</h4>
...
+didUrl
         </dt>
         <dd>
 A conformant <a>DID URL</a> as a single string. This is the <a>DID URL</a> to
 dereference. To dereference a <a>DID fragment</a>, the complete <a>DID URL</a>
 including the <a>DID fragment</a> MUST be used. This input is REQUIRED.
         </dd>
         <dt>
-did-url-dereferencing-input-metadata
+options
```
jricher, line 4135:
```suggestion
dereferencingOptions
```

index.html
jricher: Yes, the dereferencing function can return something other than the did document itself.

index.html
```diff
@@ -3944,283 +3947,243 @@ <h4>updated</h4>
...
+<code>didUrl</code> itself. Properties defined by this specification are in <a
+href="#did-url-dereferencing-options"></a>. This input is REQUIRED, but the
+structure MAY be empty.
         </dd>
       </dl>
 
       <p>
-The output variables of this function are as follows:
+The return value of the <code>dereference</code> function contains the
+following properties:
```
jricher, line 4148: Same comment on return as above:

```suggestion
This function returns multiple values, and no limitations
are placed on how these values are returned together.
The return values of the <code>dereference</code> include
<code>dereferencingMetadata</code>, <code>contentStream</code>,
and <code>contentMetadata</code>:
```

index.html
```diff
@@ -4318,24 +4285,26 @@ <h2>Metadata Structure</h2>
...
 <a data-cite="INFRA#null">null</a>. The values within any complex data
 structures such as maps and lists MUST be one of these data types as well.
-All metadata property definitions MUST define the value type, including any
+All metadata property definitions registered in the DID Specification
+Registries [[?DID-SPEC-REGISTRIES]] MUST define the value type, including any
 additional formats or restrictions to that value (for example, a string
 formatted as a date or as a decimal integer). It is RECOMMENDED that property
-definitions use strings for values.
+definitions use strings for values. The entire metadata structure MUST be
+serializable in a conforming <a>representation</a>.
```
jricher, line 4293: Why was this additional requirement added? The implementation might do any number of things to serialize the metadata separately from the document itself. As long as it gets in and out of the map, that's fine. By requiring the metadata to re-use the type and data structure system it inherits this property anyway.

```suggestion
definitions use strings for values.
```

index.html
```diff
@@ -3944,283 +3947,243 @@ <h4>updated</h4>
...
-the resolution process.
-
-If the dereferencing is unsuccessful, this value MUST be empty.
+contain a <a>resource</a> corresponding to the <a>DID URL</a>. The
+<code>contentStream</code> MAY be a <a>resource</a> such as a <a href="">DID
+Document</a>,  <a href="#verification-methods">Verification Method</a>,  or <a
+href="#services">service</a> that is serializable in one of the conformant
+<a>representations</a> obtained through the resolution process
+or any other resource format that can be identified via a MIME type. If the
+dereferencing is unsuccessful, this value MUST be empty.
```
jricher, line 4175: I think the additional examples were added in the wrong spot, as written this implies that a service output is in a DID Document Representation format.

```suggestion
contain a <a>resource</a> corresponding to the <a>DID URL</a>. The
<code>contentStream</code> MAY be a <a>resource</a> such as a <a href="">DID
Document</a> that is serializable in one of the conformant
<a>representations</a>,  a <a href="#verification-methods">Verification Method</a>,  or a <a
href="#services">service</a> obtained through the resolution process. If the
dereferencing is unsuccessful, this value MUST be empty.
```

jricher (changes requested):

index.html
OR13: agreed.

index.html
OR13: @msporny ^ has this comment been addressed?

index.html
```diff
@@ -3686,46 +3692,42 @@ <h1>Resolution</h1>
...
+URL dereferencing</a>.  Their exact implementation is out of scope for this
+specification, but some considerations for implementors are discussed in
+[[?DID-RESOLUTION]].
     </p>
 
     <p>
 All conformant <a>DID resolvers</a> MUST implement the <a>DID resolution</a>
-functions for at least one <a>DID method</a> and MUST be able to return a <a>DID
-document</a> in at least one conformant <a>representation</a>.
+functions for at least one <a>DID method</a> and MUST be able to produce a
```
peacekeeper, line 3702:
```suggestion
functions for at least one <a>DID method</a> and MUST be able to return a
```

Propose to leave this unchanged, to avoid confusion with "production" of representations.

index.html
```diff
@@ -3686,46 +3692,42 @@ <h1>Resolution</h1>
...
+    <section data-dfn-for="Resolver">
         <h2>DID Resolution</h2>
         <p>
-The <a>DID resolution</a> functions resolve a <a>DID</a> into a <a>DID
-document</a> by using the "Read" operation of the applicable <a>DID method</a>.
-(See <a href="#read-verify"></a>.) The details of how this process is
-accomplished are outside the scope of this specification, but all conformant
-implementations implement two functions which have the following abstract
-forms:
+The <a>DID resolution</a> function resolves a <a>DID</a> into a <a>DID
```
peacekeeper, line 3709:
```suggestion
The <a>DID resolution</a> functions resolve a <a>DID</a> into a <a>DID
```

index.html
peacekeeper: +1 to not using the term "properties" here. @jricher in your suggestion you removed "properties" in once sentence, but you forgot to also remove it in the other sentence.

index.html
```diff
@@ -3686,46 +3692,42 @@ <h1>Resolution</h1>
...
 
         <p>
 The <code>resolve</code> function returns the <a>DID document</a> in its
-abstract form.
-The <code>resolveRepresentation</code> function returns a byte stream of the
-<a>DID Document</a> formatted in the corresponding representation.
-        </p>
-        <p>
-The input variables of these functions are as follows:
+abstract form. The <code>resolveRepresentation</code> function returns a byte
```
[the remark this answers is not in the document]
OR13: hmm, so there is not a single method that returns concrete representations? I though that was the whole point of having 2 functions... are you saying they both return the ADM?

index.html
OR13: yes.

index.html
peacekeeper: @jricher my understanding is that it is useful to say this here in order to make the resolve() function testable (see the title of the PR). I agree with @msporny 's proposed addition here.

index.html
TallTed: @jricher - I think you missed deleting ` properties` from line 3759? (Someday, Github will figure out that suggestions might be needed on suggestions...)

index.html
```diff
@@ -3864,14 +3862,13 @@ <h3>DID Resolution Metadata Properties</h3>
 contentType
                 </dt>
                 <dd>
-The MIME type of the returned <code>did-document-stream</code>. This property is
-REQUIRED if resolution is successful and if the
-<code>resolveRepresentation</code> function was called. This property MUST NOT
+The MIME type of the returned <code>didDocumentStream</code>. This property is
+REQUIRED if resolution is successful. This property MUST NOT
```
peacekeeper, line 3866: Why this change?

peacekeeper (changes requested):

OR13 (changes requested):
@msporny I think my only remaining concern is this comment by @jricher 

https://github.com/w3c/did-core/pull/601/files#r573007873

It was my understanding that:

```
resolve -> DID Document: INFRA<MAP>
````

Whereas:

```
resolveRepresentation -> DID Document: ByteStream / resolution response object in `json`/`cbor`, etc...NOT the ADM.
````

We need at least one concrete resolution function that returns something testable...

As long agree to at least one testable resolution function which returns `json` / `cbor` I am good, I don't care what we call it.

**Comment by peacekeeper** [comment]
This PR is not easy to review.. There are many many changes, most of which have nothing to do with the title of the PR (make resolution concrete and testable).

File: index.html (hunk 40 above)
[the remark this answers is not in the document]
Comment (OR13) [reply index.html:3727]: I may be misinterpreting this change request... 

`resolveRepresentation` still returns json or cbor bytes correct?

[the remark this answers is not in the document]
Comment (jricher) [reply index.html:3727]: @OR13 `resolveRepresentation` returns a byte stream in a conformant format ... as stated literally on this line. :)

File: index.html (hunk 28 above)
Comment (msporny) [reply index.html:3784]: The language that was here before didn't say you go back to the ADM... it said that you go back to a representation. The reference to "conforming DID Document" was confusing, because conforming DID Documents are expected to be expressed in a representation... which was not the intent here. So, the change was made to make sure we were talking about a `didDocument` in the abstract sense... but made sure that was testable by saying you had to be able to take whatever that was and get it out to a representation (in order to make it clearly testable).

File: index.html (hunk 41 above)
Comment (msporny) [reply index.html:3869]: > Why this change?

```suggestion
REQUIRED if resolution is successful and if the
<code>resolveRepresentation</code> function was called. 
This property MUST NOT
```

index.html
msporny: Yes, the implementation might do any number of things -- but we can't test "any number of things" -- we need to test a "specific thing". The text was added to provide something specific we can test for... it doesn't say you have to serialize it... it says it "must be serializable" -- that is, you must be able to do it (and the test suite will test that). Implementations aren't required to serialize, just ensure that it is possible to do so.

index.html
msporny: The "or any other resource format that can be identified via a MIME type" is important, because of the whole `resource=true` angle... the thing that's returned could really be anything -- an image, a PDF file, etc.

index.html
msporny: Yes, it is a MAY now.

TallTed (approved):

index.html
msporny: Fixed a while ago.

msporny:
@OR13 wrote:
> As long as we can agree to at least one testable resolution function which returns json / cbor I am good, I don't care what we call it.

At present, I believe both `resolve` and `resolveRepresentation` are testable. `resolve` is now testable because we say you must be able to express `didDocument` (which is a map) in a representation. `resolveRepresentation` is testable because it returns a representation.

I believe I have addressed all of your concerns, @OR13 (and merged your change suggestions).

@rhiaro, @TallTed, and @peacekeeper, I believe I've addressed all of your concerns (and merged your change suggestions).

@jricher, I merged the majority of your change suggestions. There are three changes that I didn't apply with reasoning on why each one didn't go in. Please re-review and let us know if the current state of the PR works for you.

index.html
```diff
@@ -3821,37 +3822,37 @@ <h2>DID Resolution</h2>
 method-specific internal function to perform the actual <a>DID resolution</a>
 process. <a>DID resolver</a> implementations might implement and expose
 additional functions with different signatures in addition to the
-<code>resolve</code> function specified here.
+<code>resolve</code> and <code>resolveRepresentation</code> functions specified here.
```
kdenhartog, line 3825: non blocking editorial comment: Why mention that they can be implemented differently here? This seems to encourage extensions that could lead to divergence at the API level and ultimately cause interop concerns. I think this is best removed.

kdenhartog (approved):
One minor non-blocking comment. Looks good in my opinion.

index.html
jricher: It's actually saying the opposite of that: you can't implement `resolve` and `resolveRepresentation` with different signatures at all, you always have to use these signatures. However you can also add additional functions with different signatures and expose those to your callers -- and nothing we say in the spec would stop that, so it's best to clearly admit it here.

index.html
jricher: The change I have proposed does not limit the response at all, but we can add back the language about additional mime types in a more appropriate place. The problem I sought to fix above was that the expressed wording made it sound like a DID Document could come back as a MIME type other than a conformant representation, which we DON'T want.

```suggestion
contain a <a>resource</a> corresponding to the <a>DID URL</a>. The
<code>contentStream</code> MAY be a <a>resource</a> such as a <a href="">DID
Document</a> that is serializable in one of the conformant
<a>representations</a>,  a <a href="#verification-methods">Verification Method</a>,  a <a
href="#services">service</a>, or any other resource format that can be identified via a MIME type
and obtained through the resolution process. If the
dereferencing is unsuccessful, this value MUST be empty.
```

index.html
jricher: I do not agree that this makes it any more "testable" than before. The burden is on the test suite to serialize the result in either case, and a conformant DID document already needs to be representable. This requirement is redundant and implies that there are DID documents that could possibly :not: be representable in some form.

index.html
jricher: I do not agree that this makes it any more "testable" than before. The burden is on the test suite to serialize the result in either case, and a conformant DID document already needs to be representable. This requirement is redundant and implies that there are DID documents that could possibly :not: be representable in some form.

index.html
jricher: There is no such thing as "a conformant representation" for metadata. The representations are defined in terms of the did documents. This requirement does not make sense.

OR13: I thought we wanted to be able to return metadata as cbor, json, yaml, etc... are we saying we want to be able to return a json did document with cbor meta data?

My understanding was that the entire response values would be in the requested representation... so if you asked for json, you would not get back cbor... but perhaps you see this as way to return response data in http codes / headers... which will never be JSON / CBOR?

jricher: Metadata is not ever defined in terms of JSON, CBOR, or anything. It's a data structure (a map). 

The ability to call the `resolve` function for a specific platform is going to require some way to get the map values in and out. We specify this in terms of the data structure -- map -- and not in terms of a serialization for that data structure. The method of passing that data structure in and out is the binding of the abstract function to a specific implementation.

jricher: It sounds like what you're writing here is a requirement for hooking functions to the test harness, and not requirements for the function itself. This is a dangerous line to cross.

msporny: > There is no such thing as "a conformant representation" for metadata. 
> Metadata is not ever defined in terms of JSON, CBOR, or anything. It's a data structure (a map).

Here's the problem that we continue to have with the Resolution section:

The original authors intend an unspecified subset of it to be a purely abstract thing while simultaneously wanting there to be normative language related to the abstract concepts. While it is true that we can write normative statements about abstract concepts, it has been asserted that doing so is less useful than being able to have testable statements. This point has been made repeatedly throughout the past decade or more at W3C, across multiple Working Groups, with repeated consensus positions reaffirming that groups shouldn't write normative language around abstract concepts that are not realizable in concrete tests.

In other words, we cannot test abstract things in the test suite if they are never realized as something concrete. It is problematic to have normative language on abstract concepts because it's very easy for people to make subjective assertions about their conformance to those statements. When you don't make things testable, you harm interoperability.

We have three options that were explored:

1. Migrate the entire Resolution section into a NOTE.
2. Remove normative language that is associated with abstract concepts/statements. 
3. Make the normative language around abstract concepts/statements testable by providing a path to a concrete expression of those abstract concepts/statements.

The third item seemed to be the one that had the best chances of achieving consensus.

In order for that to happen, we need a proposal that enables us to get the metadata into a concrete form so the test suite can test it. One approach is to say that "the metadata structure MUST be serializable in a conforming representation". If we look at the JSON production rules, it doesn't say anything about production ONLY applying to DID Documents... it just states that "if you have data type X, then you can serialize by doing Y." ... and if we can do that with the metadata structure, then we have a path to testability.

I'm also open to any other paths to testability.

At this point, @jricher, we need a counter-proposal on how we get the metadata structure to testability, or we need you to clarify if you are objecting or if you can live with the language.

index.html
msporny: The change makes it explicit that the return value is in ADM form, and that it must be serializable in a representation. By doing so, we achieve what you want (which is for the return value to be abstract), but also enable the return value to be testable by the test suite (and therefore usable by an implementation).

@jricher, if you disagree with this change, we need a counter-proposal from you on how we get this particular return value to testability, or we need you to clarify if you are objecting or if you can live with the language.

index.html
jricher: I am not going to formally object to anything in here, but I am on record in saying that this is a bad idea that makes for a weak specification.

index.html
jricher: An alternative proposal to testability is to fix the problem in the test suite itself:

To connect to the test suite, and implementation has to define a concrete means for the test suite to call the `resolve`, `resolveResolution`, and `dereference` functions, including concrete means for supplying all inputs and receiving all outputs. 

This is text that does not belong in the specification proper (in my opinion) but makes it clear what harnesses the test suite will need to test implementations of these functions.

jricher: Another way to look at it is that this "testability" push is turning this functional definition section into a _protocol_. It is my belief that this is explicitly not a protocol, and as such calling these functions in an implementation might require a network call, writing files to a directory, linking to a library and calling a native function, running a utility script, or any number of possible ways to call a function. This is the problem that I have had with putting the burden on the test suite to test implementations of these functions in this way. If we were defining a protocol, the arguments around serializability and normativity would make sense, but as it is they do not and I believe this is a result of overreach from the test suite, which needs to define its own requirements for common testability, effectively creating a protocol on top of these functions for the purposes of the test suite.

msporny: > To connect to the test suite, and implementation has to define a concrete means for the test suite to call the resolve, resolveResolution, and dereference functions, including concrete means for supplying all inputs and receiving all outputs.

This is typically done by W3C Working Groups by defining the concrete API in the specification and writing a test suite against that concrete API. Implementers then expose that concrete API in their implementations:

https://w3c.github.io/FileAPI/#file-section

We are being far more lax with the test-ability of the Resolution section by doing a hand-wave over the API and saying that a serialization will be available to the test suite to test.

That said, your opinion is noted and there is a record of it.

index.html
msporny: In lieu of a counter-proposal that can achieve stronger consensus, I'm hearing that you won't object to keeping the language as it stands.

msporny:
> @jricher, I merged the majority of your change suggestions. There are three changes that I didn't apply with reasoning on why each one didn't go in. Please re-review and let us know if the current state of the PR works for you.

I believe we have discussed all of your concerns with the current PR @jricher, and while you think some of these changes weaken the specification and/or are unnecessary, you would not object to the PR if it was merged at this time. Can you please confirm this so I can merge the PR?

jricher:
@msporny My concerns have been noted and discussed, and I disagree with the conclusions and solutions to those concerns that are part of this PR. That said my opinion is just one of many in the working group, and so I am not attempting to block this pull request with just my own stance, regardless of how and why I disagree.

msporny:
> @msporny My concerns have been noted and discussed, and I disagree with the conclusions and solutions to those concerns that are part of this PR. That said my opinion is just one of many in the working group, and so I am not attempting to block this pull request with just my own stance, regardless of how and why I disagree.

Acknowledged, thank you. 

I am going to pull this PR in with the following caveat -- we still have a few weeks left to improve this section if anyone has a better idea on how to strike the abstract/testable balance. We're almost certainly headed for a 2nd PR... I've never seen a specification with 130+ testable normative statements get it right the first time. This PR addresses all the outstanding testability concerns with the Resolution section, but in a way that most everyone is not entirely happy with... so if anyone has any better ideas in the meantime, please feel free to raise a PR -- the worst that can happen is that the group decides to not take it before CR.

msporny:
Normative, multiple reviews via #587, #591, #592, and #601, changes requested and made, remaining objections are non-blocking, merging with caveat in https://github.com/w3c/did-core/pull/601#issuecomment-777092645.

PR #602, opened 2021-02-07
Add bbs examples

OR13:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/602.html" title="Last updated on Feb 7, 2021, 4:40 PM UTC (e291eae)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/602/59687a4...e291eae.html" title="Last updated on Feb 7, 2021, 4:40 PM UTC (e291eae)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -5595,6 +5595,71 @@ Model</a> for additional examples.
 }
     </pre>
 
+    <pre class="example"
+    title="Verifiable Credential linked to a bls12381 verification method">
+{  <span class="comment">// external (all terms in this example)</span>
+  "@context": [
+    "https://www.w3.org/2018/credentials/v1",
+    "https://w3id.org/security/bbs/v1",
+    {
+      "name": "https://schema.org/name",
+      "birthDate": "https://schema.org/birthDate"
+    }
+  ],
+  "id": "urn:uuid:c499e122-3ba9-4e95-8d4d-c0ebfcf8c51a",
+  "type": ["VerifiableCredential"],
+  "issuanceDate": "2021-02-07T16:02:08.571Z",
+  "issuer": {
+    "id": "did:example:123"
+  },
+  "credentialSubject": {
+    "id": "did:example:456",
+    "name": "John Smith",
+    "birthDate": "2021-02-07"
+  },
+  "proof": {
+    "type": "BbsBlsSignature2020",
+    "created": "2021-02-07T16:02:10Z",
+    "proofPurpose": "assertionMethod",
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -5595,6 +5595,71 @@ <h2>Proving</h2>
...
+  "issuer": {
+    "id": "did:example:123"
+  },
+  "credentialSubject": {
+    "id": "did:example:456",
+    "name": "John Smith",
+    "birthDate": "2021-02-07"
+  },
+  "proof": {
+    "type": "BbsBlsSignature2020",
```
msporny, line 5621:
```suggestion
    "type": "BbsBlsSignature2020",
```

Non-blocking feedback to @tplooker and @kdenhartog:

I'm wondering if we should rename this suite "SelectiveDisclosureSignature2020" -- that we're using BBS+ BLS is an implementation detail. For example, we may switch to STARK aggregation in SelectiveDisclosureSignature2025. We want the type name to give developers that know nothing about BBS+ BLS an idea of what the signature mechanism provides.

msporny: Also, as an aside -- we want to get some Ed25519Signature2020 examples in there too... but can wait until CR to do that since examples are editorial/non-normative.

OR13: hmm, yes... we need those examples asap... I can do a separate PR for them.

OR13:
@msporny can you approve or request changes on this PR?

msporny (approved):

index.html
kdenhartog: Let's stick with `BbsBlsSignature2020` for now in these examples, but I'm open to the suggestion of looking to move to `SelectiveDisclosureSignature2021` when we start making updates to a 2021 version. This way the examples are pointing to in use types and we can spend the time thinking a bit more about how best to name these suites in the mean time and continue the discussion in the LDP work.

kdenhartog (approved):
Thanks @OR13 for adding these!

tplooker (approved):

index.html
tplooker: Yeah we wont be changing the current (2020) suite name but are open to new naming conventions for future versions of the suite subject to community consensus. Just a note, if we do go down this path then a case could be made to rename suites like Ed25519Signature2020 which is also technically named rather than by function.

msporny: Acknowledged @tplooker and @kdenhartog -- just wanted to make sure everyone is keeping this in mind as we move forward.

msporny:
Editorial, multiple reviews, changes suggested and deferred, merging.

PR #610, opened 2021-02-09
Describe `@context` as JSON member in JSON-LD production/consumption.

peacekeeper:
Okay I know how controversial this will be, considering the discussions around my diagrams in https://github.com/w3c/did-core/pull/596 and https://github.com/w3c/did-core/pull/597), but I have to try.

This changes the language around `@context` in the JSON-LD production/consumption sections to treat it primarily as a member in a JSON object (concrete DID document representation), rather than an entry in the INFRA map (abstract DID document data model).

Explanation:

The JSON-LD production section as it's currently written seems to contradict itself. First it says

_"The DID document **MUST include the @context entry.**"_

but then it says

_"This entry [..] **could be present** [..]. **If the entry is present**" [..]._

I am pretty sure the intention here is to say that `@context` MUST be present in the JSON-LD representation, and it can be preserved as an entry in the abstract data model, but it is definitely NOT a REQUIRED entry in the abstract data model (it's not a Core Property). This PR clarifies this.

As an additional indication to support this, look at the examples in this section. They are JSON examples, not INFRA examples.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/610.html" title="Last updated on Feb 10, 2021, 10:01 AM UTC (aa2a17b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/610/6645c05...aa2a17b.html" title="Last updated on Feb 10, 2021, 10:01 AM UTC (aa2a17b)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2859,8 +2859,8 @@ require them.
         <p>
 The <a>DID document</a> MUST be serialized according to the
 <a href="#production">production rules for JSON</a>, with one additional
-requirement: The <a>DID document</a> MUST include the <code>@context</code>
-entry.
+requirement: The JSON object that represents the <a>DID document</a>
+MUST include the <code>@context</code> member.
         </p>
 
         <dl>
@@ -2869,8 +2869,8 @@ entry.
 
             <p>
 The <a href="https://www.w3.org/TR/json-ld11/#the-context">JSON-LD
-specification</a> defines values that are valid for this entry. This entry
-contains representation-specific syntax and therefore could be present in the <a
+specification</a> defines values that are valid for this member. It
+contains representation-specific syntax and therefore could be present as an entry in the <a
 href="#data-model">data model</a> to aid in lossless conversion. If the entry
 is present in the <a href="#data-model">data model</a>, it MUST be used during
 production unless either an alternative <code>@context</code> value is
@@ -2884,7 +2884,7 @@ The value of <code>@context</code> MUST be exactly one of these values.
 
             <ul>
               <li>
-The <a href="https://infra.spec.whatwg.org/#strings">INFRA string</a>
+The <a href="https://tools.ietf.org/html/rfc8259#section-7">JSON String</a>
 <code>https://www.w3.org/ns/did/v1</code>.
                 <pre class="example">
 {
@@ -2894,11 +2894,11 @@ The <a href="https://infra.spec.whatwg.org/#strings">INFRA string</a>
                 </pre>
               </li>
               <li>
-An <a data-cite="INFRA#ordered-set">INFRA ordered set</a>,
-with first item the <a href="https://infra.spec.whatwg.org/#strings">INFRA
-string</a> <code>https://www.w3.org/ns/did/v1</code>, and subsequent items of
-type <a href="https://infra.spec.whatwg.org/#strings">INFRA string</a> or
-<a href="https://infra.spec.whatwg.org/#maps">INFRA map</a>.
+A <a href="https://tools.ietf.org/html/rfc8259#section-5">JSON Array</a>,
+with first item the <a href="https://tools.ietf.org/html/rfc8259#section-7">JSON
+String</a> <code>https://www.w3.org/ns/did/v1</code>, and subsequent items of
+type <a href="https://tools.ietf.org/html/rfc8259#section-7">JSON String</a> or
+<a href="https://tools.ietf.org/html/rfc8259#section-4">JSON Object</a>.
                 <pre class="example">
 {
   "@context": [
@@ -2912,9 +2912,9 @@ type <a href="https://infra.spec.whatwg.org/#strings">INFRA string</a> or
             </ul>
 
             <p>
-All members of the <code>@context</code> property SHOULD exist in the DID
+All JSON-LD terms defined by the <code>@context</code> SHOULD exist in the DID
 Specification Registries [[?DID-SPEC-REGISTRIES]] in order to achieve
-interoperability across different <a>representations</a>. If a member does not exist
+interoperability across different <a>representations</a>. If a term does not exist
 in the DID Specification Registries, then the <a>DID document</a> might not be
 interoperable across <a>representations</a>.
             </p>
@@ -2942,15 +2942,15 @@ defined via the <code>@context</code>. Properties that are not defined via the
         <p>
 The <a>DID document</a> MUST be deserialized as a JSON document according to
 the <a href="#consumption">consumption rules for JSON</a>, with one additional
-requirement: The <a>DID document</a> MUST include the <code>@context</code>
-entry and be processed according to the rules below.
+requirement: The JSON object that represents the <a>DID document</a> MUST include the <code>@context</code>
+member and be processed according to the rules below.
         </p>
 
         <dl>
           <dt>@context</dt>
           <dd>
             <p>
-The value of the <code>@context</code> entry conforms to the
+The value of the <code>@context</code> member conforms to the
 <a href="#production-0">JSON-LD Production Rules</a>. If more than one
 <a>URI</a> is provided, the <a>URIs</a> MUST be interpreted as an
 <a data-cite="INFRA#ordered-set">ordered set</a>.
--- a/index.html
+++ b/index.html
@@ -2859,8 +2859,8 @@ require them.
         <p>
 The <a>DID document</a> MUST be serialized according to the
 <a href="#production">production rules for JSON</a>, with one additional
-requirement: The JSON object that represents the <a>DID document</a>
-MUST include the <code>@context</code> member.
+requirement: The serialized <a>DID document</a> MUST include the
+<code>@context</code> member.
         </p>
 
         <dl>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -2859,8 +2859,8 @@ <h3>Production</h3>
         <p>
 The <a>DID document</a> MUST be serialized according to the
 <a href="#production">production rules for JSON</a>, with one additional
-requirement: The <a>DID document</a> MUST include the <code>@context</code>
-entry.
+requirement: The JSON object that represents the <a>DID document</a>
+MUST include the <code>@context</code> member.
```
brentzundel, line 2863:
```suggestion
requirement: The serialized <a>DID document</a> MUST include the
<code>@context</code> member.
```

index.html
```diff
@@ -2869,8 +2869,8 @@ <h3>Production</h3>
 
             <p>
 The <a href="https://www.w3.org/TR/json-ld11/#the-context">JSON-LD
-specification</a> defines values that are valid for this entry. This entry
-contains representation-specific syntax and therefore could be present in the <a
+specification</a> defines values that are valid for this member. It
+contains representation-specific syntax and therefore could be present as an entry in the <a
```
brentzundel, line 2873:
```suggestion
contains representation-specific syntax and therefore could be present as an
entry in the <a
```

index.html
```diff
@@ -2942,15 +2942,15 @@ <h3>Consumption</h3>
         <p>
 The <a>DID document</a> MUST be deserialized as a JSON document according to
 the <a href="#consumption">consumption rules for JSON</a>, with one additional
-requirement: The <a>DID document</a> MUST include the <code>@context</code>
-entry and be processed according to the rules below.
+requirement: The JSON object that represents the <a>DID document</a> MUST include the <code>@context</code>
+member and be processed according to the rules below.
```
brentzundel, line 2946:
```suggestion
requirement: The serialized <a>DID document</a> MUST include the
<code>@context</code> member and be processed according to the rules below.
```

brentzundel (commented):
Some suggestions that try to keep your intent while (I hope) being more readable.

msporny:
I'm not sure why this would be controversial unless I'm missing some nuance... it fixes a few bugs in the JSON/JSON-LD production rules. That it's not controversial makes me think that it's not addressing the point of contention.

In other words... where does `@context` come from? IIRC, there is something that is not the ADM that produces it... but people are free to dump it into the ADM if they want to... especially the JSON folks, because they can't differentiate between `@context` and `@foobar`... treating them the same, by preserving them in the ADM.

iherman:
> I'm not sure why this would be controversial unless I'm missing some nuance... it fixes a few bugs in the JSON/JSON-LD production rules. That it's not controversial makes me think that it's not addressing the point of contention.

@msporny I believe the controversy comes from the reactions in  #596 and #597 which seem to say that the ADM _MUST_ include the `@context` values. Obviously, the `@context` _MUST_ be in the serialization if this serialization is in JSON-LD, but that should be the only _MUST_ in this respect. Methods may decide to store a context, may decide to pick up the value from the registry for non-core terms, or ignore them altogether because that particular method works only in JSON or in CBOR. How this is done is implementation dependent.

The [latest commit](https://github.com/w3c/did-core/pull/610/commits/aa2a17b8df3f8dcc96bfa5ed3c36f2e691080cc3) goes in this direction indeed.

peacekeeper:
I agree with @iherman 's comment. If there's agreement that the ADM does not HAVE to contain `@context` in order for production to work, then maybe there's less controversy than I thought.

jricher (approved):

OR13 (approved):
The spirit of this PR overturns hard fought consensus reached after many calls, and years worth of work... 

The reasons we agreed to preserving unregistered properties was to maximize extensibility via JSON... 

This PR goes against the spirit of the JSON representation and the JSON-LD representation... the whole point of the ADM is that it is capable of representing INFRA Map structures... wether they be things like `authorization` or `@context`... 

However.... I like most of the language changes, and since this PR does not block `application/did+json` from containing `$schema` or `@context`... I am fine with it... I think....

However... `@context` IS PRESENT in the Abstract Data Model when consuming a DID Document which contained it... and it is preserved.

This PR does nothing to change that, and as such `@context` and `$schema` will be ignored by `application/did+json`... and can be present in any JSON variant that allows for JSON... so the idea that this PR somehow "changes things" is not correct... all this PR does is use JSON instead of INFRA to define a JSON format....

brentzundel (approved):

dlongley (approved):

index.html
```diff
@@ -2859,8 +2859,8 @@ <h3>Production</h3>
         <p>
 The <a>DID document</a> MUST be serialized according to the
 <a href="#production">production rules for JSON</a>, with one additional
-requirement: The <a>DID document</a> MUST include the <code>@context</code>
-entry.
+requirement: The serialized <a>DID document</a> MUST include the
```
OR13, line 2862: Can we debate these changes in https://github.com/w3c/did-core/pull/626 and address the other production rules by themselves.

index.html
```diff
@@ -2942,15 +2943,15 @@ <h3>Consumption</h3>
         <p>
 The <a>DID document</a> MUST be deserialized as a JSON document according to
 the <a href="#consumption">consumption rules for JSON</a>, with one additional
-requirement: The <a>DID document</a> MUST include the <code>@context</code>
-entry and be processed according to the rules below.
+requirement: The serialized <a>DID document</a> MUST include the
```
OR13, line 2946: I am fine with this here.

index.html
```diff
@@ -2869,8 +2869,9 @@ <h3>Production</h3>
 
             <p>
 The <a href="https://www.w3.org/TR/json-ld11/#the-context">JSON-LD
-specification</a> defines values that are valid for this entry. This entry
-contains representation-specific syntax and therefore could be present in the <a
+specification</a> defines values that are valid for this member. It
+contains representation-specific syntax and therefore could be present as an
+entry in the <a
 href="#data-model">data model</a> to aid in lossless conversion. If the entry
```
OR13, line 2875: I addressed some of this problematic language here: https://github.com/w3c/did-core/pull/626

I believe it's compatible with this PR.

iherman:
> 
> In other words... where does `@context` come from? IIRC, there is something that is not the ADM that produces it... but people are free to dump it into the ADM if they want to... especially the JSON folks, because they can't differentiate between `@context` and `@foobar`... treating them the same, by preserving them in the ADM.

Strictly speaking, that is not true. The end of [§6.2.2](https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/610.html#consumption) (on JSON consumption) has an extra note:

> Note that the `@context` object member, if present, will not have additional processing applied to its value, which will be added verbatim to the data model.

Not sure it makes all that big difference, but clearly the spec gives a special role to `@context` already.

index.html
msporny, line 2862: This doesn't say where `@context` comes from. I don't think that's a big deal, JSON-LD folks will know to inject it if it doesn't exist in the ADM. However, there are cases where if you start with JSON-only AND don't include a `@context` AND want to be able to serialize to JSON-LD, it is possible to create a situation where you cannot serialize to JSON-LD (because you cannot know what the original intent of all of the terms were if `@context` was removed). Consider the following:

1. Assume three terms in two different contexts
2. termA in contextA, and termB in contextB
3. contextA and contextB each contain termC
4. contextA establishes termC to mean 'source'
5. contextB establishes termC to mean 'destination'
6. termA and termB are registered in DID Spec Registries
7. termC is not registered in DID Spec Registries
8. You have a DID Document that contains termA, termB, and termC

At this point, it's impossible to deterministically construct the `@context` for the JSON-LD representation using only the knowledge above. It is, therefore, impossible to use the DID Spec Registries to recreate `@context` for all valid DID Documents.

Folks in the WG seem to continue to be confused by this, thinking that the DID Spec registries can be used to recreate the `@context` field.

msporny (approved):

msporny:
Editorial, multiple reviews, changes requested and made, non-blocking concerns raised, merging.

index.html
peacekeeper: I think in 99% of the DID documents we see today, adding the `@context` during JSON-LD production is trivial. One of the reasons why the DID Spec Registries were created was to help with constructing the `@context` if you don't have it (e.g. if you have an `ethereumAddress` property in the DID document). I do agree that this is not always possible if the DID document uses unregistered extensions.

OR13: more importantly vanilla json does not care about the registries.... so you can very easily do this:

```ts
const didDocument = consume(json);
didDocument.foobarbaz = 123;
didDocument.produce(json-ld);
```

One of 2 things should happen...

1.    an error should be thrown `@context` is not defined.
2.  `didDocument.produce(json-ld, { `@context`, `foobarbaz: https://exmaple.com/open-world-assumption`)`;

IMO its always better to throw an error than silently "fix things" for developers... explicit over implicit is almost always better.

throwing an error teaches the developer how to do things correctly...

PR #611, opened 2021-02-10
Editorial rework of Introduction section.

msporny:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/611.html" title="Last updated on Feb 12, 2021, 3:01 PM UTC (3fe2caa)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/611/6645c05...3fe2caa.html" title="Last updated on Feb 12, 2021, 3:01 PM UTC (3fe2caa)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -249,78 +249,73 @@ As individuals and organizations, many of us use globally unique identifiers in
 a wide variety of contexts. They serve as communications addresses (telephone
 numbers, email addresses, usernames on social media), ID numbers (for passports,
 drivers licenses, tax IDs, health insurance), and product identifiers (serial
-numbers, barcodes, RFIDs). Resources on the Internet are identified by globally
-unique identifiers in the form of <abbr title="Media Access Control">MAC</abbr>
-addresses; URIs (Uniform Resource Identifiers) are used for resources on the Web
-and each web page you view in a browser has a globally unique URL (Uniform
-Resource Locator).
+numbers, barcodes, RFIDs). URIs (Uniform Resource Identifiers) are used for
+resources on the Web and each web page you view in a browser has a globally
+unique URL (Uniform Resource Locator).
     </p>
     <p>
 The vast majority of these globally unique identifiers are not under our
 control. They are issued by external authorities that decide who or what they
 identify and when they can be revoked. They are useful only in certain contexts
-and recognized only by certain bodies (not of our choosing). They may disappear
-or cease to be valid with the failure of an organization. They may unnecessarily
-reveal personal information. And in many cases they can be fraudulently
-replicated and asserted by a malicious third-party ("identity theft").
+and recognized only by certain bodies not of our choosing. They might
+disappear or cease to be valid with the failure of an organization. They might
+unnecessarily reveal personal information. In many cases, they can be
+fraudulently replicated and asserted by a malicious third-party, which is
+more commonly known as "identity theft".
     </p>
     <p>
 The Decentralized Identifiers (DIDs) defined in this specification are a new
-type of globally unique identifier designed to enable individuals and
-organizations to generate our own identifiers using systems we trust, and to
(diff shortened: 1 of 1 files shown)


OR13 (approved):

index.html
```diff
@@ -249,79 +249,74 @@ <h1>Introduction</h1>
...
+more commonly known as "identity theft".
     </p>
     <p>
 The Decentralized Identifiers (DIDs) defined in this specification are a new
-type of globally unique identifier designed to enable individuals and
-organizations to generate our own identifiers using systems we trust, and to
-prove control of those identifiers (authenticate) using cryptographic proofs
-(for example, digital signatures).
+type of globally unique identifier. They are designed to enable individuals and
+organizations to generate our own identifiers using systems we trust. These
```
brentzundel, line 269:
```suggestion
organizations to generate their own identifiers using systems they trust. These
```

index.html
```diff
@@ -249,79 +249,74 @@ <h1>Introduction</h1>
...
+new identifiers enable us to prove control of over them by authenticating
+using cryptographic proofs such as digital signatures.
```
brentzundel, line 271:
```suggestion
new identifiers enable users to prove control of over them by authenticating
using cryptographic proofs such as digital signatures.
```

index.html
```diff
@@ -249,79 +249,74 @@ <h1>Introduction</h1>
...
-personal or private data should be revealed, and without depending on a central
-authority to guarantee the continued existence of the identifier.
+Since we control the generation and assertion of Decentralized Identifiers, each
+of us can have as many DIDs as we need to respect our desired separation of
+identities, personas, and interactions. We can scope the use of these
+identifiers to the most appropriate contexts. We can interact with other people,
+institutions or systems that require us to identify ourselves, or things we
+control, while maintaining control over how much personal or private data should
+be revealed, and without depending on a central authority to guarantee the
+continued existence of the identifier.
```
brentzundel, line 281:
```suggestion
Since the generation and assertion of Decentralized Identifiers is
user-controlled, each user can have as many DIDs as they need to respect a
desired separation of identities, personas, and interactions. The use of these
identifiers can be scoped appropriately to different contexts. They support
interactions with other people, institutions, or systems that require users to
identify themselves, or things they control, while providing control over how
much personal or private data should be revealed, all without depending on a
central authority to guarantee the continued existence of the identifier.
```

brentzundel (commented):
I wasn't a fan of the first-person POV, so I've made suggestions that change it.

index.html
rhiaro: I agree with not using the first person, but would prefer to avoid "users" when we could say things like "people" or more specific roles.

index.html
```diff
@@ -249,79 +249,74 @@ <h1>Introduction</h1>
...
+      </li>
+
+      <li>
+Software developers that want to produce and consume Decentralized Identifiers
+and their associated data formats;
       </li>
       <li>
-Developers who want to enable their systems to accept user-controlled
-identifiers (consumers of DIDs);
+Systems integrators that want to understand how to utilize Decentralized
```
rhiaro, line 311:
```suggestion
Systems integrators that want to understand how to use Decentralized
```

rhiaro (approved):

index.html
brentzundel: entities?

index.html
TallTed: In line 275, here, I'd change `as many DIDs as they need to respect a` to `as many DIDs as necessary to maintain their`...

index.html
msporny: Updated with "entities".

index.html
msporny: Changed user->entity. Applied @TallTed's suggestion.

msporny:

PR #626, opened 2021-02-12
Update jsonld-production rules

OR13:
- Clarify that `@context` might be present in the ADM as a result of consumption (weaken implication that it might be there for other reasons)
- Clarify that `@context` MUST be added to the ADM prior to serialization of `application/did+ld+json` (explain where it comes from without requiring it to always be present in the ADM)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/626.html" title="Last updated on Feb 12, 2021, 1:33 AM UTC (2d51131)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/626/1dda7f5...2d51131.html" title="Last updated on Feb 12, 2021, 1:33 AM UTC (2d51131)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2891,13 +2891,11 @@ entry.
             <p>
 The <a href="https://www.w3.org/TR/json-ld11/#the-context">JSON-LD
 specification</a> defines values that are valid for this entry. This entry
-contains representation-specific syntax and therefore could be present in the <a
-href="#data-model">data model</a> to aid in lossless conversion. If the entry
-is present in the <a href="#data-model">data model</a>, it MUST be used during
-production unless either an alternative <code>@context</code> value is
-explicitly provided to the producer or if the value from the data model is not
-valid according to the consumption rules.
-            </p>
+contains representation-specific syntax that could be present in the <a
+href="#data-model">data model</a> as a result of consuming <code>application/did+ld+json</code>. 
+If absent, this entry MUST be added to the <a href="#data-model">data model</a>, 
+prior to serialization of <code>application/did+ld+json</code>. 
+          </p>
 
             <p>
 The value of <code>@context</code> MUST be exactly one of these values.


msporny (approved):

index.html
```diff
@@ -2891,13 +2891,11 @@ <h3>Production</h3>
...
-href="#data-model">data model</a> to aid in lossless conversion. If the entry
-is present in the <a href="#data-model">data model</a>, it MUST be used during
-production unless either an alternative <code>@context</code> value is
-explicitly provided to the producer or if the value from the data model is not
-valid according to the consumption rules.
-            </p>
+contains representation-specific syntax that could be present in the <a
+href="#data-model">data model</a> as a result of consuming <code>application/did+ld+json</code>. 
+If absent, this entry MUST be added to the <a href="#data-model">data model</a>, 
+prior to serialization of <code>application/did+ld+json</code>.
```
peacekeeper, line 2897:
```suggestion
If absent, this member MUST be added to the the serialized <a>DID document</a>, 
during serialization of <code>application/did+ld+json</code>. 
```

This makes your PR compatible with https://github.com/w3c/did-core/pull/610 (clarify that `@context` is a required member in the produced JSON document, but it's not a required entry in the abstract data model).

peacekeeper (changes requested):

index.html
iherman: +1 to that.

iherman (approved):

index.html
```diff
@@ -2891,13 +2891,11 @@ <h3>Production</h3>
             <p>
 The <a href="https://www.w3.org/TR/json-ld11/#the-context">JSON-LD
 specification</a> defines values that are valid for this entry. This entry
-contains representation-specific syntax and therefore could be present in the <a
-href="#data-model">data model</a> to aid in lossless conversion. If the entry
-is present in the <a href="#data-model">data model</a>, it MUST be used during
-production unless either an alternative <code>@context</code> value is
-explicitly provided to the producer or if the value from the data model is not
-valid according to the consumption rules.
```
jricher, line 2899: Why remove the requirement to check for validity of the property? Without this, a conforming resolver would be susceptible to injection of bad context values that it would blindly pass through. This seems like a gaping security and interoperability hole.

index.html
OR13: sorry, but this won't work, and creates confusion when you consider consumption... where `@context` is a member of the ADM... see the JSON production rules... you are treating an `entry` that is listed already as `representation-specific`, that is preserved by consumption... in a way that is incompatible with working group consensus.

If `during serialization` means I get to check if the context defines all terms and throw errors until the context is updated to define all of them... or if production for JSON-LD is defined like this:

```
produce(adm, context) -> json
```

I can actually live with this change... it creates the following confusion....

```
consume_jsonld(json)-> produce_json(adm) -> `@context` is preserved
consume_json(json)-> produce_jsonld(adm, context) -> `@context` is preserved
consume_json(json)-> produce_json(adm) -> no at context
```

It was my understanding that we agreed that `@context` was representation specific syntax of the ADM... it appears that maybe we did not have consensus on that... the current production language says that it is.

I find it fairly hilarious (in a bad spec way) that consuming jsonld preserves `@context` in the ADM, but somehow production requires it to come from another location.

It would seem logical that consumption would preserve it in the same abstract location that production requires it... if thats no in the abstract data model... we should have a special topic call to discuss where that is.

index.html
OR13: @jricher you want `or if the value from the data model is not valid according to the consumption rules.`
to be added back? that would require it to be in the data model... perhaps you would prefer that the context provided to the json-ld production must be valid according to the consumption rules?

I

index.html
OR13: I guess another way of expressing my concern is `during serialization` seems to be his WG defining a serialization process for JSON-LD that is different than JSON... Its also confusing since "serialization" usually means capturing data from a data model in a representation... implying that somehow JSON-LD does not view `@context` as part of its data model... which is clearly false (its a required field).

OR13: See https://github.com/transmute-industries/did-core I prefer to keep production a pure function of the ADM.... This means you need to do some "setup" to produce JSON-LD... which the other representations don't have, because they don't have any schema validation!

peacekeeper: We can't have a situation where representations put constraints on the ADM, which is what you are suggesting by effectively saying "JSON-LD production only works if the ADM contains `@context`". This was definitely never WG consensus, and it contradicts the concept of an **abstract** data model.

We all agree that individual representations should be able to make use of whatever extensions they want. JSON-LD DID documents can contain as many contexts as they want, JSON DID documents can include "foo" properties that are preserved, CBOR DID document can contain integer keys, etc. But all representation-specific features must remain encapsulated by the representation, and not impose requirements on the ADM.

If we allow this, then we will end up in a situation where you will need different contents in the ADM, depending on what representation you want. This is a layer violation, may make support for future representations impossible, and defeats the whole purpose of the data model design.

msporny:
@OR13 -- I don't know what to do with respect to this PR and PR #654. I think you approved of #654 -- don't know if that includes the changes you wanted from this PR?

OR13:
I no longer see the need for these changes.

PR #627, opened 2021-02-12
Fix editorial/normative issues in DID Parameters.

msporny:
Mostly editorial clean ups. Two normative REQUIRED statements around `service` and `relativeRef` were downgraded to OPTIONAL because keeping them at REQUIRED would make the `did:key` DID method invalid -- `did:key` has no support for `service` endpoints or `relativeRef`.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/627.html" title="Last updated on Feb 16, 2021, 1:50 PM UTC (56c0bff)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/627/1dda7f5...56c0bff.html" title="Last updated on Feb 16, 2021, 1:50 PM UTC (56c0bff)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -918,18 +918,24 @@ representations in Section <a href="#application-did-ld-json"></a>.
         <h2>DID Parameters</h2>
 
         <p>
-The <a>DID URL</a> syntax supports a simple format for parameters
-based on the <code>query</code> component (See <a href="#query"></a>). Adding
-a DID parameter to a <a>DID URL</a> means that the parameter becomes part of
-the identifier for a <a>resource</a>.
+The <a>DID URL</a> syntax supports a simple format for parameters based on the
+<code>query</code> component described in <a href="#query"></a>. Adding a DID
+parameter to a <a>DID URL</a> means that the parameter becomes part of the
+identifier for a <a>resource</a>.
         </p>
+
+        <pre class="example nohighlight"
+          title="A DID URL with a 'versionTime' DID parameter">
+did:example:123?versionTime=2021-05-10T17:00:00Z
+        </pre>
+
         <p>
 Some DID parameters are completely independent of of any specific <a>DID
-method</a>, and function the same way for all <a>DIDs</a>. Other DID
-parameters are not necessarily supported by all <a>DID methods</a>. Where
-optional parameters are supported, they are expected to operate uniformly
-across the <a>DID methods</a> that do support them. Requirements which enable
-this are detailed in the following table.
+method</a> and function the same way for all <a>DIDs</a>. Other DID parameters
+are not supported by all <a>DID methods</a>. Where optional parameters are
+supported, they are expected to operate uniformly across the <a>DID methods</a>
+that do support them. The following table provides common DID parameters
+that function the same way across all <a>DID methods</a>.
         </p>
 
         <table class="simple">
@@ -951,8 +957,8 @@ Description
               </td>
               <td>
 Identifies a service from the <a>DID document</a> by service ID. Support for
-this parameter is REQUIRED. The associated value MUST be an <a data-lt="ascii
-string">ASCII string</a>.
+this parameter is OPTIONAL. If present, the associated value MUST be an <a
+data-lt="ascii string">ASCII string</a>.
               </td>
             </tr>
             <tr>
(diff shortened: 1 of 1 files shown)


peacekeeper:
Strictly speaking, I don't think having those parameters REQUIRED makes did:key invalid. You can construct a valid DID URL such as:

`did:key:z6Mkfriq1MqLBoPWecGoDLjguo1sB9brj6wT3qZ5BxkKpuP6?service=agent&relativeRef=/credentials`

and a DID URL dereferencer would totatlly be able to process it. The functionality associated with this parameter is independent of the DID method. In the case of `did:key`, it would just always happen to return a **notFound** error, because there's no service in the DID document, but that doesn't mean the DID method is invalid.

Anyway, I would still be fine with changing REQUIRED to OPTIONAL, no objection...

But could you maybe add the example back in? Or is there a reason for deleting it?

OR13 (approved):

brentzundel (approved):

index.html
```diff
@@ -1013,45 +1021,24 @@ <h2>DID Parameters</h2>
         <p>
 Implementers as well as <a>DID method</a> specification authors might use
 additional DID parameters that are not listed here. For maximum
-interoperability, it is RECOMMENDED that DID parameters use the official W3C
-DID Specification Registries mechanism [[?DID-SPEC-REGISTRIES]], to avoid
-collision with other uses of the same DID parameter with different semantics.
+interoperability, it is RECOMMENDED that DID parameters use the official W3C DID
```
rhiaro, line 1024:
```suggestion
interoperability, it is RECOMMENDED that DID parameters use the DID
```

index.html
```diff
@@ -1013,45 +1021,24 @@ <h2>DID Parameters</h2>
...
+Specification Registries mechanism [[?DID-SPEC-REGISTRIES]], to avoid collision
+with other uses of the same DID parameter with different semantics.
         </p>
 
         <p>
 DID parameters might be used if there is a clear use case where the parameter
-needs to be part of a <a>URI</a> that can be used as a link, or as a <a>resource</a>
-in RDF / JSON-LD documents.
+needs to be part of a <a>URL</a> that references a <a>resource</a> with more
+precision than merely using the <a>DID</a>. It is expected that DID parameters
```
rhiaro, line 1032:
```suggestion
precision than using the <a>DID</a> alone. It is expected that DID parameters
```

index.html
```diff
@@ -1013,45 +1021,24 @@ <h2>DID Parameters</h2>
...
+will <em>not</em> be used if the same functionality can be expressed by passing
```
rhiaro, line 1033:
```suggestion
are <em>not</em> used if the same functionality can be expressed by passing
```

rhiaro (approved):

msporny:
> Strictly speaking, I don't think having those parameters REQUIRED makes did:key invalid. You can construct a valid DID URL such as:
> 
> did:key:z6Mkfriq1MqLBoPWecGoDLjguo1sB9brj6wT3qZ5BxkKpuP6?service=agent&relativeRef=/credentials
>
> and a DID URL dereferencer would totatlly be able to process it. 

That line of reasoning would mean that we could make *every* DID Parameter REQUIRED. It also depends on the meaning of "REQUIRED"... which typically means that you must include specific code for the feature. To say that did:key MUST support `service` by stating that it doesn't support it, is strange. There is no such thing as a service property in a did:key document, or relative references for that matter. Weakening the requirement to OPTIONAL doesn't create any issues that I can think of... and therefore, we can weaken it and address all concerns.

> The functionality associated with this parameter is independent of the DID method. In the case of did:key, it would just always happen to return a notFound error, because there's no service in the DID document, but that doesn't mean the DID method is invalid.

It's sounding like the requirement is on DID URL dereferencers, not the DID syntax (which is what this section is about). A future PR should probably make that distinction.

> Anyway, I would still be fine with changing REQUIRED to OPTIONAL, no objection...

Acknowledged.

> But could you maybe add the example back in? Or is there a reason for deleting it?

The `versionTime` example was moved to the top. I deleted the `service=agent` example because it's not clear what that example is attempting to express.

msporny:

PR #636, opened 2021-02-13
Fix editorial issues with Verification Relationship intro section.

msporny:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/636.html" title="Last updated on Feb 16, 2021, 3:15 PM UTC (a9b7c74)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/636/448648e...a9b7c74.html" title="Last updated on Feb 16, 2021, 3:15 PM UTC (a9b7c74)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1956,9 +1956,9 @@ invalid or revoked. Each <a>DID method</a> specification is expected to detail
 how revocation is performed and tracked.
       </p>
       <p>
-This specification defines several <a>verification relationships</a> below. A
-<a>DID document</a> MAY include any of these, or other properties, to express
-a specific <a>verification relationship</a>. In order to maximize global
+The following sections define several useful <a>verification relationships</a>.
+A <a>DID document</a> MAY include any of these, or other properties, to
+express a specific <a>verification relationship</a>. In order to maximize global
 interoperability, any such properties used SHOULD be registered in the DID
 Specification Registries [[?DID-SPEC-REGISTRIES]].
       </p>
--- a/index.html
+++ b/index.html
@@ -1950,7 +1950,7 @@ key agreement protocols with the <a>DID subject</a>&mdash;the value of the
       </p>
       <p>
 The <a>DID document</a> does not express revoked keys using a <a>verification
-relationship</a>. If a referenced verification method is not in the <a>DID
+relationship</a>. If a referenced verification method is not in the latest <a>DID
 Document</a> used to dereference it, then that verification method is considered
 invalid or revoked. Each <a>DID method</a> specification is expected to detail
 how revocation is performed and tracked.


index.html
```diff
@@ -2029,9 +2029,16 @@ <h2>Verification Relationships</h2>
...
-a specific <a>verification relationship</a>. In order to maximize global
+The <a>DID document</a> does not express revoked keys using a <a>verification
+relationship</a>. If a referenced verification method is not in the <a>DID
+Document</a> used to dereference it, then that verification method is considered
+invalid or revoked. Each <a>DID method</a> specification is expected to detail
+how revocation is performed and tracked.
+      </p>
+      <p>
+The following sections define several common <a>verification relationships</a>
+below. A <a>DID document</a> MAY include any of these, or other properties, to
```
TallTed, line 2040:
```suggestion
The sections below define several common <a>verification relationships</a>.
A <a>DID document</a> MAY include any of these, or other properties, to
```
or
```suggestion
The following sections define several common <a>verification relationships</a>.
A <a>DID document</a> MAY include any of these, or other properties, to
```

kdenhartog (dismissed):
While I recognize this isn't a normative statement, I'm not certain that this change is in alignment with the working group consensus based on the discussion in #14 and the number of thumbs down on this comment from you: https://github.com/w3c/did-core/issues/14#issuecomment-613514751

Can we make sure to get review from those who participated in that discussion to approve this before merging?

In my personal opinion, I believe this should have been addressed at the did-core spec level rather than the method level, but we've implicitly passed this onto method authors due to not being able to achieve consensus within the WG. Therefore I don't thing we should be explicitly stating this which to me implies the WG has agreed upon this

msporny:
> I'm not certain that this change is in alignment with the working group consensus

@kdenhartog, I'm confused by your statement. I think the text that you are referring to exists in the specification today. See the text right before Example 14:

https://w3c.github.io/did-core/#verification-method-types

I just moved it to a different part of the specification. 

That text was added to the specification in PR #490 to address issues #423, #240, and #386, was reviewed by seven people (including you), and merged after multiple edits and no objections:

https://github.com/w3c/did-core/pull/490

What am I missing? Exactly what text are you concerned about?

kdenhartog (commented):
Looks like this text was originally added when it was a work item in the CCG in this commit: https://github.com/w3c-ccg/did-spec/commit/2aaf2789fd4f4dfb4e8a042eca74479fefc4f4b0

Since then the text has been moved and adjusted slightly through a variety of editorial and normative changes (changes from MUST to MAY etc). I was always under the impression that handling of revoked keys was always a concern that we never achieved consensus on in the WG and so by default this text remained because we never came to a conclusion as a group of how we wanted it to be modified.

In any case, my review leads me to believe that this text has not been added by mistake and rather it's an artifact of the entire process that's led to where we are today. I'll go ahead and remove my "request changes" status even though I'm not certain this statement accurately reflects WG consensus.

index.html
```diff
@@ -2029,9 +2029,16 @@ <h2>Verification Relationships</h2>
 <code><a>keyAgreement</a></code> property needs to be used for that.
       </p>
       <p>
-This specification defines several <a>verification relationships</a> below. A
-<a>DID document</a> MAY include any of these, or other properties, to express
-a specific <a>verification relationship</a>. In order to maximize global
+The <a>DID document</a> does not express revoked keys using a <a>verification
```
OR13, line 2032: While I would prefer this to be true, I am fairly certain others don't agree. @oed @dhh1128 ... I suppose you would agree with `the latest <a>DID document</a>  does not express revoked keys`... which would introduce the concept of time / when revocation happened... I fear the complexity of such language and the security issues it produces / attempts to resolve.

I think the sentence cannot be true, without defining revocation...  a DID Document might contain revoked GPG keys OR keys that have had their private component tweeted, and where the attacker controls the did document... 

this is bad, it means that the attacker gains the value of all keys in a did document and can use this statement to argue that since they are present they are authoritative... "look I have your house keys, I must own your house"...

are we saying it's impossible for a did document to contain "revoked keys" isn't that like saying "DID Documents" do not contain emojis ?

this section also contradicts the informative section on "verification method revocation".... 

This is a topic it would have been better to spend call time on than `@context`.

oed: Yeah, this only makes sense if it refer to the *latest* DID document. If you resolve a DID document with the `versionId` param you should expect to get the state of the document at that point in time!

msporny: +1 to clarifying with "latest" language, that makes the statement more truthy.

> this is bad, it means that if we keep the propose change, when the attacker gains the value of all / some keys in a did document the attacker can use this statement to argue that since the attacker has not revoked the keys and they are present they are authoritative.

It's not good or bad, it's just reality. We can weaken the language, but the contract here is -- if the key is in the DID Document (and you have no other signals to say the key has been compromised), then you should assume the key hasn't been compromised.

That is true for any system that depends on PKI.

> are we saying it's impossible for a did document to contain "revoked keys" 

No, we are not saying that. I'll try to put in language to make sure people don't get that impression (although, I'd argue that's a pretty strange thing to take away from a read of the specification given the security/privacy considerations sections). We might want to write more about this in security considerations.

rhiaro (approved):
+1 to changes suggested by @TallTed and @oed

index.html
```diff
@@ -2029,9 +2029,16 @@ <h2>Verification Relationships</h2>
...
+relationship</a>. If a referenced verification method is not in the <a>DID
```
msporny, line 2033:
```suggestion
relationship</a>. If a referenced verification method is not in the latest <a>DID
```

index.html
OR13: > Even if a verification method is present in a DID Document, additional information, such as a public key revocation certificate, or an external allow or deny list, might be used to determine whether a verification method has been revoked.

https://w3c.github.io/did-core/#verification-method-revocation

perhaps this inconsistency is no longer present once `latest` is added :)

msporny:

PR #640, opened 2021-02-14
Fix editorial issues with Capability Invocation section.

msporny:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/640.html" title="Last updated on Feb 16, 2021, 3:27 PM UTC (5e93833)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/640/d355bda...5e93833.html" title="Last updated on Feb 16, 2021, 3:27 PM UTC (5e93833)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2223,7 +2223,7 @@ to wrap a decryption key for the recipient.
 The `capabilityInvocation` <a>verification relationship</a> is used to specify
 a <a>verification method</a> that might be used by the <a>DID subject</a> to
 invoke a cryptographic capability, such as the authorization to update the
-<a>DID Document</a> or the authorization to access an HTTP API.
+<a>DID Document</a>.
         </p>
 
         <dl>
@@ -2237,12 +2237,22 @@ embedded or referenced.
         </dl>
 
         <p>
-An example of when this property is useful is when a <a>DID subject</a>
-chooses to invoke their capability to start a vehicle through the combined
-usage of a <a>verification method</a> and the <code>StartCar</code>
-capability. In this example, the vehicle would be the <em>verifier</em> and
-would need to verify that the <a>verification method</a> exists in the
-<code><a>capabilityInvocation</a></code> property.
+An example of when this property is useful is when a <a>DID subject</a> needs to
+access a protected HTTP API that requires authorization in order to use it. In
+order to authorize when using the HTTP API, the <a>DID subject</a>
+<em>invokes</em> a capability that is associated with a particular URL that is
+exposed via the HTTP API. The invocation of the capability is typically a
+digitally signed message that is placed into the HTTP Headers.
+        </p>
+        <p>
+The server providing the HTTP API is the <em>verifier</em> of the capability and
+would need to verify that the <a>verification method</a> referred to by the
+invoked capability exists in the <code><a>capabilityInvocation</a></code>
+property of the <a>DID document</a>. The verifier would also check to make sure
+that the action being performed is valid and the capability matches the resource
+being accessed. If the verification is successful, the server has
+cryptographically determined that the invoker is authorized to access the
+protected resource.
         </p>
 
         <pre class="example nohighlight" title="Capability invocation property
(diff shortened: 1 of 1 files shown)


kdenhartog (approved):
I'm good with the switch in examples. It's going to help implementors infer how these are used better in my opinion. However, I did agree with @tplooker suggestion that switching "invoke" language to use (without changing the property) would help understand the usage of delegation/invocation better.

peacekeeper:
I like the changes, but maybe it would be sufficient and more readable to have the example use case (access to protected HTTP API via capabilities) in one place rather than two places (here and in https://github.com/w3c/did-core/pull/641). There seems to be some overlap and duplication.

Maybe it's better to consolidate and have a single example that describes both `capabilityInvocation` and `capabilityDelegation` simulataneously?

index.html
```diff
@@ -2237,12 +2237,22 @@ <h2>Capability Invocation</h2>
...
-usage of a <a>verification method</a> and the <code>StartCar</code>
-capability. In this example, the vehicle would be the <em>verifier</em> and
-would need to verify that the <a>verification method</a> exists in the
-<code><a>capabilityInvocation</a></code> property.
+An example of when this property is useful is when a <a>DID subject</a> needs to
+access a protected HTTP API that requires authorization in order to use it. In
+order to authorize when using the HTTP API, the <a>DID subject</a>
+<em>invokes</em> a capability that is associated with a particular URL that is
+exposed via the HTTP API. The invocation of the capability is typically a
+digitally signed message that is placed into the HTTP Headers.
```
dlongley, line 2245: -1 to "typically", +1 to describing an example.

```suggestion
exposed via the HTTP API. The invocation of the capability could be
expressed in a number of ways, e.g., as a digitally signed
message that is placed into the HTTP Headers.
```

index.html
```diff
@@ -2237,12 +2237,22 @@ <h2>Capability Invocation</h2>
...
+        </p>
+        <p>
+The server providing the HTTP API is the <em>verifier</em> of the capability and
+would need to verify that the <a>verification method</a> referred to by the
```
dlongley, line 2249:
```suggestion
it would need to verify that the <a>verification method</a> referred to by the
```

index.html
```diff
@@ -2237,12 +2237,22 @@ <h2>Capability Invocation</h2>
...
+invoked capability exists in the <code><a>capabilityInvocation</a></code>
+property of the <a>DID document</a>. The verifier would also check to make sure
+that the action being performed is valid and the capability matches the resource
+being accessed. If the verification is successful, the server has
```
dlongley, line 2253:
```suggestion
that the action being performed is valid and the capability is appropriate for the
resource being accessed. If the verification is successful, the server has
```

dlongley (approved):
Approved w/suggestions.

rhiaro (approved):

index.html
```diff
@@ -2237,12 +2237,22 @@ <h2>Capability Invocation</h2>
...
-An example of when this property is useful is when a <a>DID subject</a>
-chooses to invoke their capability to start a vehicle through the combined
-usage of a <a>verification method</a> and the <code>StartCar</code>
-capability. In this example, the vehicle would be the <em>verifier</em> and
-would need to verify that the <a>verification method</a> exists in the
-<code><a>capabilityInvocation</a></code> property.
+An example of when this property is useful is when a <a>DID subject</a> needs to
+access a protected HTTP API that requires authorization in order to use it. In
+order to authorize when using the HTTP API, the <a>DID subject</a>
+<em>invokes</em> a capability that is associated with a particular URL that is
```
msporny, line 2243:
```suggestion
uses a capability that is associated with a particular URL that is
```

msporny: Apply @tplooker's suggestion to avoid the use of "invoke" and say "use" instead.

msporny:
> I like the changes, but maybe it would be sufficient and more readable to have the example use case (access to protected HTTP API via capabilities) in one place rather than two places (here and in #641). There seems to be some overlap and duplication.
> Maybe it's better to consolidate and have a single example that describes both `capabilityInvocation` and `capabilityDelegation` simulataneously?

Yes, I tried combining the Capability section and it became a bit large, so decided to keep it separate. We can try to do some more editorial cleanup after this PR is in.

msporny:

PR #644, opened 2021-02-15
Fix substantive issues related to Representations intro section.

msporny:
This PR cleans up the introductory portion of the Representations section by 1) providing a non-normative background to what representation/production/consumption is, 2) groups the normative statements into 3 groups -- all representations, all conforming producers, and all conforming consumers, 3) downgrades a MAY statement that wasn't easily testable but keeps the language around to make the intent clear. I tried to keep the essence of the section intact, but did end up modifying a few normative statements... so this PR makes substantive changes while trying to preserve what the section was doing.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/644.html" title="Last updated on Feb 17, 2021, 8:27 PM UTC (81e4417)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/644/6269382...81e4417.html" title="Last updated on Feb 17, 2021, 8:27 PM UTC (81e4417)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2369,26 +2369,27 @@ href="#authentication-service-endpoints"></a>.
 A concrete serialization of any data structure in this specification is called a
 <a>representation</a>. A <a>representation</a> is created by serializing the <a
 href="#data-model">data model</a> through a process called
-<dfn>production</dfn>. A <a>representation</a> is transformed back into the <a
+<dfn>production</dfn>. A <a>representation</a> is transformed into the <a
 href="#data-model">data model</a> through a process called
-<dfn>consumption</dfn>. The <em>production</em> and <em>consumption</em> process
-enables the conversion of information from one <a>representation</a> to another.
-Although this specification defines <a>representations</a> for JSON, JSON-LD,
-and CBOR, developers can use any other <a>representation</a>, such as XML or
+<dfn>consumption</dfn>. The <em>production</em> and <em>consumption</em> processes
+enable the conversion of information from one <a>representation</a> to another.
+This specification defines <a>representations</a> for JSON, JSON-LD,
+and CBOR, and developers can use any other <a>representation</a>, such as XML or
 YAML, that is capable of expressing the <a href="#data-model">data model</a>.
 The following sections define the general rules for <a>production</a> and
-<a>consumption</a> and the JSON, JSON-LD, and CBOR <a>representations</a>.
+<a>consumption</a>, as well as the JSON, JSON-LD, and CBOR <a>representations</a>.
     </p>
 
     <section>
       <h2>Production and Consumption</h2>
 
       <p>
-The rules in this section apply to all implementations seeking to be compatible
-with independent implementations of the specification. Deployments of this
-specification can use a custom agreed-upon <a>representation</a>, including
-rules agreed to by a pair of producers and consumers for handling properties not
-listed in the registry. See <a href="#extensibility"></a> for more information.
+In addition to the <a>representations</a> defined in this specification,
+implementers can use other <a>representations</a>, providing each such
+<a>representation</a> is properly specified (including rules for
+interoperable handling of properties not listed in the DID Specification
+Registries [[?DID-SPEC-REGISTRIES]]). See <a href="#extensibility"></a>
+for more information.
       </p>
       <p>
 The requirements for all <a>representations</a> are as follows:
@@ -2408,7 +2409,7 @@ type that are conformant with the fragment processing rules defined in
 <a href="#fragment"></a>.
         </li>
         <li>
-A <a>representations</a> SHOULD use the lexical representation of <a
+A <a>representation</a> SHOULD use the lexical representation of <a
 href="#data-model">data model</a> data types. For example, JSON and JSON-LD use
 the XML Schema <code>dateTime</code> lexical serialization to represent
 <a>datetimes</a>. A <a>representation</a> MAY choose to serialize the <a
@@ -2422,7 +2423,7 @@ represent the number of seconds since the Unix epoch.
 A <a>representation</a> MAY define representation-specific entries that can be
 stored in a data structure expressible using the <a href="#data-model">data
 model</a> for use during the <a>production</a> and <a>consumption</a> process.
-These entries are utilized when consuming or producing to aid in ensuring
+These entries are used when consuming or producing to aid in ensuring
 lossless conversion.
         </li>
         <li>
--- a/index.html
+++ b/index.html
@@ -2366,7 +2366,7 @@ href="#authentication-service-endpoints"></a>.
     <h1>Representations</h1>
 
     <p>
-A concrete serialization of any data structure in this specification is called a
+A concrete serialization of a <a>DID document</a> in this specification is called a
 <a>representation</a>. A <a>representation</a> is created by serializing the <a
 href="#data-model">data model</a> through a process called
 <dfn>production</dfn>. A <a>representation</a> is transformed into the <a
--- a/index.html
+++ b/index.html
@@ -2366,120 +2366,129 @@ href="#authentication-service-endpoints"></a>.
     <h1>Representations</h1>
 
     <p>
-All concrete <a>representations</a> of a <a>DID document</a> are serialized using a
-deterministic mapping that is able to be unambiguously parsed into the <a
-href="#data-model">data model</a> defined in this specification. All
-serialization methods MUST define rules for the bidirectional translation of a
-<a>DID document</a> both into and out of the <a>representation</a> in question. An
-implementation MUST NOT convert between <a>representations</a> without first parsing to
-a <a href="#data-model">data model</a> (described in Sections <a
-href="#data-model"></a> and <a href="#core-properties"></a>); translation
-between any two <a>representations</a> is done by parsing the source <a>representation</a>
-into the <a href="#data-model">data model</a> and then serializing the <a
-href="#data-model">data model</a> into the target <a>representation</a>.
-    </p>
-
-    <p>
-Although syntactic mappings are provided for JSON, JSON-LD, and CBOR here,
-applications and services MAY use any other data <a>representation</a> syntax that is
-capable of expressing the <a href="#data-model">data model</a>, such as XML or
-YAML.
-    </p>
-
-    <p>
-Producers MUST indicate which <a>representation</a> of a document has been used via a
-media type in the document's metadata. Consumers MUST determine the
-<a>representation</a> of a <a>DID document</a> via the <code>contentType</code> <a>DID
-resolver</a> metadata field (see <a href="#did-resolution"></a>), <em>not</em>
-through the content of the <a>DID document</a> alone.
-    </p>
-
-    <p>
-A <a>conforming producer</a>
-MUST NOT produce a <a>DID document</a> containing different entries with
-the same <code>id</code> value. A <a>conforming consumer</a> MUST produce
-an error if it detects different entries with the same <code>id</code> value.
-    </p>
-
-    <p>
-Unrecognized entries in the <a href="#data-model">data model</a> MUST be
-preserved. An unrecognized entry is any entry that does not have explicit
-processing rules known to the consumer or producer. Consumers MUST add all
-entries that do not have explicit processing rules for the
-<a>representation</a> being consumed to the <a href="#data-model">data
-model</a> using only the <a>representation</a>'s generic type processing rules.
-Producers MUST serialize all entries in the <a href="#data-model">data
-model</a> that do not have explicit processing rules for the
-<a>representation</a> being produced using only the <a>representation</a>'s
-generic type processing rules.
-    </p>
-
-    <p class="note" title="Representation-specific entries">
-Note that entries that contain representation-specific syntax will only have
-special processing rules defined by a single <a>representation</a>. Consumers of a
-different <a>representation</a> are required to include these entries in the <a
-href="#data-model">data model</a> using only their generic type processing rules
-to enable lossless conversion of <a>representations</a>. Similarly, producers are
-required to treat entries containing representation-specific syntax using
-generic type processing rules when producing a <a>representation</a> for which
-the entry is not defined. <a>Representations</a> are required to define
-producer behavior for any such entries defined by the <a>representation</a>.
-    </p>
-
-    <p class="note" title="Representation lexical and value space">
-It is RECOMMENDED that <a>representations</a> use the lexical representation
-of registered data types. For example, JSON and JSON-LD use the XML
-Schema dateTime lexical representation to represent <a>datetimes</a>.
-A <a>representation</a> MAY choose to represent the data types differently.
-For example, some CBOR-based <a>representations</a> express <a>datetime</a>
-values using integers to represent the number of seconds since the Unix epoch.
-    </p>
-
-    <p>
-The production and consumption rules in this section apply to all
-implementations seeking to be fully compatible with independent implementations
-of the specification. Deployments of this specification MAY use a custom
-agreed-upon <a>representation</a>, including rules agreed to by a pair of producers
-and consumers for handling properties not listed in the registry. See Section
-<a href="#extensibility"></a> for more information.
+A concrete serialization of any data structure in this specification is called a
+<a>representation</a>. A <a>representation</a> is created by serializing the <a
+href="#data-model">data model</a> through a process called
+<dfn>production</dfn>. A <a>representation</a> is transformed back into the <a
+href="#data-model">data model</a> through a process called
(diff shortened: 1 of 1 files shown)


OR13 (approved):

index.html
```diff
@@ -2455,113 +2455,129 @@ <h2>Services</h2>
...
-href="#data-model">data model</a> defined in this specification. All
-serialization methods MUST define rules for the bidirectional translation of a
-<a>DID document</a> both into and out of the <a>representation</a> in question. An
-implementation MUST NOT convert between <a>representations</a> without first parsing to
-a <a href="#data-model">data model</a> (described in Sections <a
-href="#data-model"></a> and <a href="#core-properties"></a>); translation
-between any two <a>representations</a> is done by parsing the source <a>representation</a>
-into the <a href="#data-model">data model</a> and then serializing the <a
-href="#data-model">data model</a> into the target <a>representation</a>.
+A concrete serialization of any data structure in this specification is called a
```
peacekeeper, line 2458:
```suggestion
A concrete serialization of a <a>DID document</a> in this specification is called a
```

Let's not change this please, especially since the text right after this sentence refers to the DID document data model. I imagine you're probably trying to cover the metadata structures as well, but I think it's better to keep those separate from the DID document data model and representations for now...

index.html
```diff
@@ -2455,113 +2455,129 @@ <h2>Services</h2>
...
-<a>representation</a> being consumed to the <a href="#data-model">data
-model</a> using only the <a>representation</a>'s generic type processing rules.
-Producers MUST serialize all entries in the <a href="#data-model">data
-model</a> that do not have explicit processing rules for the
-<a>representation</a> being produced using only the <a>representation</a>'s
-generic type processing rules.
-    </p>
+      <p>
+The rules in this section apply to all implementations seeking to be compatible
+with independent implementations of the specification. Deployments of this
```
peacekeeper, line 2477: This language sounds funny to me, not sure exactly what it means :)

peacekeeper (commented):
I'm mostly fine with this (minus one small change in the beginning).

I still think it's super confusing to talk about "representation-specific entries" in a data model that's supposed to be abstract. But I like how this PR makes the section easier to read and understand.

Not sure what you meant by _"downgrades a MAY statement that wasn't easily testable"_, I couldn't find that.

index.html
```diff
@@ -2455,113 +2455,129 @@ <h2>Services</h2>
...
+specification can use a custom agreed-upon <a>representation</a>, including
+rules agreed to by a pair of producers and consumers for handling properties not
+listed in the registry. See <a href="#extensibility"></a> for more information.
```
rhiaro, line 2480:
```suggestion
As well as the <a>representations</a> defined in this specification,
implementers can use other <a>representations</a>, providing each
<a>representation</a> is properly specified (including
rules for handling properties not listed in the DID Specification
Registries [[?DID-SPEC-REGISTRIES]]). See <a
href="#extensibility"></a> for more information.
```

I'm having trouble parsing this. I've tried to rephrase, but not sure if I've captured what you meant.

index.html
```diff
@@ -2455,113 +2455,129 @@ <h2>Services</h2>
...
+A <a>representation</a> MUST be uniquely associated with an IANA-registered
+MIME type.
+        </li>
+        <li>
+A <a>representation</a> MUST define fragment processing rules for its MIME
+type that are conformant with the fragment processing rules defined in
+<a href="#fragment"></a>.
+        </li>
+        <li>
+A <a>representations</a> SHOULD use the lexical representation of <a
```
rhiaro, line 2500:
```suggestion
A <a>representation</a> SHOULD use the lexical representation of <a
```

index.html
```diff
@@ -2455,113 +2455,129 @@ <h2>Services</h2>
...
+serializations as long as the <a>consumption</a> process back into the <a
+href="#data-model">data model</a> is lossless. For example, some CBOR-based
+<a>representations</a> express <a>datetime</a> values using integers to
+represent the number of seconds since the Unix epoch.
+        </li>
+        <li>
+A <a>representation</a> MAY define representation-specific entries that can be
+stored in a data structure expressible using the <a href="#data-model">data
+model</a> for use during the <a>production</a> and <a>consumption</a> process.
+These entries are utilized when consuming or producing to aid in ensuring
```
rhiaro, line 2514:
```suggestion
These entries are used when consuming or producing to aid in ensuring
```

rhiaro (changes requested):

index.html
TallTed: Maybe...?
```suggestion
In addition to the <a>representations</a> defined in this specification,
implementers can use other <a>representations</a>, providing each such
<a>representation</a> is properly specified (including rules for
interoperable handling of properties not listed in the DID Specification
Registries [[?DID-SPEC-REGISTRIES]]). See <a href="#extensibility"></a>
for more information.
```

index.html
msporny: We need to change this in order to make the DID Resolution section testable. There has to be a clear path to serialization for the data structures in the Resolution section, namely the metadata structure. We can achieve that by making sure the ADM isn't hard coded to just apply to DID Documents. I have made the changes in each section and it all seems to stick together fairly well. We can discuss on the call today. If we don't make this change, we're back to the Resolution section not having a clear path to a testable serialization.

index.html
msporny: This was language that @jricher created, I believe... I think what he's saying is "If you are writing an implementation, and you want to be compatible with other implementations, these rules apply to you."

I do agree that the sentence is hard to read and that we don't really need it. Let's see if @jricher has alternative language, and if not, I can come up with something (or just remove the language, since it's fairly obvious that the rules are there in order to achieve interoperability).

index.html
msporny: Applied @TallTed's suggestion.

msporny:
I'm holding this PR until we can resolve @peacekeeper's ADM serialization concern.

index.html
```diff
@@ -2366,120 +2366,130 @@ <h2>Services</h2>
...
+        <li>
+A <a>representation</a> SHOULD use the lexical representation of <a
+href="#data-model">data model</a> data types. For example, JSON and JSON-LD use
+the XML Schema <code>dateTime</code> lexical serialization to represent
+<a>datetimes</a>. A <a>representation</a> MAY choose to serialize the <a
+href="#data-model">data model</a> data types using a different lexical
+serializations as long as the <a>consumption</a> process back into the <a
+href="#data-model">data model</a> is lossless. For example, some CBOR-based
+<a>representations</a> express <a>datetime</a> values using integers to
+represent the number of seconds since the Unix epoch.
```
jonnycrunch, line 2420: `some CBOR-based representations express data time values using integers` 

I don't believe this is correct.  We constrained the CBOR to major type 3, which is string.   I think this type of extensibility could be made in the DID-Spec Registries, not in Core.

index.html
OR13: I don't think this proposed change does what you think it does @peacekeeper ... we define a DID Document as an INFRA map.... I would be happy to take the change.... Since a DID Document is just an INFRA map, with a required property of `id`.

iherman:
This issue was discussed in [a meeting](https://www.w3.org/2019/did-wg/Meetings/Minutes/2021-02-16-did#section6-1).

- No actions or resolutions
6.1. resolution testing<br/>
See github pull request #644.<br/>
Manu Sporny: one issue with the resolution section is that it is abstract so we are still trying to figure out how to make it concrete enough to test<br/>
… one approach is to say that the metadata structure that represents the abstract data model can be serialized so that it can then be translated into various representations<br/>
… in order to do that we are suggesting a change to the abstract data model<br/>
… up to this point it has only been about DID documents<br/>
… but we are now considering making the abstract data model can represent other data structures that may be needed to create a specific representation<br/>
… and then you can use the representation production rules in order to create a concrete representation<br/>
… Markus and Justin have expressed some reservations about this<br/>
… so we are open to discussing it<br/>
Markus Sabadello: I understand the motivation to make things more testable…<br/>
… but the intention was to make the DID resolution section abstract because concrete resolution protocols were out of scope<br/>
… so it can be confusing to say that the abstract data model can be used for other things besides DID documents<br/>
… so Markus would prefer that the abstract data model stays focused on DID documents and not include metadata structures<br/>
… since metadata structures are “really something else”<br/>
Brent Zundel: We have repeatedly used “testable” to be machine testable. However “testable” does not have to mean machine-testable.<br/>
Justin Richer: What concerns me about this proposal is that it is “leaking requirements” of the test suite into the specification<br/>
… not all of the requirements of the test suite need to be represented in the spec<br/>
… for example, in the resolution spec we agreed to define “contracts”<br/>
… that does not mean that the core specification needs to go into this level of detailing<br/>
… and it does not mean the spec needs to change its nature to fit the needs of an implementation<br/>
… this is dangerous and backwards<br/>
Markus Sabadello: In the section on metadata structures, we defined them the same way as the abstract data model where we use Infra and data maps<br/>
… so we do have examples for how this abstract data structure that could be represented in JSON<br/>
… so we could add some examples that show how the abstract data model can be made machine testable<br/>
Brent Zundel: Roughly the two approaches are either we need to describe how to make concrete the abstract parts of the spec in order to make them machine-testable…<br/>
… or we alter the test suite in order to machine-test those statements<br/>
Manu Sporny: The issue we are in is that the group defined resolution in the abstract…<br/>
… but when the language was written for the resolution section, we used a whole bunch of normative MUST statements<br/>
… then we decided that we should have machine-testing of normative MUST statements<br/>
… so now we have a resolution section that needs testing<br/>
… so why don’t we focus on the actual language of the PRs<br/>
… there are two proposals: make the normative statements testable OR<br/>
… put the logic for testing into the test suite and say that is good enough<br/>
… we need more input from the WG about which direction we should go<br/>
… and see what specific language we can agree on<br/>
… for example, Markus suggests that we can provide examples of how the abstract data model can be serialized<br/>
Brent Zundel: Let’s move in the direction of looking at the PR text<br/>
Orie Steele: where is the list of normative requirements.<br/>
Jonathan Holt: Adding this detail this late in the game is not a good idea - it can be handled in the DID Spec Registries<br/>
… I added an example for CBOR that uses date time stamps using vanilla CBOR<br/>
… so I added that detail to DID Spec Registries instead<br/>
Markus Sabadello: We should really look at the concrete impacts. I think this is a big change.<br/>
… I understand the motivations, but I consider the metadata to be very different than the DID document data<br/>
… so keeping the abstract data model focused on the DID document is more precise<br/>
… it is not helpful to use the abstract data model for metadata<br/>
… we defined the abstract data model for DID documents, not for other metadata<br/>
Orie Steele: why is the type of did document an infra map… if its not an infra map?<br/>
Markus Sabadello: the metadata has some similarities, but it will be handled differently than the DID document data<br/>
Justin Richer: +1 to markus_sabadello<br/>
Markus Sabadello: my proposal is to expand the examples of what the metadata would look like in JSON<br/>
… and then make it testable in the test suite<br/>
Manu Sporny: I would like people to look at the text that is under debate. WG members should read the PR.<br/>
… I disagree with what the abstract data model is supposed to do.<br/>
… it should apply to everything in the specification<br/>
… so from my perspective, the ADM was supposed to be for any data structure specified in DID Core<br/>
… for those who are saying that the ADM should not apply, I’d like to know why we are “creating this other world” that needs to be respecified in the same way we define the ADM<br/>
… but we may just want to move on to other PRs at this point as we need the rest of the WG to engage on this issue<br/>
… but not deciding about this issue is going to slow down finishing this section<br/>
Justin Richer: to me, the ADM was always meant to be only for DID Documents. that we use the same definitions for other data structures don’t mean they’re the same<br/>
Brent Zundel: Note that the special topic call today will be a continuation of this discussion<br/>
Ivan Herman: Taking it to a higher level: originally the goal of resolution was defining a contract<br/>
… regardless of whether INFRA can be used to define these contracts, we must ultimately make these contracts machine-testable<br/>
… so it has come up several times in the past few weeks which relies on human reports of implementability<br/>
Justin Richer: +1 to ivan<br/>
Ivan Herman: and whether going to one extreme is the basis of the disagreement<br/>
Manu Sporny: I worry that we are speaking out of both sides of our mouth<br/>
… we should either agree that we’re not doing machine testability of the resolution section<br/>
Justin Richer: -1<br/>
Amy Guy: +1 DID resolution in a Note<br/>
Orie Steele: -1 to removing resolution… its 99% of what the spec is about…<br/>
Justin Richer: ORie: that’s why it got added in a year ago, it doesn’t make sense to talk about DIDs and Documents w/o talking about what’s in between<br/>
Drummond Reed: It feels like if we lock in on what the requirements in the spec are for the contract<br/>
… and if we say the contract is abstract, then maybe the way to test it can be described in the test suite<br/>
Justin Richer: +1 to drummond<br/>
Ivan Herman: +1 to what Drummond said - in addition, the presentation of things whereby having a normative statement must be backed up my machine-testing is not a requirement<br/>
… it is okay to have a normative statement that is not directly machine-testable<br/>
… we should be more liberal in that respect<br/>
Joe Andrieu: I think the interpreted rule that all normative statements much be machine-testable is too restrictive<br/>
… the definition of a contract is that it will say what you must do, and whether or not it is machine-testable is a different issue<br/>
Daniel Buchner: I strongly agree with Joe<br/>
Daniel Buchner: Folks, JOE AND I STRONGLY AGREE<br/>
Daniel Buchner: this means it is the right thing, obviously<br/>
Markus Sabadello: Based on what others are saying here, I want to ask: is it acceptable to have normative statements in the spec that are not directly machine testable without additional information…<br/>
… but that additional information is readily available in some additional documents<br/>
Brent Zundel: +1 to Markus<br/>
Markus Sabadello: for example in other protocol specification documents in the W3C Credentials Community Group or at the Decentralized Identity Foundation<br/>
Manu Sporny: I’m confused now as to whether we are testing DID resolution or not<br/>
… some are staying that we shouldn’t test it at all, vs. testing with some additional info that’s not in the spec<br/>
… I have been operating under the assumption that because there are so many normative statements in the resolution section, that we should make it all machine testable<br/>
Ivan Herman: First, to answer to Markus, I’m not sure I fully understand the question<br/>
… the implementation of the resolution contract should not depend on another spec or document<br/>
… the #1 question is whether the contract is fulfilled or not<br/>
… to reply to Manu, my view is that we have a contract definition. We can define “tests” that, for specific data, this is how the contract can be fulfilled and this is the response it should return<br/>
… but whether or not that is machine-testable is not required<br/>
… if it is machine-testable, that is good, but not required<br/>
Daniel Buchner: There are 3 classes of testability. 1) 100% testable via a local DID document. Fully deterministic.<br/>
… 2) the class of untestable language, such as “you must not use data describing a human”<br/>
Justin Richer: +1 to the coupling idea, that’s what I’ve been saying all along<br/>
Markus Sabadello: Can’t we use this in the test suite? <a href="https://w3c-ccg.github.io/did-resolution/#did-resolution-result">https://w3c-ccg.github.io/did-resolution/#did-resolution-result</a><br/>
Orie Steele: I spent some time trying to implement the abstract data model this weekend and I’m starting to agree with some of what Justin and Markus are saying<br/>
… there is a class of tests that are not valuable to use as spec authors<br/>
… there probably needs to be a way to test what is in the spec without talking about a specific implementation<br/>
… this is challenging because it requires some concrete implementation of the ADM in the test suite<br/>
… anyone who says that we should not test testable statements as long as there is some way to test them is harming the spec<br/>
… the ADM is testable, but what’s hard is that the language around the ADM how to interpret it<br/>
… for example, if production works directly from the ADM, or does it take additional arguments<br/>
… so we have a problem here. There are people talking past each other.<br/>
… the solution is to gain clarity around the function signatures for the processes defined in the spec<br/>
Markus Sabadello: +1 to Orie that production is too vague right now<br/>
Orie Steele: then writing tests will be much easier<br/>
Jonathan Holt: Orie: +1, that was extremely well stated!<br/>
Daniel Buchner: We should test certain things with a default set of functions that implementers can put their own function bodies into them<br/>
Manu Sporny: Yes, +1 to Orie, that’s the problem.<br/>
Daniel Buchner: for ex: you can test against in-mem default values, but allow devs to replace the function with their own DB fetch function to get the same values<br/>
Daniel Buchner: that’s still a deterministic test<br/>
</details>

index.html
msporny: The CBOR-LD encoding of datetime uses an integer.

index.html
msporny: Applied @peacekeeper's change per DID WG call resolution: https://www.w3.org/2021/02/16-did-topic-minutes.html#r01

msporny:
Substantive, multiple reviews, changes requested and made, objection raised by @peacekeeper and address via spec text change, no objections remain, merging.

PR #645, opened 2021-02-15
references DID Spec registries for normative values to version properly

kdenhartog:
This modification helps us to avoid having to have a normative errata change post recommendation status when we want to version the did context value. It coincides with PR https://github.com/w3c/did-spec-registries/pull/251 in the spec registries as well.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/645.html" title="Last updated on Feb 15, 2021, 1:14 AM UTC (765b49c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/645/d355bda...765b49c.html" title="Last updated on Feb 15, 2021, 1:14 AM UTC (765b49c)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2910,13 +2910,15 @@ valid according to the consumption rules.
             </p>
 
             <p>
-The value of <code>@context</code> MUST be exactly one of these values.
+The value of <code>@context</code> MUST be registered in the DID
+Specification Registries [[?DID-SPEC-REGISTRIES]].
             </p>
 
             <ul>
               <li>
 The <a href="https://tools.ietf.org/html/rfc8259#section-7">JSON String</a>
-<code>https://www.w3.org/ns/did/v1</code>.
+being registered in the DID
+Specification Registries [[?DID-SPEC-REGISTRIES]].
                 <pre class="example">
 {
   "@context": "https://www.w3.org/ns/did/v1",
@@ -2927,7 +2929,8 @@ The <a href="https://tools.ietf.org/html/rfc8259#section-7">JSON String</a>
               <li>
 A <a href="https://tools.ietf.org/html/rfc8259#section-5">JSON Array</a>,
 with first item the <a href="https://tools.ietf.org/html/rfc8259#section-7">JSON
-String</a> <code>https://www.w3.org/ns/did/v1</code>, and subsequent items of
+  String</a> being a registered value in the DID Specification Registries
+[[?DID-SPEC-REGISTRIES]] and subsequent items of
 type <a href="https://tools.ietf.org/html/rfc8259#section-7">JSON String</a> or
 <a href="https://tools.ietf.org/html/rfc8259#section-4">JSON Object</a>.
                 <pre class="example">


OR13 (approved):

index.html
```diff
@@ -2910,13 +2910,15 @@ <h3>Production</h3>
...
             <p>
-The value of <code>@context</code> MUST be exactly one of these values.
+The value of <code>@context</code> MUST be registered in the DID
+Specification Registries [[?DID-SPEC-REGISTRIES]].
             </p>
 
             <ul>
               <li>
 The <a href="https://tools.ietf.org/html/rfc8259#section-7">JSON String</a>
-<code>https://www.w3.org/ns/did/v1</code>.
```
msporny, line 2919: Hmm. I get what you're going for here -- you want us to be able to move to a v2 base DID Context w/o having to rev the spec -- I'm not sure we actually want that. We are publishing v1.0 of the DID Core specification -- the only version of the context that's appropriate is v1; the only `@context` you should use is the one listed. When we advance to v2.0, we will update this URL in the specification via the DID Maintenance WG (or a newly rechartered WG). The v1.0 context is tightly bound to the specification, since it's the specification that defines the did/v1 context. 

The DID Spec Registries also don't require WG consensus to update/revise. That could be problematic because a tiny subset of us could decide to do a v2 rev w/o WG input, leading to destabilizing the core specification and ecosystem. I'd rather we not open ourselves up to that possibility.

Until we can address those concerns, I'm a -1 on this change.

msporny (changes requested):

index.html
```diff
@@ -2910,13 +2910,15 @@ <h3>Production</h3>
             </p>
 
             <p>
-The value of <code>@context</code> MUST be exactly one of these values.
+The value of <code>@context</code> MUST be registered in the DID
+Specification Registries [[?DID-SPEC-REGISTRIES]].
```
rhiaro, line 2914: This is a normative (and critical) dependency on the Registries, which is not possible as the Registries is not a normative specification.

index.html
kdenhartog: If this is the case, then we should only be including terms defined within the did-core spec inside the did context to prevent context errors like we encountered with the VC context. This would mean removing the LDP suites from the did-core context. 

We should also then consider how we're going to handle extensions in the best way possible.

1. Define a superset context of did-core+extensions and modify the language in did-core to state "MUST use `https://w3.org/ns/did/v1` or a compatible context which MUST NOT redefine the `@protected` terms defined in `https://w3.org/ns/did/v1`"
2. Define an extensions context which doesn't include any of the did-core properties and any time an extension is used that context would need to be included. No did-core spec text changes needed.
3. Utilize the contexts provided in the registry directly. No modifications made to did-core spec text.

That's my preferences ordered. Are there any options that I'm missing that should be considered?

I'm alright with not merging these changes (especially since @rhiaro points out this places a normative dependency on a non-normative document), but I want to make sure that we finish thinking through our strategy of supporting extensions (and get it changed in the docs) before we cause the headache that's been encountered with the VC spec.

kdenhartog:
Closing this PR in favor of discussion continuing on https://github.com/w3c/did-spec-registries/pull/251

If we decide to go with option 1 which requires spec text changes I'll open a new PR.

PR #593, opened 2021-02-02
Remove DagCBOR from the specification.

msporny:
This PR implements the WG decision to decouple DagCBOR from DID Core and move it into it's own specification. The resolution can be found here: https://www.w3.org/2021/02/02-did-minutes.html#r01

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/593.html" title="Last updated on Feb 2, 2021, 10:55 PM UTC (b84f561)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/593/bad3368...b84f561.html" title="Last updated on Feb 2, 2021, 10:55 PM UTC (b84f561)">Diff</a>

PR #650, opened 2021-02-15
Remove the IANA section for application/did+dag+cbor.

msporny:
This should have been removed by PR #593 https://github.com/w3c/did-core/pull/593/files#diff-0eb547304658805aad788d320f10bf1f292797b5e6d745a3bf617584da017051L6254 ... but the section has somehow reappeared in the spec. Removing again. Harder, this time.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/650.html" title="Last updated on Feb 15, 2021, 5:20 PM UTC (631174a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/650/d355bda...631174a.html" title="Last updated on Feb 15, 2021, 5:20 PM UTC (631174a)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -5992,66 +5992,6 @@ according to the rules defined in <a href="#fragment"></a>.
       </p>
     </section>
 
-    <section>
-      <h2>application/did+dag+cbor</h2>
-      <dl>
-        <dt>Type name:</dt>
-        <dd>application</dd>
-        <dt>Subtype name:</dt>
-        <dd>did+dag+cbor</dd>
-        <dt>Required parameters:</dt>
-        <dd>None</dd>
-        <dt>Optional parameters:</dt>
-        <dd>None</dd>
-        <dt>Encoding considerations:</dt>
-        <dd>
-See <a data-cite="RFC8949#section-4.2.1">RFC&nbsp;8949, section 4.2.1</a>.
-        </dd>
-        <dt>Security considerations:</dt>
-        <dd>
-See <a data-cite="RFC8949#section-10">RFC&nbsp;8949, section 10</a> [[RFC8949]].
-        </dd>
-        <dt>Interoperability considerations:</dt>
-        <dd>Not Applicable</dd>
-        <dt>Published specification:</dt>
-        <dd>http://www.w3.org/TR/did-core/</dd>
-        <dt>Applications that use this media type:</dt>
-        <dd>
-Any application that requires an identifier that is decentralized, persistent,
-cryptographically verifiable, and resolvable. Applications typically consist of
-cryptographic identity systems, decentralized networks of devices, and
-websites that issue or verify W3C Verifiable Credentials.
-        </dd>
-        <dt>Additional information:</dt>
-        <dd>
-          <dl>
-            <dt>Magic number(s):</dt>
-            <dd>Not Applicable</dd>
-            <dt>File extension(s):</dt>
-            <dd>.did</dd>
-            <dt>Macintosh file type code(s):</dt>
-            <dd>TEXT</dd>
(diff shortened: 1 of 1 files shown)


OR13:
hmm, might we consider applying for reserving the following:

- `application/did+json`
- `application/did+ld+son`
- `application/did+cbor`
- `application/did+ld+cbor`
- `application/did+dag+cbor`
- `application/did+yaml`
- `application/did+xml`

It feels like maybe all of these belong in did-spec registries... and that we use that repo to coordinate with IANA... otherwise we will be creating a problem for future representations that are created after the spec is published....

In short, I would accept this PR IF we had a PR that added `application/did+dag+cbor` to did spec registries, and it had been merged.

OR13 (changes requested):
IMO, we need to handle this the same way for all representations or we don't have an abstract data model.

msporny:
> hmm, might we consider applying for reserving the following:
> 
> * `application/did+ld+cbor`
> * `application/did+dag+cbor`
> * `application/did+yaml`
> * `application/did+xml`

To reserve something at IANA in the "standards tree", it needs to exist in specification form. The representations above do not exist in specification form, therefore if we tried to register them, the registration would be denied:

https://tools.ietf.org/html/rfc6838#section-3

> It feels like maybe all of these belong in did-spec registries... and that we use that repo to coordinate with IANA... otherwise we will be creating a problem for future representations that are created after the spec is published....

The only responsibility the DID WG has in coordinating with IANA is handling the Media type registrations that it is creating. The DID Spec Registries, nor the DID WG have any responsibility for specifications that are not under our jurisdiction.

> In short, I would accept this PR IF we had a PR that added `application/did+dag+cbor` to did spec registries, and it had been merged.

We can't do that w/o a specification... registration of that specification would be up to the author of that specification.

The DID WG had consensus to remove the DagCBOR section to a separate specification. It is up to the people that want to see that specification happen to do the work to make it happen.

If you are suggesting that the Representations section be moved into separate specifications, we don't have consensus on doing that and you'd need to raise the issue with the WG and get consensus on that topic.

It sounds like your concern has to do with there not being a clear section in DID Specification Registries to register representations? I have added language in a recent PR to address this: https://github.com/w3c/did-core/pull/644/files#diff-0eb547304658805aad788d320f10bf1f292797b5e6d745a3bf617584da017051R2518-R2520

The only remaining thing that needs to be done is for someone to add that section in DID Specification Registries.

rhiaro (approved):

OR13 (approved):
@msporny agree with your feedback.

iherman:
I agree with @msporny that the DID Core spec must include only those IANA sections that apply to types that are specified for usage in the present document. The Core spec should not be used to register anything else.

Additionally, a small remark:

> The only remaining thing that needs to be done is for someone to add that section in DID Specification Registries. I suggest you raise a PR to do that [...]

Just a warning. W3C has a specific procedure so that a W3C Recommendation (more exactly a specific section thereof) is accepted as the official media type documentation by IANA. If a media type is done in a W3C Note, this procedure may not be available for us anymore, and the media type documentation may become a bit more complicated. (I would have to check.)

iherman (approved):

OR13:
@iherman the media type specification will be registered in the note. but defined elsewhere... eventually... the core point being that the did wg will end, but did spec registries will live on.

iherman:
> did wg will end, but did spec registries will live on...

... happily ever after :-)

+1

msporny:
Since already removed in PR #593 -- Editorial, multiple positive reviews, no changes requested, no objections, merging.

PR #654, opened 2021-02-16
Fix editorial and substantive issues in JSON-LD section.

msporny:
This PR is a fairly significant restructuring of the JSON-LD section. All the previous normative statements are more or less still active, save for the ones that are handled by other parts of the specification. I don't think implementations would need to change based on the modifications in this section, but I'm still marking it as a substantive change so people put a bit more diligence into reviewing it to make sure it does everything important that was being done before while attempting to simplify the language a bit (and cover some edge cases/advisements).

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/654.html" title="Last updated on Feb 17, 2021, 8:42 PM UTC (4654bf5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/654/d355bda...4654bf5.html" title="Last updated on Feb 17, 2021, 8:42 PM UTC (4654bf5)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2862,7 +2862,7 @@ maps</a>.
         <h3>Production</h3>
 
         <p>
-All data structures expressed by the <a href="#data-model">data model</a> MUST
+The <a>DID document</a> and any DID document data structures expressed by the <a href="#data-model">data model</a> MUST
 be serialized to the JSON-LD <a>representation</a> according to the JSON
 <a>representation</a> <a>production</a> rules as defined in <a
 href="#json"></a>.
@@ -2948,8 +2948,8 @@ href="#json"></a>.
 
         <p>
 In addition to using the JSON <a>representation</a> <a>consumption</a> rules,
-JSON-LD consumption MUST add the representation-specific entries into a
-representation-specific <a href="#data-model">data model</a> according to the
+JSON-LD consumption MUST add the representation-specific entries into
+the <a href="#data-model">data model</a> according to the
 JSON <a>representation</a> <a>consumption</a> rules.
         </p>
 
--- a/index.html
+++ b/index.html
@@ -2903,29 +2903,21 @@ rules.
 All implementers creating <a>conforming producers</a> that produce JSON-LD
 <a>representations</a> are advised to ensure that their algorithms
 produce valid JSON-LD [[JSON-LD11]] documents. Invalid JSON-LD documents will
-cause JSON-LD processors to produce errors.
+cause JSON-LD processors to halt and report errors.
         </p>
 
         <p>
 In order to achieve interoperability across different <a>representations</a>,
 all JSON-LD Contexts and their terms SHOULD be registered in the DID
-Specification Registries [[?DID-SPEC-REGISTRIES]]. If a term does not exist in
-the DID Specification Registries, then any <a>DID document</a> that uses it
-might not be interoperable across <a>representations</a>. It is RECOMMENDED that
-dereferencing each <a>URI</a> value of the <code>@context</code> entry results
-in a document containing machine-readable information about the JSON-LD Context.
+Specification Registries [[?DID-SPEC-REGISTRIES]].
         </p>
 
         <p>
 A <a>conforming producer</a> that generates a JSON-LD <a>representation</a>
-SHOULD NOT produce a <a>DID document</a> that contains properties not defined
+SHOULD NOT produce a <a>DID document</a> that contains terms not defined
 via the <code>@context</code> as <a>conforming consumers</a> might remove
-unknown properties.
-        </p>
-
-        <p>
-When serializing a JSON-LD <a>representation</a> of a <a>DID document</a>, a
-<a>conforming producer</a> MUST specify a media type of
+unknown terms. When serializing a JSON-LD <a>representation</a> of a <a>DID
+document</a>, a <a>conforming producer</a> MUST specify a media type of
 <code>application/did+ld+json</code> to downstream applications such as
 described in <a href="#did-resolution-metadata"></a>.
         </p>
@@ -2965,12 +2957,12 @@ JSON <a>representation</a> <a>consumption</a> rules.
 All implementers creating <a>conforming consumers</a> that consume JSON-LD
 <a>representations</a> are advised to ensure that their algorithms only accept
 valid JSON-LD [[JSON-LD11]] documents. Invalid JSON-LD documents will cause
-JSON-LD processors to produce errors.
+JSON-LD processors to halt and report errors.
         </p>
 
         <p>
-<a>Conforming consumers</a> that consume a JSON-LD <a>representation</a> SHOULD
-drop all properties from a <a>DID document</a> that are not defined via the
+<a>Conforming consumers</a> that process a JSON-LD <a>representation</a> SHOULD
+drop all terms from a <a>DID document</a> that are not defined via the
 <code>@context</code>.
         </p>
 
--- a/index.html
+++ b/index.html
@@ -2940,7 +2940,7 @@ Issue 208</a>.
         <h3>Consumption</h3>
 
         <p>
-All data structures expressed by a JSON-LD <a>representation</a> MUST be
+The <a>DID document</a> and any DID document data structures expressed by a JSON-LD <a>representation</a> MUST be
 deserialized into the <a href="#data-model">data model</a> according to the JSON
 <a>representation</a> <a>consumption</a> rules as defined in <a
 href="#json"></a>.
--- a/index.html
+++ b/index.html
@@ -2915,8 +2915,8 @@ Specification Registries [[?DID-SPEC-REGISTRIES]].
         <p>
 A <a>conforming producer</a> that generates a JSON-LD <a>representation</a>
 SHOULD NOT produce a <a>DID document</a> that contains terms not defined
-via the <code>@context</code> as <a>conforming consumers</a> might remove
-unknown terms. When serializing a JSON-LD <a>representation</a> of a <a>DID
+via the <code>@context</code> as <a>conforming consumers</a> are expected
+to remove unknown terms. When serializing a JSON-LD <a>representation</a> of a <a>DID
 document</a>, a <a>conforming producer</a> MUST specify a media type of
 <code>application/did+ld+json</code> to downstream applications such as
 described in <a href="#did-resolution-metadata"></a>.
--- a/index.html
+++ b/index.html
@@ -2833,20 +2833,104 @@ additional processing applied to its value, which will be added verbatim to the
     </section>
 
     <section>
-      <h2>
-JSON-LD
-      </h2>
+      <h2>JSON-LD</h2>
 
       <p>
-JSON-LD is a JSON-based format used to serialize
-<a href="http://www.w3.org/TR/ld-glossary/#linked-data">Linked Data</a>.
-This section establishes the requirements for producing and consuming <a>DID
-documents</a> that are expressed as JSON-LD. JSON-LD <a>DID documents</a>
-are indicated by a <code>contentType</code> resolver metadata field that is
-set to <code>application/did+ld+json</code>.
+JSON-LD [[JSON-LD11]] is a JSON-based format used to serialize <a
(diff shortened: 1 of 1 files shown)


OR13 (approved):

index.html
```diff
@@ -2833,136 +2833,105 @@ <h3>Consumption</h3>
...
-          </dd>
-        </dl>
         <p>
-Producers SHOULD NOT produce <a>DID documents</a> that contain properties not
-defined via the <code>@context</code>. Properties that are not defined via the
-<code>@context</code> MAY be dropped by Consumers.
+A <a>conforming producer</a> that generates a JSON-LD <a>representation</a>
+SHOULD NOT produce a <a>DID document</a> that contains terms not defined
+via the <code>@context</code> as <a>conforming consumers</a> might remove
+unknown terms. When serializing a JSON-LD <a>representation</a> of a <a>DID
```
rhiaro, line 2919:
```suggestion
via the <code>@context</code> as <a>conforming consumers</a> are expected
to remove unknown terms. When serializing a JSON-LD <a>representation</a> of a <a>DID
```

In Consumption it says dropping undefined terms is a SHOULD, so suggest strengthening the language here a bit to match.

rhiaro (approved):

index.html
```diff
@@ -2833,136 +2833,105 @@ <h3>Consumption</h3>
...
 
       <section>
         <h3>Production</h3>
 
         <p>
-The <a>DID document</a> MUST be serialized according to the
-<a href="#production">production rules for JSON</a>, with one additional
-requirement: The serialized <a>DID document</a> MUST include the
-<code>@context</code> member.
+All data structures expressed by the <a href="#data-model">data model</a> MUST
```
peacekeeper, line 2865: As I noted in https://github.com/w3c/did-core/pull/644#discussion_r576329987, it would be a major change to now apply the representations to other things than DID documents. The DID resolution functions and data structures were intentionally left abstract, for good reasons; let's not change that now.

index.html
```diff
@@ -2971,27 +2940,30 @@ <h3>Production</h3>
         <h3>Consumption</h3>
 
         <p>
-The <a>DID document</a> MUST be deserialized as a JSON document according to
-the <a href="#consumption">consumption rules for JSON</a>, with one additional
-requirement: The serialized <a>DID document</a> MUST include the
-<code>@context</code> member and be processed according to the rules below.
+All data structures expressed by a JSON-LD <a>representation</a> MUST be
```
peacekeeper, line 2943: Same comment as above. Data model and representations apply to DID documents, not other data structures.

index.html
```diff
@@ -2971,27 +2940,30 @@ <h3>Production</h3>
...
-<a href="#production-0">JSON-LD Production Rules</a>. If more than one
-<a>URI</a> is provided, the <a>URIs</a> MUST be interpreted as an
-<a data-cite="INFRA#ordered-set">ordered set</a>.
-            </p>
-          </dd>
-        </dl>
+        <p>
+In addition to using the JSON <a>representation</a> <a>consumption</a> rules,
+JSON-LD consumption MUST add the representation-specific entries into a
+representation-specific <a href="#data-model">data model</a> according to the
```
peacekeeper, line 2952: What's a "representation-specific data model"?

peacekeeper (changes requested):
Sorry but I'm against expanding the data model and representations to structures other than the DID document, and I'm also against the introduction of "representation-specific data models".

index.html
OR13: I assume its a `data-model` with representation specific properties... for example an abstract data model with cbor integers for keys.

index.html
OR13: technically and INFRA map with only an `id` being required could represent anything... and still be a DID Document... this is a side effect  / feature of the abstract data model and supporting unregistered properties.

index.html
OR13: I don't read this as changing the requirements... resolution returns an INFRA map with at least an `id`... you can't say anything other than that.... and have it still return a "DID Document" :)

msporny: > The DID resolution functions and data structures were intentionally left abstract, for good reasons; let's not change that now.

The new DID Resolution section made them concrete by saying the following: 

> If the resolution is successful, and if the resolve function was called, this MUST be a DID document abstract data model (a map) as described in § 4. Data Model that is capable of being transformed into a conforming DID Document (representation), using the production rules specified by the representation.

and

> The entire metadata structure MUST be serializable in a conforming representation.

I was very careful to make sure the language didn't preclude other ways of working with the didDocument or metadata (when dealing with it in programming environments) or requiring it to be serialized in a specific representation (serializing it to HTTP is fine, as long as you can ALSO serialize the same data to a representation).

This is not a big change... it allows you to keep things abstract if you want to (as long as there is a mechanism to get it to a representation -- and that makes sure everything in the abstract in DID Core is also serializable).

index.html
msporny: Agree with @OR13 -- we are using a data model that is capable of expressing all the data structures in the specification. A DID Document is one of those data structures. A verification method is another. A service is another. The ADM supports expressing all of those things (because it supports expressing maps). 

To put it another way... the Java language gives you the Object class... it doesn't then say, "You can only use the Object class to express DID Documents." <-- that would be an unnecessary restriction, which would mean that you'd need to create yet another thing to express all the other data structures you need to use.

As an aside, this is why creating bespoke abstract data models for specifications is such a bad idea -- you get pulled down into these esoteric, academic tarpits... but since we were all pulled into this tarpit, we do have a fairly generalized data model (thanks to INFRA) and it can be applied to all our data structures (because other people went through the pain of defining INFRA so that this was possible).

index.html
msporny: @OR13's is one interpretation of how we could use the data model... except that it wouldn't work because we can't have integers as keys there.

I meant something slightly different, in an attempt to make your diagrams work @peacekeeper -- think of it as having two instantiations of the data model... one to hold the DID Document, and another to hold representation-specific properties. Your diagrams would then have two boxes that production to JSON-LD would pull from... the DID Document data model, and the representation-specific entries data model.

You could even make the case that if you see the same entry in the DID Document data model, and the representation-specific entries data model... and you don't recognize the one in representation-specific entries as something specific to your representation, then you don't need to serialize it. HOWEVER, I expect that @OR13 and I would object to that, UNLESS, the JSON representation noted that you have to preserve `@context` (because that's what we do in our implementations).

OR13: if you don't want an `@context` don't consume JSON or JSON-LD that contains it.... craft your representation in the abstract and then produce ANY representation other than JSON-LD :)

peacekeeper:
@msporny @OR13 actually the idea of "representation (-specific) data models" does resonate with me a bit... A few months ago I suggested that we should have a "common data model" and multiple "representation data models". But in my mind the latter was a superset of the former, so not sure if that's aligned with your thinking.

For reference, here's the diagram I was using back then:

Screenshot from 2021-02-16 23-42-38

index.html
peacekeeper: > The entire metadata structure MUST be serializable in a conforming representation.

I will note that there was an objection to adding this language (see https://github.com/w3c/did-core/pull/601#discussion_r573031499), and that this language was added without addressing the objection, and with very little review by the WG.

msporny:
> But in my mind the latter was a superset of the former (not a separate bucket), so not sure if that's aligned with your thinking.

Yes, aligned w/ my thinking -- but I was trying to not introduce yet another concept into the specification. Even if we were to add the concept of a "DID Document data model", and "Representation data models", we would still need yet another data model, which is shared among `produce` calls. @OR13 described it as `productionOptions`, which would contain stuff like `@context`. I was thinking more along the lines of `representationEntries` which would contain things like `@context`.

I think we're circling around the same concepts, but at this point, we need to settle on a way of talking about it in the spec. 

What I could do is clean this PR up and pretend we didn't have the discussion above... raise an issue, and then in a separate PR try to make the whole "different data models and/or productionOptions and/or representationEntries" work across the data model and all representations? In other words, make a pass to get stuff in a decent shape to then make these changes above easier to see/reason about. Thoughts?

index.html
msporny: That objection was resolved here: https://github.com/w3c/did-core/pull/601#issuecomment-777092645

msporny, line 2865:
```suggestion
The <a>DID document</a> and any DID document data structures expressed by the <a href="#data-model">data model</a> MUST
```

index.html
msporny: @peacekeeper I'll put the language back to what I believe you're comfortable with in https://github.com/w3c/did-core/pull/654/files#r577651396 -- this means that the only valid data structures are the DID Document and it's data structures. As a result, no change is needed for this text.

index.html
msporny, line 2952:
```suggestion
JSON-LD consumption MUST add the representation-specific entries into
the <a href="#data-model">data model</a> according to the
```

OR13:
I left my proposal here: https://github.com/w3c/did-core/issues/662

I don't see a need for more than 1 bucket of entires in the ADM for a number of reasons...

The most compelling one is simplicity... as @selfissued  would say, this spec should be easy for developers to implement... if we start making 3 buckets that all get used its worse than 2 which is worse than 1.

The spec as written today has only 1 bucket, and while that makes representation specific data models awkward... it also makes them very simple.... our default stance should be no change.

If we can gain consensus on a change, it should remove awkwardness without increasing complexity... in my mind the only way do to that is by altering the signature of production.... but let's look at the alternatives:

### Option 1 (no change)

```ts
const produce = (adm) => Buffer
const consume = (Buffer) => adm
```

### Option 2 (my proposal on #662 )

```ts
const produce = (adm, representation) => Buffer
const consume = (Buffer) => adm
```

### Option 3 

```ts
const produce = (adm, representation) => Buffer // we agreed representation specific properties would be preserved, this will open the door to dropping properties again... which will lead to implementation differencs.
const consume = (Buffer) => adm, representation // here we will fail to agree on what is representation and what is adm.... and so will all future implementers... this will lead to differences in what ends in the ADM... and undermine its usefulness.
```

### Option 4 (for completeness)

```ts
const produce = (adm) => Buffer // this doesn't address to objections
const consume = (Buffer) => adm, representation
```

My preference is for option 1, and maybe option 2....I think 3 is a really bad idea and undermines the entire point of a "big tent" / shared abstract data model.... I would need to see actual code to be convinced otherwise... we are past the point of text being convincing.

index.html
msporny: @peacekeeper I've put this back and raised issue #664 to address this one. @OR13 has also raised #662 with a slightly different approach. I would prefer that we merge this PR as is (because it implements current consensus, even though none of us like it), and then we do a future PR to fix this once and for all once issues #662 and #664 are settled.

index.html
msporny, line 2943:
```suggestion
The <a>DID document</a> and any DID document data structures expressed by a JSON-LD <a>representation</a> MUST be
```

msporny: @peacekeeper applied your suggestion in https://github.com/w3c/did-core/pull/654/files#r577928397 to address your objection.

msporny:
Substantive, multiple reviews, changes requested and made, objection raised by @peacekeeper and addressed via spec text change, no remaining objections, merging.

index.html
```diff
@@ -2833,136 +2833,105 @@ <h3>Consumption</h3>
...
-          <dt>@context</dt>
-          <dd>
-
-            <p>
-The <a href="https://www.w3.org/TR/json-ld11/#the-context">JSON-LD
-specification</a> defines values that are valid for this member. It
-contains representation-specific syntax and therefore could be present as an
-entry in the <a
-href="#data-model">data model</a> to aid in lossless conversion. If the entry
-is present in the <a href="#data-model">data model</a>, it MUST be used during
```
peacekeeper, line 2906: For the record, this language was added quite recently in https://github.com/w3c/did-core/pull/454 after much discussion, and it has been important for those of us who DON'T think `@context` is a REQUIRED entry in the abstract data model. Removing this part could therefore be problematic. But it looks like in https://github.com/w3c/did-core/issues/664 we're on a good path to make everyone happy!

PR #663, opened 2021-02-17
Fix editorial and substantive issues in CBOR section.

msporny:
This PR reworks the CBOR section in the same way the JSON and JSON-LD sections were reworked. The changes are largely editorial or non-substantive, but there is one change that is substantive. 

There were rules regarding preserving "unknown data types"... except that there is no such thing in the ADM... we either know about the data type, or we can't represent it in the ADM. It is impossible for an "unknown data type" to exist in the ADM... all data types are known, or they're not representable. Similarly, it is impossible for a CBOR consumer to consume data with an unknown data type and preserve it in the ADM... because there is nowhere to put it (because the data type is unknown). You can't convert it to a string (which is what was specified) because the production rules would spit it out as a string (it wasn't a string when it was consumed -- so you'd get data corruption). I removed that normative statement because it was unimplementable.

There is also a rule that I'm concerned about, and that we handwave over, but is required if we want to preserve the CBOR Canonical Rules requirement. There is a requirement in CBOR Canonical Rules that you must specify whether you preserve CBOR Tags or not. If you don't, you can't round-trip from CBOR to some internal representation and back out to CBOR. The problem is that the ADM has no way to preserve the incoming CBOR Tags. So, the current consumption rules say: "CBOR tags MUST be retained for CBOR production". The production rules say "All CBOR tags MUST be retained regardless of whether they are optional." The statements are testable because you can provide a CBOR document with tags and fail if the same tags don't get added when you produce CBOR. In other words, if you can't round-trip, you don't have a conforming CBOR implementation... but honestly, this is a really high bar and it's not clear what we're getting out of it. The same is true for the CBOR Canonical Rules -- they're not easy to do exhaustive tests on... but I retained them because @jonnycrunch was so intent on having them in the spec.

So, keep an eye out for those items as you're reviewing.

/cc @jonnycrunch

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/663.html" title="Last updated on Feb 18, 2021, 2:30 PM UTC (4b201b5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/663/24a6c92...4b201b5.html" title="Last updated on Feb 18, 2021, 2:30 PM UTC (4b201b5)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2948,7 +2948,7 @@ section will be removed from this specification.
         <h3>Production</h3>
 
         <p>
-All data structures expressed by the <a href="#data-model">data model</a>
+All <a>DID document</a> data structures expressed by the <a href="#data-model">data model</a>
 MUST be serialized to the CBOR <a>representation</a> according to the
 following <a>production</a> rules:
         </p>
--- a/index.html
+++ b/index.html
@@ -2931,6 +2931,12 @@ defined via the <code>@context</code>.
 
     <section>
       <h2>CBOR</h2>
+
+      <p>
+This section defines the <a>production</a> and <a>consumption</a> rules
+for the CBOR <a>representation</a>.
+      </p>
+
       <p class="issue atrisk">
 The Working Group is seeking volunteers to write tests for, and at least two
 independent and interoperable implementations of, the CBOR representation
@@ -2938,32 +2944,13 @@ during the Candidate Recommendation phase. If these goals are not met, this
 section will be removed from this specification.
       </p>
 
-      <p>
-Like Javascript Object Notation (JSON) [[RFC8259]], Concise Binary Object
-Representation (CBOR) [[RFC8949]] defines a set of formatting rules for the
-portable representation of structured data. CBOR is a more concise,
-machine-readable, language-independent data interchange format that is
-self-describing and has built-in semantics for interoperability.
-      </p>
-
-      <p>
-The following sections outline the rules for producing and consuming
-<a>DID documents</a> that are expressed in CBOR as indicated by a
-<code>contentType</code> of <code>application/did+cbor</code> in the resolver
-metadata.
-      </p>
-
       <section>
         <h3>Production</h3>
 
         <p>
-A <a>DID document</a> MUST be a single <a data-cite="RFC8949#section-3.1">CBOR
-Map</a> conforming to [[RFC8949]]. All topmost entries of the <a>DID
-document</a> MUST be represented by using the entry key as the name of the
-key of the CBOR Map. The values of entries of the <a href="#data-model">data
-model</a> described in Section <a href="#data-model"></a>, including all
-extensions, are encoded in CBOR [[RFC8949]] by mapping entry values to CBOR
-types as follows:
+All data structures expressed by the <a href="#data-model">data model</a>
+MUST be serialized to the CBOR <a>representation</a> according to the
+following <a>production</a> rules:
         </p>
 
         <table class="simple" id="cbor-representation-production">
@@ -2984,11 +2971,10 @@ CBOR Representation Type
 <a data-cite="INFRA#maps">ordered&nbsp;map</a>
               </td>
               <td>
-<a data-cite="RFC8949#section-3.1">CBOR map</a> (major type 5), each entry
-is represented as a member of the CBOR Map with the entry key, expressed as
-a <a data-cite="RFC8949#section-3.1">CBOR string</a> (major type 3</a>),
-as the key and the entry value according to its type, as defined in this
-section
+A <a data-cite="RFC8949#section-3.1">CBOR map</a> (major type 5), where each
+entry is represented as a member of the CBOR map. The entry key is expressed
+as a <a data-cite="RFC8949#section-3.1">CBOR string</a> (major type 3) as the
+key and the entry value according to its type, as defined in this table.
               </td>
             </tr>
             <tr>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -2931,39 +2931,26 @@ <h3>Consumption</h3>
...
 
         <p>
-A <a>DID document</a> MUST be a single <a data-cite="RFC8949#section-3.1">CBOR
-Map</a> conforming to [[RFC8949]]. All topmost entries of the <a>DID
-document</a> MUST be represented by using the entry key as the name of the
-key of the CBOR Map. The values of entries of the <a href="#data-model">data
-model</a> described in Section <a href="#data-model"></a>, including all
-extensions, are encoded in CBOR [[RFC8949]] by mapping entry values to CBOR
-types as follows:
+All data structures expressed by the <a href="#data-model">data model</a>
```
msporny, line 2951:
```suggestion
All <a>DID document</a> data structures expressed by the <a href="#data-model">data model</a>
```

index.html
```diff
@@ -2984,39 +2971,38 @@ <h3>Production</h3>
...
               <td>
-<a data-cite="RFC8949#section-3.1">CBOR map</a> (major type 5), each entry
-is represented as a member of the CBOR Map with the entry key, expressed as
-a <a data-cite="RFC8949#section-3.1">CBOR string</a> (major type 3</a>),
-as the key and the entry value according to its type, as defined in this
-section
+A <a data-cite="RFC8949#section-3.1">CBOR map</a> (major type 5), where each
+entry is represented as a member of the CBOR map. The entry key is expressed
+as a <a data-cite="RFC8949#section-3.1">CBOR string</a> (major type 3) as the
+key and the entry value according to its type, as defined in this table.
```
TallTed, line 2977:
```suggestion
key, and the entry value according to its type, as defined in this table.
```

index.html
```diff
@@ -3054,45 +3040,57 @@ <h3>Production</h3>
...
+        <p>
+In addition to the data type production rules above, the following rules apply
+for <a>conforming producers</a> that serialize CBOR <a>representations</a>:
+        </p>
+        <ol>
           <li>
-Other Data Types not otherewise defined in the <a href="#data-model">data model</a> MUST be
-represented as a <a data-cite="RFC8949#section-3.1">CBOR
-String type</a> (major type 3).
+Indefinite-length items are not allowed and MUST be made <a
```
TallTed, line 3065:
```suggestion
Indefinite-length items are not allowed and MUST be made a <a
```

index.html
```diff
@@ -3054,45 +3040,57 @@ <h3>Production</h3>
...
         <p>
-In addition to the production rules in this section, implementations
-MUST implement the four
-<a data-cite="RFC8949#section-4.2.1">Canonical CBOR rules</a>
-listed in [[RFC8949]]. All CBOR tags that appear in a message MUST be retained
-regardless of whether they are optional.
+All entries of a <a>DID document</a> MUST be included in the root <a
+data-cite="RFC8949#section-3.1">CBOR map</a> (major type 5). Entries MAY contain
+additional data substructures subject to the value representation rules in the
+list above. When serializing a <a>DID document</a> in the CBOR
```
TallTed, line 3081:
```suggestion
list above. When serializing a <a>DID document</a> to its CBOR
```

TallTed: `its` might be `a` if there's more than one potential CBOR representation of a single state of a DID document... but I think there's only one.

index.html
```diff
@@ -3054,45 +3040,57 @@ <h3>Production</h3>
...
+<a>representation</a>, a <a>conforming producer</a> MUST specify a media type of
+<code>application/did+cbor</code> to downstream applications such as described
+in <a href="#did-resolution-metadata"></a>.
+        </p>
+
+        <p>
+The following examples express a CBOR <a>representation</a> of a
```
TallTed, line 3088:
```suggestion
The following examples express the CBOR <a>representation</a> of a
```
As above -- if there's more than one possible CBOR representation of a single state of a DID document, then `a` should remain; if there's only one possible CBOR representation of a single state, then this change is appropriate.

index.html
```diff
@@ -3247,30 +3243,42 @@ <h3>Consumption</h3>
...
+All CBOR tags MUST be retained for CBOR <a>production</a> regardless of whether
+they are optional.
           </li>
-        </ul>
+        </ol>
+
+        <p>
+If media type information is available to a <a>conforming consumer</a> and the
+media type value is <code>application/did+json</code>: the data structure  being
+consumed is a <a>DID document</a> and the root element MUST be a <a
```
TallTed, line 3274:
```suggestion
media type value is <code>application/did+json</code>, then the data structure being
consumed is a <a>DID document</a>, and the root element MUST be a <a
```

TallTed (changes requested):
A number of small suggestions, with a few that have notes where I'm not quite certain I grok the totality, so my suggestion might not be appropriate.

brentzundel (approved):

rhiaro (approved):

index.html
msporny: Yeah, the language here is tricky. There are at least two CBOR-based representations... Vanilla CBOR (this spec) and CBOR-LD (a future state, but with an implementation and DID document compression test). We are writing this in the CBOR section, so hopefully the reader understands that the "CBOR" we're talking about is "the CBOR representation described in this section".

All that said, I'm merging THIS suggestion.

msporny:
Substantive, multiple reviews, changes requested and made, no objections, merging.

Issue #664, opened 2021-02-17
DID Document, Representation, and Representation-Entries data model?


msporny:
At present, the DID Core specification only has one data model -- the one that's used to express DID Documents and their sub data structures (verification methods, services, identifiers, etc.)

This has always given a subset of the group heartburn -- @peacekeeper has wanted to separate representation-specific entries from the DID Document entries. There is a valid argument to do so, in order to not mix one set of map entries with another set of map entries.

There is also an open question related to: "What data model holds the data structures that you have after you process the input bytes during consumption?" For example, our current CBOR rules require you to preserve CBOR Tags, but we do a giant handwave over exactly how that's accomplished. It would typically be accomplished by working on the map data structures that you get after you do CBOR.decode(...) in most libraries.

So, this raises two questions:

1. Do we want to introduce the concept of a **representation data model**, which will be vaguely defined with no normative statements associated with it, but can answer the question of "What data model holds the data structures that you have after you process the input bytes during consumption?"
2. Do we want to introduce the concept of a **representation-entries data model**, which would hold representation-specific entries that could be used by any production process and would aid in cross-representation conversion w/o dumping representation-specific garbage into the **DID Document** data model?

OR13:
See also : https://github.com/w3c/did-core/pull/654#issuecomment-780635390

OR13:
1. is probably DOA... I think we will not be able to agree to 1 (some of us think JSON-LD and JSON Schema are JSON, others don't).... I think we have beaten this topic to death, multiple special topic calls, many resolutions, including preserving unregistered properties, etc... I really don't want to re-litigate this issue... especially at the last minute.

2. I can get behind 2 but only as long as it stays limited to production.... and its still dangerous / harmful complexity...

TallTed:
@OR13 - The very long inline comments on Option 3 are hard to digest in this form. Would you please break those comments into multiple shorter lines?

peacekeeper:
From my perspective, Option 3.1 and Option 3.3 would both work. In my mind it has always been like that anyway. There is an abstract data model that contains services, verification methods, etc. And then there are representation-specific things such as CBOR tags or `@context` that should not be mixed with the abstract (representation-independent) things.

Both Options would also allow @OR13 and @msporny to put `@context` into an `application/did+json` DID document (even though personally I think that's a bad idea that contradicts why we introduced the abstract data model concept in the first place).

msporny:
[...]

PR #678, opened 2021-02-19
Implement Data Model / Resolution serialization mechanisms.

msporny:
Implements the following WG decision:

> RESOLUTION: The ADM is only used to represent a DID document or parts of a DID document (e.g. a verification method). For DID Resolution data structures, INFRA will be referenced normatively, as well as the JSON serialization rules for INFRA, to note that the data structure must be able to be serialized using INFRA->JSON rules.

https://www.w3.org/2021/02/16-did-topic-minutes.html#r01

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/678.html" title="Last updated on Feb 22, 2021, 2:32 PM UTC (919d1ba)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/678/793057f...919d1ba.html" title="Last updated on Feb 22, 2021, 2:32 PM UTC (919d1ba)">Diff</a>

PR #679, opened 2021-02-19
Add representation-specific entry map in Production and Consumption.

msporny:
This PR attempts to address @peacekeeper's long standing concern wrt. production and consumption and representation-specific entries by using the proposal for Option 3.3 in issue https://github.com/w3c/did-core/issues/664#issuecomment-780647353. This is also an attempt to get PRs #596 and #597 unstuck. 

Specifically, it defines the input and return parameters for conforming producers and consumers and splits the representation-specific entries into their own map that can be serialized and returned along with the DID Document data model. It adds a requirement that all representations detect known representation-specific entries, which is required to cleanly create the two classes of information. The change is substantive.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/679.html" title="Last updated on Feb 22, 2021, 2:47 PM UTC (bd42543)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/679/4fa8635...bd42543.html" title="Last updated on Feb 22, 2021, 2:47 PM UTC (bd42543)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1134,8 +1134,9 @@ an absolute <a>DID URL</a> value of
 This specification defines a data model that can be used to express <a>DID
 documents</a> and DID document data structures, which can then be serialized
 into multiple concrete <a>representations</a>. This section provides a
-high-level description of the data model, how different types of properties are
-expressed in the data model, and instructions for extending the data model.
+high-level description of the data model, descriptions of the ways different
+types of properties are expressed in the data model, and instructions for
+extending the data model.
     </p>
     <p>
 A <a>DID document</a> consists of a <a data-cite="INFRA#maps">map</a> of <a
@@ -2823,7 +2824,7 @@ maps</a>.
 
         <p>
 The <a>DID document</a>, DID document data structures, and
-representation-specific entries <a data-cite="INFRA#maps">map</a> MUST be MUST
+representation-specific entries <a data-cite="INFRA#maps">map</a> MUST
 be serialized to the JSON-LD <a>representation</a> according to the JSON
 <a>representation</a> <a>production</a> rules as defined in <a
 href="#json"></a>.
--- a/index.html
+++ b/index.html
@@ -2419,11 +2419,11 @@ href="#data-model">data model</a> is lossless. For example, some CBOR-based
 represent the number of seconds since the Unix epoch.
         </li>
         <li>
-A <a>representation</a> MAY define representation-specific entries that can be
-stored in a data structure expressible using the <a href="#data-model">data
-model</a> for use during the <a>production</a> and <a>consumption</a> process.
-These entries are used when consuming or producing to aid in ensuring
-lossless conversion.
+A <a>representation</a> MAY define representation-specific entries that are
+stored in a representation-specific entries <a data-cite="INFRA#maps">map</a>
+for use during the <a>production</a> and <a>consumption</a> process. These
+entries are used when consuming or producing to aid in ensuring lossless
+conversion.
         </li>
         <li>
 In order to maximize interoperability, <a>representation</a> specification
@@ -2432,34 +2432,30 @@ Registries [[?DID-SPEC-REGISTRIES]].
         </li>
       </ol>
 
-      <p class="note" title="Representation-specific entries">
-Note that representation-specific entries will only have special processing
-rules defined by a single <a>representation</a>. Consumers of a different
-<a>representation</a> are required to preserve these entries in the
-<a>DID document</a> <a
-href="#data-model">data model</a> using their generic type processing rules
-to enable lossless conversion of <a>representations</a>. Similarly, producers
-are required to treat entries containing representation-specific syntax using
-generic type processing rules when producing a <a>representation</a> for which
-the entry is not defined. <a>Representations</a> are required to define producer
-behavior for any such entries defined by the <a>representation</a>.
-      </p>
-
       <p>
 The requirements for all <a>conforming producers</a> are as follows:
       </p>
 
       <ol>
         <li>
-A <a>conforming producer</a> MUST serialize all entries in the <a
-href="#data-model">data model</a> that do not have explicit processing rules for
-the <a>representation</a> being produced using only the <a>representation</a>'s
-data type processing rules.
+A <a>conforming producer</a> MUST take a <a>DID document</a> <a
+href="#data-model">data model</a> and a representation-specific entries <a
+data-cite="INFRA#maps">map</a> as input into the <a>production</a> process.
+The <a>conforming producer</a> MAY accept additional options as input
+into the <a>production</a> process.
         </li>
         <li>
-A <a>conforming producer</a> MUST indicate which <a>representation</a> has been
-used for a <a>DID document</a> via a Media Type as described in <a
-href="#did-resolution-metadata"></a>.
+A <a>conforming producer</a> MUST serialize all entries in the <a>DID
+document</a> <a href="#data-model">data model</a>, and the
+representation-specific entries <a data-cite="INFRA#maps">map</a>, that do not
+have explicit processing rules for the <a>representation</a> being produced
+using only the <a>representation</a>'s data type processing rules and
+return the serialization after the <a>production</a> process completes.
+        </li>
+        <li>
+A <a>conforming producer</a> MUST return the Media Type <a
+data-cite="INFRA#string">string</a> associated with the <a>representation</a>
+after the <a>production</a> process completes.
         </li>
       </ol>
 
@@ -2469,15 +2465,30 @@ The requirements for all <a>conforming consumers</a> are as follows:
 
       <ol>
         <li>
-A <a>conforming consumer</a> MUST add all entries that do not have explicit
-processing rules for the <a>representation</a> being consumed to the <a
-href="#data-model">data model</a> using only the <a>representation</a>'s data
-type processing rules.
+A <a>conforming consumer</a> MUST take a <a>representation</a> and
+Media Type <a data-cite="INFRA#string">string</a> as input into
+the <a>consumption</a> process. A <a>conforming consumer</a> MAY accept
+additional options as input into the <a>consumption</a> process.
         </li>
         <li>
 A <a>conforming consumer</a> MUST determine the <a>representation</a> of a
-<a>DID document</a> using the associated Media Type as described in <a
-href="#did-resolution-metadata"></a>.
+<a>DID document</a> using the Media Type input <a
+data-cite="INFRA#string">string</a>.
+        </li>
+        <li>
+A <a>conforming consumer</a> MUST detect any representation-specific
+entry across all known <a>representations</a> and place the entry into a
+representation-specific entries <a data-cite="INFRA#maps">map</a> which is
+returned after the <a>consumption</a> process completes. A list of
+all known representation-specific entries is available in the
+DID Specification Registries [[DID-SPEC-REGISTRIES]].
+        </li>
+        <li>
+A <a>conforming consumer</a> MUST add all non-representation-specific entries
+that do not have explicit processing rules for the <a>representation</a> being
+consumed to the <a>DID document</a> <a href="#data-model">data model</a> using
+only the <a>representation</a>'s data type processing rules and return the
+<a>DID document</a> data model after the <a>consumption</a> process completes.
         </li>
       </ol>
 
@@ -2503,9 +2514,10 @@ for the JSON <a>representation</a>.
         <h3>Production</h3>
 
         <p>
-The <a>DID document</a> and any DID document data structures expressed by the <a href="#data-model">data model</a>
-MUST be serialized to the JSON <a>representation</a> according to the
-following <a>production</a> rules:
+The <a>DID document</a>, DID document data structures, and
+representation-specific entries <a data-cite="INFRA#maps">map</a> MUST be
+serialized to the JSON <a>representation</a> according to the following
+<a>production</a> rules:
         </p>
 
         <table class="simple" id="json-representation-production">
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -2470,15 +2466,30 @@ <h2>Production and Consumption</h2>
...
 A <a>conforming consumer</a> MUST determine the <a>representation</a> of a
-<a>DID document</a> using the associated MIME type as described in <a
-href="#did-resolution-metadata"></a>.
+<a>DID document</a> using the MIME type input <a
+data-cite="INFRA#string">string</a>.
+        </li>
+        <li>
+A <a>conforming consumer</a> MUST detect any representation-specific
+entry across all known <a>representations</a> and place the entry into a
+representation-specific entries <a data-cite="INFRA#maps">map</a> which is
```
msporny, line 2482: I had played around with alternative language here that effectively said: "any representation specific entries that could appear in the representation" -- which basically boils down to just `@context`... `%YAML` and `xmlns` are highly unlikely to appear in JSON and CBOR.

index.html
```diff
@@ -2811,7 +2823,8 @@ <h2>JSON-LD</h2>
         <h3>Production</h3>
 
         <p>
-The <a>DID document</a> and any DID document data structures expressed by the <a href="#data-model">data model</a> MUST
+The <a>DID document</a>, DID document data structures, and
+representation-specific entries <a data-cite="INFRA#maps">map</a> MUST be MUST
```
msporny, line 2827:
```suggestion
representation-specific entries <a data-cite="INFRA#maps">map</a> MUST be
```

OR13 (changes requested):
I would prefer to see abstract function signatures for consume and produce.

something like:

``` 
produce = (didDocument, representation-entries, representation-options) => didDocumentByteStream
consume = (didDocumentByteStream) => didDocument, representation-entries
```

I don't understand why we define abstract function for resolution, and then not for production and consumption.

`representation-entries` and `representation-options` need to be defined next to `didDocumnt` and `didDocumentByteStream`

If an issue is added pointing to this comment, I am happy to approve and do the PR myself.

index.html
```diff
@@ -2433,34 +2433,30 @@ <h2>Production and Consumption</h2>
...
+The <a>conforming producer</a> MAY accept additional options as input
+into the <a>production</a> process.
+        </li>
+        <li>
+A <a>conforming producer</a> MUST serialize all entries in the <a>DID
+document</a> <a href="#data-model">data model</a>, and the
+representation-specific entries <a data-cite="INFRA#maps">map</a>, that do not
+have explicit processing rules for the <a>representation</a> being produced
+using only the <a>representation</a>'s data type processing rules and
+return the serialization after the <a>production</a> process completes.
```
TallTed, line 2454: After reading this over several times, I *think* I've figured out a small
set of changes that make it clearer after fewer readings, hopefully down
to one read for sufficient clarity to act on.
```suggestion
A <a>conforming producer</a> MUST serialize all entries in the <a>DID
document</a> <a href="#data-model">data model</a>, and the relevant
entries in the representation-specific entries <a
data-cite="INFRA#maps">map</a>, that do not have explicit processing
rules for the <a>representation</a> being produced, using only the
<a>representation</a>'s data type processing rules; and return the
serialization after the <a>production</a> process completes.
```

index.html
```diff
@@ -2433,34 +2433,30 @@ <h2>Production and Consumption</h2>
...
         </li>
         <li>
-A <a>conforming producer</a> MUST indicate which <a>representation</a> has been
-used for a <a>DID document</a> via a MIME type as described in <a
-href="#did-resolution-metadata"></a>.
+A <a>conforming producer</a> MUST return the MIME type <a
```
TallTed, line 2457:
```suggestion
A <a>conforming producer</a> MUST return the Media Type <a
```
_MIME Types_ became _Media Types_ with [RFC 1590](https://tools.ietf.org/html/rfc1590) in 1994.  (There are several, if not many, places this change is needed.)

index.html
```diff
@@ -2470,15 +2466,30 @@ <h2>Production and Consumption</h2>
 
       <ol>
         <li>
-A <a>conforming consumer</a> MUST add all entries that do not have explicit
-processing rules for the <a>representation</a> being consumed to the <a
-href="#data-model">data model</a> using only the <a>representation</a>'s data
-type processing rules.
+A <a>conforming consumer</a> MUST take a <a>representation</a> and
+MIME type <a data-cite="INFRA#string">string</a> as input into
```
TallTed, line 2470:
```suggestion
Media Type <a data-cite="INFRA#string">string</a> as input into
```
or "media type" ... Capitalization may be important for clarity.

index.html
```diff
@@ -2811,7 +2823,8 @@ <h2>JSON-LD</h2>
         <h3>Production</h3>
 
         <p>
-The <a>DID document</a> and any DID document data structures expressed by the <a href="#data-model">data model</a> MUST
+The <a>DID document</a>, DID document data structures, and
+representation-specific entries <a data-cite="INFRA#maps">map</a> MUST be MUST
 be serialized to the JSON-LD <a>representation</a> according to the JSON
```
TallTed, line 2828:
```suggestion
representation-specific entries <a data-cite="INFRA#maps">map</a> MUST be
serialized to the JSON-LD <a>representation</a> according to the JSON
```

index.html
TallTed: `MUST be MUST be` becomes `MUST be be` ... so take my suggestion below instead

msporny:
> I don't understand why we define abstract function for resolution, and then not for production and consumption.

With my Editor's hat firmly on: We never should have defined abstract functions for resolution using our own bespoke abstract function language... I cannot underscore how terrible the half-measure that is the Resolution section is -- we're bending over backwards and doing really strange stuff there to half-define something. We should have either just kept all of it to a testable algorithm in prose, like https://www.w3.org/TR/json-ld11-api/#expansion-algorithm does... or actually defined a concrete set of functions using Javascript/Typescript/WebIDL. What we have is a weird half-measure and I'm reluctant to double-down on that mistake.

I did start to write that up initially, and then decided against it because...

> representation-options

We have no way to realize this parameter in the way the Resolution section is written (which again, is strange and bespoke and isn't clearly specified for anything that is not INFRA->JSON... like, you know, CBOR). None of the data types in the data model or INFRA support it. Consider the use case of injecting encoded CBOR values into the root CBOR map... integers as keys -- can't do it... no concrete representation in INFRA for CBOR. So the best I could do is use prose to say that you can take other options and those options can be used to affect production. We're not testing MAY statements... that's the only way I could get to making representation-options a possibility.

It's not that it's impossible to write the spec up with bespoke abstract function language... but it makes it worse, IMHO, not better. All that said, I'll draft a PR that does the abstract function thing (even though I'm totally against the concept... what we have now is testable and flexible enough to support representation-options). If I write it up like we did for Resolution, representation-options are still not going to be a thing because we have no way of expressing them.

index.html
msporny: Yes, I'm going to do this in a big spec-wide PR in one go. Rejecting your change suggestion, even though it's correct, want to bundle all those MIME type -> Media Type changes together /after/ the spec has settled.

index.html
msporny: Rejecting per https://github.com/w3c/did-core/pull/679/files#r579329532

index.html
msporny: Done.

index.html
TallTed: Fair ... though I don't know what would be lost by doing some of these changes sooner, and think that having a mix of MIME & Media would be more likely to flag the need to replace the former than having all MIME would.  I'll count on you having a list of such "will do these later in batches" changes.

index.html
msporny: Fixed in d915bea303b36c00ec59b27b8ac10845da828fea

index.html
msporny: Fixed in d915bea303b36c00ec59b27b8ac10845da828fea.

index.html
```diff
@@ -1134,10 +1134,10 @@ <h2>Relative DID URLs</h2>
...
     <p>
 This specification defines a data model that can be used to express <a>DID
-documents</a> and other data structures that can then be serialized into
-multiple concrete <a>representations</a>. This section provides a high-level
-description of the data model, how different types of properties are expressed
-in the data model, and instructions for extending the data model.
+documents</a> and DID document data structures, which can then be serialized
+into multiple concrete <a>representations</a>. This section provides a
+high-level description of the data model, how different types of properties are
+expressed in the data model, and instructions for extending the data model.
```
TallTed, line 1140:
```suggestion
high-level description of the data model, descriptions of the ways different
types of properties are expressed in the data model, and instructions for
extending the data model.
```

index.html
```diff
@@ -2817,7 +2823,8 @@ <h2>JSON-LD</h2>
         <h3>Production</h3>
 
         <p>
-The <a>DID document</a> and any DID document data structures expressed by the <a href="#data-model">data model</a> MUST
+The <a>DID document</a>, DID document data structures, and
+representation-specific entries <a data-cite="INFRA#maps">map</a> MUST be MUST
```
brentzundel, line 2827:
```suggestion
representation-specific entries <a data-cite="INFRA#maps">map</a> MUST
```

brentzundel (approved):
I believe this PR includes changes from PR #678. Not sure if that was deliberate.

OR13:
> We should have either just kept all of it to a testable algorithm in prose, like https://www.w3.org/TR/json-ld11-api/#expansion-algorithm does... or actually defined a concrete set of functions using Javascript/Typescript/WebIDL.

I agree... it's not too late to fix this ambiguity. I personally find function definitions more useful than english sentences....

```
didDocument =  « 
didDocumentEntries: INFRA<map>, 
didDocumentRepresentationEntries: INFRA<map> 
»

resolve(
did: INFRA<string>, 
resolutionOptions: INFRA<map>
) →
   « 
   didResolutionMetadata: INFRA<map>, 
   didDocument: INFRA<map>, 
   didDocumentMetadata: INFRA<map> 
   »
   
resolveRepresentation(
did: INFRA<string>, 
resolutionOptions: INFRA<map>
) →
   « 
   didResolutionMetadata: INFRA<map>, 
   didDocumentStream: INFRA<ByteStream>, 
   didDocumentMetadata: INFRA<map> 
   »
      
consume(
contentType: INFRA<string>, 
didDocumentStream: INFRA<ByteStream>
) →
  « 
  didDocumentEntries: INFRA<map>, 
  representationEntries: INFRA<map> 
  »
  
produce(
contentType: INFRA<string>, 
didDocumentEntries: INFRA<map>, 
didDocumentRepresentationEntries: INFRA<map>
) →
  « 
  didDocumentStream: INFRA<ByteStream> 
  »
```

msporny:
@OR13 wrote:
> I personally find function definitions far more useful than english sentences....

Sure, I do as well, and W3C has a formal language in which to do that... it's called WebIDL... which people objected to using.

Instead, we've got a interface definition language that we invented on top of INFRA -- totally bespoke, no one understands it except for us... and we're using that to define the interfaces in the specification, which BTW, we're totally not chartered to do. The WG decided to look the other way wrt. the Resolution section on that particular point.

msporny:
This PR is blocked waiting on a review from: @peacekeeper and @jricher

index.html
```diff
@@ -2470,15 +2466,30 @@ <h2>Production and Consumption</h2>
...
         <li>
 A <a>conforming consumer</a> MUST determine the <a>representation</a> of a
-<a>DID document</a> using the associated Media Type as described in <a
-href="#did-resolution-metadata"></a>.
+<a>DID document</a> using the Media Type input <a
+data-cite="INFRA#string">string</a>.
+        </li>
+        <li>
+A <a>conforming consumer</a> MUST detect any representation-specific
+entry across all known <a>representations</a> and place the entry into a
```
peacekeeper, line 2481: Perhaps we want to say "MUST detect any representation-specific entry known to the representation being consumed", and "SHOULD/MAY detect representation-specific entries across all other known representations"? Anyway I'm also fine with this as-is.

peacekeeper (approved):
> bespoke abstract function language

I don't have the same spec writing experience you do, so I'm not sure what other specs do when you want to define an interface (with inputs and outputs) for a process that can have many concrete realizations.

We can still add examples for WebIDL (https://github.com/w3c/did-core/pull/587) and TypeScript (https://github.com/w3c/did-core/pull/591), I think that would be useful as long as it's clear that they are just examples, not the authoritative definitions.

I also like @OR13 's proposal in https://github.com/w3c/did-core/pull/679#issuecomment-782481740 to add function definitions for `produce()` and `consume()`. We can again define them in an abstract way with concrete examples in WebIDL and TypeScript.

> close to going outside of our charter

My preference at the Amsterdam F2F was also to not include DID Resolution in DID Core, but since then we've operated under the assumption that we want to define the "contract" (without its implementation) of DID Resolution (and DID URL Dereferencing).

msporny:
@peacekeeper wrote:
> My preference at the Amsterdam F2F was also to not include DID Resolution in DID Core, but since then we've operated under the assumption that we want to define the "contract" (without its implementation) of DID Resolution (and DID URL Dereferencing).

Then I think it's time to revisit that decision given new information, as that would be my preference as well. Perhaps we do this if it becomes clear that volunteers for writing the tests and two independent implementations are not forthcoming by the end of the first Candidate Recommendation phase.

index.html
msporny: Yes, I had considered that and decided not to go that route because it makes detection of other representation-specific entries optional, which would ultimately lead to interoperability issues on a critical feature needed to solve the problem we're trying to solve. Making this behavior mandatory removes the possibility of non-interoperability at the representation-specific entries data model.

msporny:
Substantive, multiple reviews, changes requested and made, no blocking objections (but a standing request from @OR13 exists to possibly define abstract functions), merging.

PR #686, opened 2021-02-19
Fix "ordered map" language and ensure testability of all MUST statements.

msporny:
I have completed the second pass to make sure that all machine-testable normative MUST/REQUIRED statements are, in fact, machine testable. There was only one issue that I found, a missing link around a `string` type. I also cleaned up the "ordered map" language -> just "map"... because we don't expect people to assume the maps are ordered so we should stop saying "ordered map" everywhere and just say "map".

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/686.html" title="Last updated on Feb 21, 2021, 9:14 PM UTC (f7f247f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/686/a93c6c0...f7f247f.html" title="Last updated on Feb 21, 2021, 9:14 PM UTC (f7f247f)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1179,7 +1179,7 @@ Considerations
       <tbody>
         <tr>
           <td>
-<a data-cite="INFRA#maps">ordered&nbsp;map</a>
+<a data-cite="INFRA#maps">map</a>
           </td>
           <td>
 A finite ordered sequence of key/value pairs, with no key appearing twice as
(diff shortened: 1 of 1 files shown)


brentzundel (approved):

index.html
```diff
@@ -1179,7 +1179,7 @@ <h1>Data Model</h1>
       <tbody>
         <tr>
           <td>
-<a data-cite="INFRA#maps">ordered&nbsp;map</a>
+<a data-cite="INFRA#maps">map</a>
           </td>
           <td>
 A finite ordered sequence of key/value pairs, with no key appearing twice as
```
iherman, line 1185: The note saying "Sometimes referred to as just "map" should be now removed or, and that is possibly better, replace with something like "Sometimes referred to as 'ordered map' in the INFRA standard"

iherman (changes requested):
I must admit my fingers were itching to do this, but I shied away... :-)

index.html
iherman: (Unfortunately, github did not allow me to point at the right line, it only allows me to comment on the changes...)

rhiaro (approved):

index.html
msporny: Fixed in f7f247fddece1403999305c0b923af2625617f85.

msporny:

TallTed:
After the fact, but probably worth recording: Approved.

PR #700, opened 2021-02-26
Fix to remaining suggested changes + adding the `drawio` file for DID URL dereference diagram

shigeya:
- Reflect remaining requested changes in previous PR #648
- Adding the `.drawio` file


shigeya:
@rhiaro @TallTed How about this?
did_url_dereference_overview

TallTed:
Better.

I think moving "relative" from ending the first line, to starting the second line, would be better still, but not vital.

msporny:
@shigeya -- I need to create a static copy of the specification for Candidate Recommendation. That means that all PRs need to be processed. This change is editorial. I'll most likely merge it in the next 48 hours. Please make sure it's ready to go into the spec in its current form.

iherman (approved):
The diagram does not replace the specification text... The spec, mainly around the usage of DID URLs, is really fairly complex, and that cannot be properly put into a diagram. The diagram plays its role, namely to provide a visual aid for the comprehension of the text, very well. I would opt for merging it.


[“The perfect is the enemy of good”](https://en.wikipedia.org/wiki/Perfect_is_the_enemy_of_good) :-)

shigeya:
@TallTed Thanks. How about this version?
did_url_dereference_overview

@msporny The above change is in the PR. Also patched ECHIDNA file. 
I think it's now complete if there is no objection.  If somebody wants me to edit in the next 48hr window, I can update PR without hustle.

TallTed:
Much better now than before.  There's always more polishing that can be done, but I think this will suffice for this doc.

msporny (approved):

Issue #684, opened 2021-02-19
This may be a minor bug: media type file extensions


iherman:
At present all file extensions for the media types in the [IANA considerations section](https://w3c.github.io/did-core/#iana-considerations) are identical: `.did`. I would think this is a copy/paste bug...

msporny:
PR #701 has been raised to fix this bug. This issue will be closed once that PR is merged.

msporny:
PR #701 has been merged, closing.

PR #701, opened 2021-02-26
Fix collisions in IANA Media Type file extensions.

msporny:
As noted in #684, our IANA Mediat Type filename extensions collide. This has been fixed using the only option available to us (that follows the patterns in other Media Type extensions). 

I considered the following patterns:

1. Keep all of them as `.did` -- this fails because associated file media type information isn't always preserved on copy and a system falling back to the file extension wouldn't be able to determine the file format based on OS MAGIC numbers (byte headers -- JSON and JSON-LD headers are the same).
2. Use `.did.json`, `.did.jsonld`, and `.did.cbor` -- looks nice, but there isn't a single example of this pattern in `/etc/mime.types` and it broke 3 of the 5 media type libraries I tested.
3. Use `.djs`, `.dld`, `.dcb` -- compatible with every known OS and filesystem going back to the 1990s... but too cryptic, people associate `js` with JavaScript, not JSON, `dld` isn't accurate, as there will be CBOR-LD, etc.
4. Use `.didjson`, `.didjsonld`, and `.didcbor` -- this seemed to be compatible with every modern operating system and file system. It won't work on filesystems that are more than 20 years old.

This PR goes with item 4 above. Marking it as non-substantive since we don't have any normative statements regarding file extensions in the specification.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/701.html" title="Last updated on Feb 26, 2021, 2:52 PM UTC (2d41799)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/701/8397cf4...2d41799.html" title="Last updated on Feb 26, 2021, 2:52 PM UTC (2d41799)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -5747,7 +5747,7 @@ websites that issue or verify W3C Verifiable Credentials.
             <dt>Magic number(s):</dt>
             <dd>Not Applicable</dd>
             <dt>File extension(s):</dt>
-            <dd>.did</dd>
+            <dd>.didjson</dd>
             <dt>Macintosh file type code(s):</dt>
             <dd>TEXT</dd>
           </dl>
@@ -5824,7 +5824,7 @@ websites that issue or verify W3C Verifiable Credentials.
             <dt>Magic number(s):</dt>
             <dd>Not Applicable</dd>
             <dt>File extension(s):</dt>
-            <dd>.did</dd>
+            <dd>.didjsonld</dd>
             <dt>Macintosh file type code(s):</dt>
             <dd>TEXT</dd>
           </dl>
@@ -5888,7 +5888,7 @@ websites that issue or verify W3C Verifiable Credentials.
             <dt>Magic number(s):</dt>
             <dd>Not Applicable</dd>
             <dt>File extension(s):</dt>
-            <dd>.did</dd>
+            <dd>.didcbor</dd>
             <dt>Macintosh file type code(s):</dt>
             <dd>TEXT</dd>
           </dl>


iherman (approved):

brentzundel (approved):

kdenhartog (approved):

davidlehn:
Another option is to drop the "did" part and just use existing extensions for the high level serialization encoding: `.json`, `.jsonld`, `.cbor`.  There are trade offs of course.

- Pros:
  - Editors and servers are already configured to handle these extensions.
  - Similar to many other use cases across every aspect of computing.  Consider all the `.json`, `.yaml`, `.xml`, etc files everyone deals with and it works.
  - It's easy to support other serializations like `.xml`, `.ttl`, or `.yaml` without any complex registration of `.didXXX`.

- Cons:
  - Editors and servers may need configuration or hints to know how to special case DID behavior such as serving the full DID media type.
  - You don't know a file contains a DID just from the extension.  But I suspect this is not an issue in practice.

Are there use cases that require special `.didXXX` extensions?

msporny:
> Are there use cases that require special .didXXX extensions?

"Yes, double-click to open using pre-registered media type handler in operating system." -- requires 1-to-1 mapping in most operating systems.

Note: systems can still choose to store as .cbor / .json / .yaml -- and if those systems store the media type beside the file, they'll continue to use the right application to open the file, even w/ a wonky file suffix. We may not want to suggest that, though.

msporny:
Non-substantive, multiple reviews, changes suggested (but not made, with explanation given), no objections, merging.

PR #709, opened 2021-03-08
Handling a respec/specref error

iherman:
There is a mysterious bug in respec and/or specref, which means that the JSON-LD11 and VC Data model references are wrong: they refer to a WD and a PR, respectively. Mainly the first one is an issue for a CR publication.

Have tried to track down the bug to yield a proper error report on specref or respec, but it does not seem to be easy (on a simple example it works correctly). As a temporary measure, to make the publication of the CR possible, I have modified, in this branch, the local biblio structure on common.js by adding both of those explicitly with the right parameters. We should use this for the CR publication.

Sigh...

### DIFF
--- a/common.js
+++ b/common.js
@@ -177,6 +177,26 @@ var ccg = {
       ],
       status: "Internet-Draft",
       publisher: "IETF"
+    },
+    "JSON-LD11": {
+      title: "JSON-LD 1.1",
+      date: "2020-07-16",
+      authors: [
+        "Gregg Kellogg", "Pierre-Antoine Champin", "Dave Longley"
+      ],
+      status: "W3C Recommendation",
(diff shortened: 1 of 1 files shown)


rhiaro (approved):

msporny (approved):

msporny:
Editorial, multiple positive reviews, merging immediately as it affects CR publication.

iherman:
@msporny or @rhiaro will one of you regenerate the CR snapshot accordingly?

msporny:
> @msporny or @rhiaro will one of you regenerate the CR snapshot accordingly?

Done https://github.com/w3c/did-core/commit/10487e087a70d11a78b3eb8a7ef59ffc24c3aafd

PR #713, opened 2021-03-17
Updated the multibase doc. reference

iherman:
The new one is https://datatracker.ietf.org/doc/html/draft-multiformats-multibase-03, and the date is February 2021

### DIFF
--- a/common.js
+++ b/common.js
@@ -169,8 +169,8 @@ var ccg = {
     },
     "MULTIBASE": {
       title: "The Multibase Encoding Scheme",
-      date: "August 2020",
-      href: "https://tools.ietf.org/html/draft-multiformats-multibase",
+      date: "February 2021",
+      href: "https://datatracker.ietf.org/doc/html/draft-multiformats-multibase-03",
       authors: [
         "Juan Benet",
         "Manu Sporny"


rhiaro (approved):

msporny (approved):

brentzundel (approved):

kdenhartog (approved):

davidlehn:
Is there general guideline about what to link to?  "tools" vs "datatracker" URLs and the main page vs a specific version?  It's inconsistent between various IETF refs.

iherman:
> Is there general guideline about what to link to? "tools" vs "datatracker" URLs and the main page vs a specific version? It's inconsistent between various IETF refs.

Whatever works :-) 

The problem was that, when publishing the CR, the W3C link checker (and indeed my own checks) returned a 404 for the https://tools.ietf.org/html/draft-multiformats-multibase link. I had to change it to get the publication done. The same link works now (go figure!).

I do not know the IETF publishing world well enough. When I say "whatever works" I actually think "whatever is stable". If the 'tools" reference is usually stable and this was a temporary glitch, then I am fine closing this PR without merge. 

Does anybody know the situation better?

jricher:
@iherman there's a delay while Tools generates the HTMLized version that shows up at that link, not sure if it's a cache thing or just a slow system but it always takes a little bit before it's oneline. The Datatracker rendering is separate and generally much faster. Datatracker is a much newer system than Tools. I think both are considered stable URLs but I believe the Tools URL is considered "more" canonical. This might be just for historical reasons, though.

OR13 (approved):

iherman:
> I think both are considered stable URLs but I believe the Tools URL is considered "more" canonical. This might be just for historical reasons, though.

Well, the only thing I know is that the usage of the 'tool' URI created a last minute hiccup in publishing the CR. I am perfectly fine leaving things as is for now (if the editors so decide), incurring the danger of repeating the problem again...

peacekeeper (approved):

Issue #717, opened 2021-03-26
equivalentId and canonicalId should be optional


peacekeeper:
The spec currently says:

> The value of `equivalentId` _MUST_ be a set where each item is a string ...

> The value of `canonicalId` _MUST_ be a string that conforms to ...

I think this means that those metadata properties are currently required. I am pretty sure the intention of the Working Group was to make them optional and say something like the following instead:

> DID document metadata _MAY_ include a `equivalentId` property. If present, its value _MUST_ be a set where each item is a string ...

> DID document metadata _MAY_ include a `canonicalId` property. If present, its value _MUST_ be a string that conforms to ...

csuwildcat:
@peacekeeper I think the wording is a bit confusing, but it is just saying that the values MUST be those type values IF the Method sets these properties. I do think your language is clearer on this point.

TallTed:
Yeah, I think we had this wording a while back (for certain, on some properties, in some spec I was working on recently).  I would just change the two `The value of ...` to `If present, the value of ...` or maybe `If present, the value of optional property ...`

peacekeeper:
My problem here is that all other metadata properties say very clearly if they SHOULD or MAY be included, whereas equivalentId and canonicalId only talk about their values, but are silent on whether they MUST/SHOULD/MAY be included in the metadata at all. So I think this needs to be clarified.

TallTed:
@peacekeeper - Can you please add a link to the specific problem points in the doc, either in the source or the preview?  I think we're in agreement, that these two should be presented as optional in the same way as their neighbors are presented.  I (and, I believe, others) would prefer that the language used for these two match that already used for the others, whatever that has been, whether it's the complex phrasing you described above or the simpler phrasing I described...

peacekeeper:
@TallTed for example, the section on `nextUpdate` says ([source here](https://github.com/w3c/did-core/blob/main/index.html#L3647) and [spec text here](https://www.w3.org/TR/2021/CR-did-core-20210318/#dfn-nextupdate)):

> DID document metadata MAY include a nextUpdate property if the resolved document version is not the latest version of the document.

On the other hand, the section on `equivalentId` says ([source here](https://github.com/w3c/did-core/blob/main/index.html#L3685) and [spec text here](https://www.w3.org/TR/2021/CR-did-core-20210318/#dfn-equivalentid)):

> The value of equivalentId MUST be a set where each item is a string that conforms to the rules in Section § 3.1 DID Syntax.

[...]

PR #720, opened 2021-03-30
Clarify that equivalentId and canonicalId are optional.

peacekeeper:
Addresses https://github.com/w3c/did-core/issues/717.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/720.html" title="Last updated on Mar 30, 2021, 11:44 AM UTC (e4a7df7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/720/10487e0...e4a7df7.html" title="Last updated on Mar 30, 2021, 11:44 AM UTC (e4a7df7)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -3699,7 +3699,8 @@ express one or more <a>DIDs</a> that are logically equivalent to the resolved
 <code><a>equivalentId</a></code>  property.
             </p>
             <p>
-The value of <code>equivalentId</code> MUST be a <a
+<a>DID document</a> metadata MAY include an <code>equivalentId</code> property.
+If present, the value MUST be a <a
 data-cite="INFRA#ordered-set">set</a> where each item is a
 <a data-cite="INFRA#string">string</a> that conforms to the rules in Section <a
 href="#did-syntax"></a>. The relationship is a statement that each
@@ -3753,7 +3754,8 @@ the canonical ID for the <a>DID subject</a> within the scope of the containing
 <a>DID document</a>.
             </p>
             <p>
-The value of <code><a>canonicalId</a></code> MUST be a <a
+<a>DID document</a> metadata MAY include a <code>canonicalId</code> property.
+If present, the value MUST be a <a
 data-cite="INFRA#string">string</a> that conforms to the rules in Section <a
 href="#did-syntax"></a>. The relationship is a statement that the
 <code><a>canonicalId</a></code> value is logically equivalent to the


TallTed (approved):
LGTM

csuwildcat (approved):

OR13 (approved):

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2019/did-wg/Meetings/Minutes/2021-04-06-did#section8-2) on 2021-04-06

- no resolutions were taken

#### 8.2. equivalentId and canonicalId should be optional 

_See github issue [#717](https://github.com/w3c/did-core/issues/717)._




_See github pull request [#720](https://github.com/w3c/did-core/pull/720)._




> *Daniel Buchner:* Let's fight!

**Daniel Burnett:** All involved people of this issue seem to be on the call, maybe we can move this forward.  

**Markus Sabadello:** I raised this because my reading of the spec text regarding the equivalent id and canonical id was that it felt like those are mandatory  
… or the spec did not say that they are optional  
… which is different from how other did doc metadata are described  
… I raised a simple PR that adds language to say that they are optional  
… I think that had support. Daniel then explained what this meant  
… and what did methods have to do  
… I'm not sure if the PR is sufficient or if there's anything else that needs to bed one  

**Daniel Buchner:** the interesting thing here is properties are required for methods that implement them. it's a directive that says if you're going to implement then you can't sometimes not throw them in  
… I don't know if we reflect that in the spec test.. optional in the general sense of did docs  
… if you choose to use these features it becomes required  
… not sure if we should put that in  

**Ted Thibodeau Jr.:** I'm fine with the text as its' in 720  
… it sounds like there's an additional piece  
… someplace else we said methods can be more stringent or can require something or can define things.. maybe there's another few words to throw in  
… I don't think this would count as a normative change to push another CR  
… the ambiguous test that suggested it was required would have meant that they forced it and the change to make it optional doesn't break anything  

**Daniel Buchner:** I can say that any method implementing these, the 4 or 5 today, would not be broken by changing it to optional in the general document sense  

**Ivan Herman:** daniel answered for those he knows about. Any change like that does it affect existing implementations?  
… if the answer is no then we are in a good place  

**Daniel Buchner:** the answer is no  
… if they didn't, if they were broken by this change it means they were misreading the rest of the properties  

**Daniel Burnett:** is there anything not said by the existing text after 720 is applied? take a look at 720 and if it's not complete make a suggestion or a new PR  
… thanks everyone  

> *Daniel Buchner:* will do

> *Daniel Buchner:* gtg, thank you for all your help (Justin, Dan, Markus, Ted, etc.)

---

</details>

kdenhartog (approved):

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2019/did-wg/Meetings/Minutes/2021-04-27-did#section3-1) on 2021-04-27

- no resolutions were taken

#### 3.1. Clarify that equivalentId and canonicalId are optional. 

_See github pull request [#720](https://github.com/w3c/did-core/pull/720)._




**Daniel Burnett:** Essentially editorial but need status update... ^  

**Manu Sporny:** Heads up: I went through all PRs and all issues and categories everything as editorial or substantive.  
… It would be good to pull these PRs in. I would appreciate multiple reviews on the ones that are editorial  
… 720 we have discussed on previous call as nonsubstantive.  
… Markus found a strangeness in the spec where we didn't explicitly say that equivalentId and canonicalId are optional.  
… What the PR does is explicitly state that. We don't think any implementations would have to change as a result.  
… If anyone disagrees, they should express that.  
… It changes a normative statement but in a way the group meant and we don't think implementations would have to change based on the update.  

**Daniel Burnett:** Can you give a target when to expect it to be merged?  

**Manu Sporny:** Ideally anything older than 7 days would be merged.  
… Because editorial, and not substantive, suggestion is to merge anything.  
… We haven't merged anything since entering CR. First time merging.  
… Does anyone have any issues with merging things that have been out there, editorial/nonsubstantive?  
… If not, will merge after this call or latest by Sunday.  

> *Ted Thibodeau Jr.:* re #720 -- No implementation that included either/both would have to change to drop it/them. No implementation that didn't include either/both would have to change to add it/them. Inclusion and omission are all OK. So, no implementation impact.

**Daniel Burnett:** I'm not hearing any objections. People will have opportunity to read the minutes.  
… You are correct on our ability to move forward based on implementations and agreement on clarification.  
… Falls into uncommon category, "clarification".  
… No disagreement or dissent about it.  
… Giving much more time than usual, to make absolutely sure.  

**Ted Thibodeau Jr.:** throwing a flag on 720  
… Just removed the issues that are deferred  
… Our interpretation works fine for generators, but it didn't matter whether they included or didn't include those properties. But consumers who read what we wrote will not be flummoxed by leaving them out.  

**Manu Sporny:** A previous implementor reading that ... It's not clear whether it always had to be there or not.  

**Ted Thibodeau Jr.:** Correct  

**Ted Thibodeau Jr.:** The previously ambiguous language could have led someone to believe that was required...  

**Manu Sporny:** ...  

**Ted Thibodeau Jr.:** It's a ... justification. Just want to make sure we are all on the same page ...  

**Daniel Buchner:** note: before that text it says obviously not putting these in if not using these features.  

**Ted Thibodeau Jr.:** Not saying it's likely... but the sort of thing that has bitten WGs in the past  

**Daniel Buchner:** Makes sense  

**Manu Sporny:** You're okay with that justification?  

**Ted Thibodeau Jr.:** Yes  

**Manu Sporny:** (earlier missed comment) A consumer software that was implementing this normative statement previously would have read that the value of equivalentId must be a string. It wasn't clear at all whether that was optional or mandatory, so making a decision one way or the other wouldn't have been supported by the text. What we are doing now is going back and clarifying that we always meant it to  

**Charles Lehner:** be optional. That is the justification for allowing it to be a nonsubstantive change to the spec.

> *Daniel Buchner:* @orie: about the PR on JSON production: I wasn't sure what exactly to do on the value tests - am I supposed to invoke JS typeof or something to prove it's a number, for example?

</details>

dlongley (approved):

msporny:
Non-substantive clarification, multiple positive reviews, no objections, merging.

Issue #707, opened 2021-03-03
CR Feedback: Support of `publicKeyMultibase`


carsonfarmer:
According to https://w3c.github.io/did-core/#verification-material, the "DID Working Group is seeking implementer feedback on the preference of the ecosystem with respect to using `publicKeyBase58` or `publicKeyMultibase`. I would like to *strongly signal preference for inclusion of `publicKeyMultibase` in the spec*.

As stated in the did-core document, `publicKeyMultibase` supports more base-representation formats and provides a more future proof path. It additionally supports publicKeyBase58 quite easily, and there are now lots of libraries and implementations of the "standard" out in the wild, most of which are fully compatible and stable. See for example: https://github.com/multiformats/multibase#implementations (linking for those less familiar with multiformats and multibase encoding). In the interest of limiting the number of formats for expressing public key material in a DID Document to the fewest possible, and to increase the likelihood of interoperability, I would suggest that `publicKeyMultibase` be the preferred encoding format.

Our team (@textileio) has already begun implementing a Go-lang version of https://github.com/decentralized-identity/did-resolver where we are explicitly supporting `publicKeyMultibase` for compatibility with the wider multibase/IPLD/IPFS ecosystem. I have also cc'ed a few other contributors/implementors that might be interested in voicing their opinion here as well.

Thank you, and apologies if this is not the right venue to signal support!

cc @sanderpick @oed

OR13:
@msporny and I have also worked on publicKeyMultibase.... https://w3c-ccg.github.io/lds-ed25519-2020/

However, we fought for a very long time about this, and settled on `publicKeyJwk` and `publicKeyBase58`.... I would be happy replacing `publicKeyBase58` with `publicKeyMultibase`... but adding yet another verification method material type to the core seems like a bad idea...

IMO the best thing we could do would be replace `publicKeyBase58` with `publicKeyMultibase` in DID Core and require an extension for `publicKeyBase58` in did spec registries.

carsonfarmer:
100% in agreement on this!

msporny:
> Thank you, and apologies if this is not the right venue to signal support!

With my Editor hat off -- Digital Bazaar is also strongly in favor of replacing `publicKeyBase58` with `publicKeyMultibase` in the specification. We might argue to still include `publicKeyBase58` in the DID Core context (but not in DID Core) as a number of implementations already use that out in the wild. I also agree with @OR13 that it would be bad to add yet another verification method material type to the core.

[...]

PR #731, opened 2021-05-08
Replace use of publicKeyBase58 with publicKeyMultibase.

msporny:
This PR addresses issue #707 by replacing the use of publicKeyBase58 with publicKeyMultibase.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/731.html" title="Last updated on May 15, 2021, 2:43 PM UTC (0be2206)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/731/79893da...0be2206.html" title="Last updated on May 15, 2021, 2:43 PM UTC (0be2206)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1842,7 +1842,7 @@ processes. Examples of verification material properties are
 material. For example, see
 <a href="https://w3c-ccg.github.io/lds-jws2020/">JSON
 Web Signature 2020</a> and <a
-href="https://w3c-ccg.github.io/lds-ed25519-2018/">Ed25519 Signature 2018</a>.
+href="https://w3c-ccg.github.io/lds-ed25519-2020/">Ed25519 Signature 2020</a>.
 For all registered <a>verification method</a> types and associated verification
 material available for <a>DIDs</a>, please see the DID Specification Registries
 [[?DID-SPEC-REGISTRIES]].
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1916,9 +1906,9 @@ <h3>Verification Material</h3>
     }
   }, {
     "id": "did:example:123456789abcdefghi#keys-1",
-    "type": "Ed25519VerificationKey2018", <span class="comment">// external (property value)</span>
+    "type": "Ed25519VerificationKey2020", <span class="comment">// external (property value)</span>
```
peacekeeper, line 1909: What does "external" mean here? I assume it means that the property value `Ed25519VerificationKey2020` is defined outside of DID Core, e.g. in the CCG's [Security Vocabulary](https://github.com/w3c-ccg/security-vocab), or somewhere in the new LDS WG, or in a cryptographic suite. This is good!

index.html
```diff
@@ -3177,7 +3167,7 @@ <h3>Production</h3>
       64                             # text(4)
          74797065                    # "type"
       78 1A                          # text(26)
-         45643235353139566572696669636174696F6E4B657932303138 # "Ed25519VerificationKey2018"
+         45643235353139566572696669636174696F6E4B657932303138 # "Ed25519VerificationKey2020"
```
peacekeeper, line 3170: The CBOR bytes also needs to be updated, but I guess that can happen in a separate PR (in the new CBOR Note repo).

peacekeeper:
@msporny I think it would still be good to get an explanation of your comment in https://github.com/w3c/did-core/issues/707#issuecomment-813760646:

> Just because we're supporting publicKeyMultibase doesn't mean we want to support ALL base-encodings (that would be a bit insane)

Which base-encodings do we support for publicKeyMultibase? Can each implementation choose which ones they support (in this case, see concerns in https://github.com/w3c/did-core/issues/707#issuecomment-811073119 about different levels of implementation support).

Or is it up to the cryptographic suite (and the verification method type) to define which base-encodings to support? That latter approach might be best, perhaps we can add a sentence to that effect just to clarify.

shigeya (approved):
I haven't check whether actual encoding values are correct byte-by-byte (besides adding Mutilbase prefix `z`).

Besides that, it looks good to me.

index.html
msporny: Yes, your assumption is correct. Someone (can't remember now) asked for all "external properties" to be clearly marked in the specification and @rhiaro went through and did that a while ago.

msporny:
Substantive (but clearly marked as at risk and subject to change), multiple positive reviews, no objections, merging.

Issue #728, opened 2021-04-30
All appendix sections needs to be cleaned up.


msporny:
The appendices in the document need one last Editorial clean up pass before we exit CR. This issue is just here to track that task.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2019/did-wg/Meetings/Minutes/2021-06-08-did#section5-2) on 2021-06-08

- no resolutions were taken

<details><summary><i>View the transcript</i></summary>

#### 5.2. All appendix sections needs to be cleaned up 

_See github issue [#728](https://github.com/w3c/did-core/issues/728)._




**Brent Zundel:** next 728  

**Manu Sporny:** there are 4 remaining PRs, once those are merged all of the appendices will have had all the editorial changes  
… means we're done with editorial work  
… modulo small adjustments  

**Brent Zundel:** good news  
… next is 729  

</details>

msporny:
This is finally done, all PRs to clean up the appendixes have been merged. Closing.

Issue #583, opened 2021-01-28
Proving Control sections are wrong


rhiaro:
The [proving control of DID docs](https://w3c.github.io/did-core/#proving-control-of-a-did-and-did-document) and [of public keys](https://w3c.github.io/did-core/#proving-control-of-a-public-key) subsections in Security Considerations are inaccurate.

We need to fix these ASAP, but this can probably be done during CR as they are not normative.

msporny:
Yes, correct. These sections were written long ago and are now misleading at best and catastrophically wrong at worst.

peacekeeper:
I agree the sections could be improved. They talk too much about signed DID documents.

But there is also some useful content in there that shouldn't be removed. E.g. to state that 1. a signed DID document doesn't prove control of a DID, and 2. proving control of a DID requires DID resolution, those are two important points to mention.

ktobich:
Paragraph 9.2.2 Proving Control of a Public Key might need to reviewed by getting inspirited by the Static Data Authentication (SDA) and the Dynamic Data Authentication (DDA) used in the payment industry and highlight maybe the differences with static and dynamic method used by the DID and described in the section. The ICAO (passport) scheme might be a good one too.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2019/did-wg/Meetings/Minutes/2021-03-30-did#section5-5) on 2021-03-30

- no resolutions were taken

<details><summary><i>View the transcript</i></summary>

#### 5.5. Proving Control sections are wrong 

_See github issue [#583](https://github.com/w3c/did-core/issues/583)._




**Brent Zundel:** Assigned to Amy, but I don't see her. manu or markus_sabadello?  

**Manu Sporny:** Amy is going to do it  

</details>

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2019/did-wg/Meetings/Minutes/2021-05-04-did#section7-1) on 2021-05-04

- no resolutions were taken

<details><summary><i>View the transcript</i></summary>

#### 7.1. Proving Control sections are wrong 

_See github issue [#583](https://github.com/w3c/did-core/issues/583)._




**Brent Zundel:** There has been a bit of conversation. Need a PR to fix issue.  

**Manu Sporny:** General statement: I triaged issue this past weekend, marked everything that was a CR comment. Marked anything ready for a PR as ready for PR.  
… Almost every one except 2 or 3 are ready for PR.  

[...]

PR #738, opened 2021-05-16
Update Proving Control and Binding section.

msporny:
Partial editorial cleanup to appendices tracked as issue #728. This also addresses issue #583.

This section was a bit of a mess and required a fairly significant rewrite... I tried to save what I could, much of the language was quite old and represented thinking from 3+ years ago. I tried to update the language to be less prescriptive, point to sections that have been created to address each item, and remove the concept of "proving control over a key" -- 'cause that's not really a use case, it's just a smaller part of larger use cases such as authentication and authorization, which we already have sections that go to lengths to talk about those mechanisms elsewhere in the spec (so I pointed there).

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/738.html" title="Last updated on May 22, 2021, 2:58 PM UTC (fe809af)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/738/ab18a18...fe809af.html" title="Last updated on May 22, 2021, 2:58 PM UTC (fe809af)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -4441,11 +4441,11 @@ such bindings has to be carefully balanced against privacy considerations (see
         </p>
 
         <p>
-The process of binding a <a>DID</a> to something in the real world, such as a
-person or an organization, for example using <a>verifiable credentials</a> with
-the same subject as that <a>DID</a>, is contemplated by this specification
-and further defined in the Verifiable Credentials
-Data Model [[VC-DATA-MODEL]].
+The process of binding a <a>DID</a> to something in the physical world, such as
+a person or an organization &mdash; for example, by using <a>verifiable
+credentials</a> with the same subject as that <a>DID</a> &mdash; is contemplated
+by this specification and further defined in the Verifiable Credentials Data
+Model [[VC-DATA-MODEL]].
         </p>
       </section>
     </section>
--- a/index.html
+++ b/index.html
@@ -4391,127 +4391,61 @@ use.
     </section>
 
     <section>
-      <h2>Binding of Identity</h2>
+      <h2>Proving Control and Binding</h2>
 
       <p>
-The following sections describe binding identities to <a>DIDs</a> and
-<a>DID documents</a>.
+Binding an entity in the digital world or the physical world to a <a>DID</a>,
+<a>DID document</a>, or cryptographic material requires the use of security
+protocols contemplated by this specification. The following sections describe
+some of these scenarios and how an entity might prove control over a
+<a>DID</a> or a <a>DID document</a> for the purposes of authentication or
+authorization.
       </p>
 
-      <section>
+      <section class="notoc">
         <h3>Proving Control of a DID and DID Document</h3>
 
-        <p class="issue" data-number="583">
-This section is inaccurate and needs revision.
-        </p>
-
         <p>
-Signatures and <a>verifiable timestamps</a> allow <a>DID documents</a> to be
-cryptographically verifiable.
-        </p>
-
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -4391,127 +4391,61 @@ <h2>Choosing DID Resolvers</h2>
...
-It can be useful to express a binding of <a>DID</a> to a person's or company's
-real world identity, in a way that is provably asserted by a trusted authority
-such as a government. This can enable interactions that can be considered
-legally valid under one or more jurisdictions; establishing such bindings has to
-be carefully balanced against privacy considerations (see <a
-href="#privacy-considerations"></a>).
+It can be useful to express a binding of a <a>DID</a> to a person's or
+organization's physical identity in a way that is provably asserted by a
+trusted authority, such as a government. This specification provides a
+<a>verification relationship</a> in <a href="#assertion"></a> for these
```
rhiaro, line 4436:
```suggestion
the <a href="#assertion"></a> <a>verification relationship</a> for these
```

rhiaro (approved):

brentzundel (approved):

peacekeeper (approved):
I think this is good and makes it clearer what "control over a DID" means.

I do think it would still be valuable to keep some of the guidance about signed DID documents. We should say that some DID methods do that, and it can make sense depending on the DID method design, but signed DID documents alone are no guarantee that the DID document is the correct one for the DID.

We could merge this PR as-is, and then I could create another PR to add this back in (either as part of the same section, or a new section).

index.html
```diff
@@ -4391,127 +4391,61 @@ <h2>Choosing DID Resolvers</h2>
...
 
         <p>
 The process of binding a <a>DID</a> to something in the real world, such as a
-person or a company, for example using <a>verifiable credentials</a> with the
-same subject as that <a>DID</a>, is out of scope for this specification. For
-more information, see the Verifiable Credentials Data Model [[VC-DATA-MODEL]].
+person or an organization, for example using <a>verifiable credentials</a> with
+the same subject as that <a>DID</a>, is contemplated by this specification
+and further defined in the Verifiable Credentials
+Data Model [[VC-DATA-MODEL]].
```
TallTed, line 4448:
```suggestion
The process of binding a <a>DID</a> to something in the physical world, such
as a person or an organization &mdash; for example, by using <a>verifiable
credentials</a> with the same subject as that <a>DID</a> &mdash; is
contemplated by this specification and further defined in the Verifiable
Credentials Data Model [[VC-DATA-MODEL]].
```

index.html
TallTed, line 4436:
```suggestion
trusted authority, such as a government. This specification provides the
<a href="#assertion"></a> <a>verification relationship</a> for these
```

TallTed (changes requested):
Focused more on the new text as itself, and less on whether the
new text carries everything it should from the old.

index.html
msporny: Applied in @rhiaro's suggestion.

index.html
msporny: Applied in fe809af3db8f85296af1bcc49648efa470c572a4.

msporny:

TallTed:
[Change applied](https://github.com/w3c/did-core/pull/738#discussion_r637413448) as `provides a the <a` still needs to lose the redundant and incorrect `a` article, which I could not suggest as a modification of the earlier suggestion from @rhiaro.

PR #569, opened 2021-01-23
Add rotation and recovery sections from previous PR

OR13:
Previous PR was corrupted with merge conflicts, this was an easier way for me to get a clean change-set...

All feedback as existing on the previous PR has been implemented (but not accepted)
https://github.com/w3c/did-core/pull/548

Specifically @dhh1128 please request any changes on this PR that I failed to address.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/569.html" title="Last updated on Jan 31, 2021, 2:41 PM UTC (cff7eed)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/569/f2bfa86...cff7eed.html" title="Last updated on Jan 31, 2021, 2:41 PM UTC (cff7eed)">Diff</a>

PR #741, opened 2021-05-16
Update Verification Method Rotation section.

msporny:
Partial editorial cleanup to appendices tracked as issue #728.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/741.html" title="Last updated on May 29, 2021, 7:17 PM UTC (7bce470)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/741/ab18a18...7bce470.html" title="Last updated on May 29, 2021, 7:17 PM UTC (7bce470)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -4627,7 +4627,7 @@ The following considerations might be of use when contemplating the use of
 <a>Verification method</a> rotation is a proactive security measure.
         </li>
         <li>
-It is considered a best practice to perform <a>verification method</a> rotation
+It is generally considered a best practice to perform <a>verification method</a> rotation
 on a regular basis.
         </li>
         <li>
@@ -4635,7 +4635,7 @@ Higher security environments tend to employ more frequent verification method
 rotation.
         </li>
         <li>
-<a>Verification method</a> rotation applies only to the current or latest
+<a>Verification method</a> rotation manifests only as changes to the current or latest
 version of a <a>DID document</a>.
         </li>
         <li>
@@ -4653,7 +4653,7 @@ rotation. In these cases, verification software might require additional
 information, such as when a particular <a>verification method</a> was
 expected to be valid as well as access to a <a>verifiable data registry</a>
 containing a historical record, to determine the validity of the proof or
-signature.
+signature. This option might not be available in all <a>DID methods</a>.
         </li>
         <li>
 The section on <a href="#method-operations">DID method operations</a> specifies
--- a/index.html
+++ b/index.html
@@ -4596,78 +4596,80 @@ of a resolver ought to be compatible with such extension behavior.
     </section>
 
     <section>
-
       <h2>Verification Method Rotation</h2>
 
-      <p>Verification method rotation is a proactive security measure.</p>
-
       <p>
-        Verification method rotation applies only to the current or latest
-        version of a DID Document.
-      </p>
-
-      <p>
-        When a verification method has been active for a long time, or used
-        for many operations, a controller might wish to perform a rotation.
-      </p>
-
-      <p class="advisement">
-        It is considered a best practice to perform verification method
-        rotation on a regular basis.
-      </p>
-
-
-      <p class="advisement">
-        Proofs or signatures that rely on verification methods that are not
-        present in the latest version of a DID Document are not impacted by
-        rotation, and might require additional information to mitigate compromise.
+Rotation is a management process that enables the secret cryptographic material
+associated with an existing <a>verification method</a> to be deactivated or
+destroyed once a new <a>verification method</a> has been added to the  <a>DID
+document</a>. Going forward, any new proofs that a <a>controller</a> would have
+generated using the old secret cryptographic material can now instead be
+generated using the new  cryptographic material and can be verified using the
+new <a>verification method</a>.
       </p>
 
       <p>
-        Section <a href="#method-operations"></a> specifies the
-        <a>DID</a> operations to be supported by a
-        <a>DID method</a> specification, including
-        <a href="#method-operations">update</a>
-        which is expected to be used to perform a verification method rotation.
+Rotation is a useful mechanism for protecting against verification method
+compromise, since frequent rotation of a verification method by the controller
+reduces the value of a single compromised verification method to an attacker.
+Performing revocation immediately after rotation is useful for verification
+methods that a controller designates for short-lived verifications, such as
+those involved in encrypting messages and authentication.
       </p>
 
       <p>
-        A controller performs a rotation when they add a new verification
-        method that is meant to replace an existing verification method after
-        some time.
+The following considerations might be of use when contemplating the use of
+<a>verification method</a> rotation:
       </p>
 
-      <p>Not all DID Methods support verification method rotation.</p>
-
-      <p>
-        Rotation is a key management process that enables the private cryptographic
-        material associated with an existing verification method to be deactivated
-        or destroyed once a new verification method has been added to the DID
-        Document. Going forward, any new proofs that a controller would have generated
-        using the old cryptographic material can now instead be generated using the new
-        material and can be verified using the new verification method.
-      </p>
-
-      <p>
-        Rotation is a useful mechanism for protecting against verification
-        method compromise, since frequent rotation of a verification method by
-        the controller reduces the value of a single compromised verification method
-        to an attacker. Performing revocation immediately after rotation is
-        useful for verification methods that a controller designates for
-        short-lived verifications, such as those involved in encrypting messages
-        and authentication.
-      </p>
-
-      <p class="note">
-        Higher security environments tend to employ more frequent verification
-        method rotation.
-      </p>
-
-      <p class="note">
-        Frequent rotation of a verification method might be frustrating for
-        parties that are forced to continuously renew or refresh associated
-        credentials.
-      </p>
+      <ul>
+        <li>
+<a>Verification method</a> rotation is a proactive security measure.
+        </li>
+        <li>
+It is considered a best practice to perform <a>verification method</a> rotation
+on a regular basis.
+        </li>
+        <li>
+Higher security environments tend to employ more frequent verification method
+rotation.
+        </li>
+        <li>
+<a>Verification method</a> rotation applies only to the current or latest
+version of a <a>DID document</a>.
+        </li>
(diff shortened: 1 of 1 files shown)


rhiaro (approved):

brentzundel (approved):

index.html
```diff
@@ -4596,78 +4596,80 @@ <h2>Key and Signature Expiration</h2>
...
-        Frequent rotation of a verification method might be frustrating for
-        parties that are forced to continuously renew or refresh associated
-        credentials.
-      </p>
+      <ul>
+        <li>
+<a>Verification method</a> rotation is a proactive security measure.
+        </li>
+        <li>
+It is considered a best practice to perform <a>verification method</a> rotation
```
dhh1128, line 4630: Can we insert the hedge word "generally" here: "It is generally considered..."

See https://www.strongsalt.com/encryption-key-rotation-is-useless%E2%80%8A-%E2%80%8Aheres-why/ for an explanation of why not everyone agrees that this is a good idea. (In short: the more rotation, the more chances for human error and key loss.)

index.html
```diff
@@ -4596,78 +4596,80 @@ <h2>Key and Signature Expiration</h2>
...
+on a regular basis.
+        </li>
+        <li>
+Higher security environments tend to employ more frequent verification method
+rotation.
+        </li>
+        <li>
+<a>Verification method</a> rotation applies only to the current or latest
```
dhh1128, line 4638: I think "applies" is vague. Better might be: "Verification method rotation manifests only as changes to the current..." or similar.

index.html
```diff
@@ -4596,78 +4596,80 @@ <h2>Key and Signature Expiration</h2>
...
+        <li>
+When a <a>verification method</a> has been active for a long time, or used for
+many operations, a controller might wish to perform a rotation.
+        </li>
+        <li>
+Frequent rotation of a <a>verification method</a> might be frustrating for
+parties that are forced to continuously renew or refresh associated credentials.
+        </li>
+        <li>
+Proofs or signatures that rely on <a>verification methods</a> that are not
```
dhh1128, line 4650: This feels at odds with verbiage in the revised revocation section, which says that omitting a verification method from a DID doc is the only universally supported method of revoking something. How do we square the two statements?

index.html
```diff
@@ -4596,78 +4596,80 @@ <h2>Key and Signature Expiration</h2>
...
+        <li>
+The section on <a href="#method-operations">DID method operations</a> specifies
+the <a>DID</a> operations to be supported by a <a>DID method</a> specification,
+including <a href="#method-operations">update</a> which is expected to be used
+to perform a <a>verification method</a> rotation.
+        </li>
+        <li>
+A <a>controller</a> performs a rotation when they add a new <a>verification
+method</a> that is meant to replace an existing <a>verification method</a> after
+some time.
```
dhh1128, line 4667: Is rotation meant to communicate an intention to replace exactly at the moment of the rotation -- or at some indefinite point in the future? If the former interpretation is canonical, then isn't revocation of the old method automatic?

TallTed:
I am suddenly very confused.

I have understood _Verification Method_ to be, roughly, "what is done with the crypto material", as opposed to "the crypto material" itself.

Rotating crypto material makes sense, for all the reasons discussed in this section.  But rotating "what you do with the crypto material" doesn't make the same kind of sense --- or at least, needs different justifications.

Can anyone help me figure out how and where I have lost the plot?

index.html
```diff
@@ -4596,78 +4596,80 @@ <h2>Key and Signature Expiration</h2>
...
+parties that are forced to continuously renew or refresh associated credentials.
+        </li>
+        <li>
+Proofs or signatures that rely on <a>verification methods</a> that are not
+present in the latest version of a <a>DID document</a> are not impacted by
+rotation. In these cases, verification software might require additional
+information, such as when a particular <a>verification method</a> was
+expected to be valid as well as access to a <a>verifiable data registry</a>
+containing a historical record, to determine the validity of the proof or
+signature.
```
kdenhartog, line 4656:
```suggestion
signature. In some <a>DID methods</a> this option may not be available.
```

kdenhartog: Specifically trying to call out not all methods will support versioning/historical records even if update/rotation is supported.

index.html
msporny, line 4630:
```suggestion
It is generally considered a best practice to perform <a>verification method</a> rotation
```

index.html
msporny, line 4638:
```suggestion
<a>Verification method</a> rotation manifests only as changes to the current or latest
```

index.html
msporny: Applied @TallTed 's suggestion to @kdenhartog 's comment.

index.html
msporny: I don't understand how it's at odds. It feels aligned. Can you provide some concrete specification text that we could change this to?

Any suggestions @OR13, since it's your language, IIRC?

index.html
msporny: My expectation is that it's "at some indefinite point in the future"... which might be within microseconds of each other. The process, as I understand it might be:

1. Add replacement key (k2).
2. Remove old key (k1).

So, it can't be considered automatic all of the time (but perhaps, sometimes). Yes, the future moment could be thought of as when the old key is revoked.

@dhh1128 please take a shot at some concrete spec text, or @OR13 if you have any suggestions here, that would be great. We need that text before I can merge.

msporny:
> I am suddenly very confused.
> 
> I have understood _Verification Method_ to be, roughly, "what is done with the crypto material", as opposed to "the crypto material" itself.
> Can anyone help me figure out how and where I have lost the plot?

We just so have a handy definition of the term here:

https://w3c.github.io/did-core/#dfn-verification-method

... and your summary isn't aligned with it. :)

Hope that helps.

TallTed:
Yes, @msporny, thank you. I was blurring DID methods with (at least) verification methods, the latter of which could apparently be described as "the crypto materials and what is done with them".

msporny:
@dhh1128 -- there are action items for you in this PR. If we don't hear from you in 48 hours (which will be 7 days from the time we requested feedback from you), we're going to merge this editorial PR. We can always put any changes you want in a future PR.

index.html
dhh1128: Under the definition you're advocating here, the only difference between "inserting" a new key and "rotating" a key is intention; there is no observable difference in DID doc state when the two operations complete. This feels weird to me. I thought "rotation" was the combination of adding a new key and removing an old one, as an atomic operation.

dhh1128: I'm not suggesting spec text here because I think the definition is fundamentally wrong. If this definition is maintained, I think the whole section needs to be rewritten. For example, the earlier text "enables the secret cryptographic material associated with an existing verification method to be deactivated or destroyed once a new verification method; has been added to the DIDdocument" has to be revised, since it implies the opposite order (that rotation happens *after* revocation).

index.html
dhh1128: What I'm trying to point out here is that "verification methods that are not present in the latest version of a DID document" are required, in the revocation section, to be interpreted as revoked methods (if they ever existed at all). Thus, this sentence could be reworded to "Verification methods that are already revoked cannot be rotated."

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2019/did-wg/Meetings/Minutes/2021-06-08-did#section5-6) on 2021-06-08

- no resolutions were taken

#### 5.6. Update Verification Method Rotation section 

_See github pull request [#741](https://github.com/w3c/did-core/pull/741)._




**Manu Sporny:** might need more people to look  
… not sure if that would help  
… biggest issue is that it as editorial update and when it was made people are now re discussing the entirety of the text  
… daniel hardman has concerns  
… orie needs to respond  
… it's their text  
… but now there's disagreement over the text that they had initially put in there by daniel  
… the default is the PR won't go in and the thing will be editorially messy  
… I don't know if anyone can chat with daniel hardman and say we need him to propose concrete text and we need orie to talk with him to get there  
… I don't have any strong preference one way or the other  
… largely editorial stuff  
… misunderstanding around text  
… need orie and daniel to talk  
… that's what will resolve it, floating for 23 days  
… neither of them are here  
… can anyone take an action to talk with daniel and/or orie to get them to resolve?  

**Brent Zundel:** i can reach out to daniel  
… we have completed the agenda in record time  

</details>

index.html
msporny: I don't have input on this PR. I was merely cleaning up text that was already there when I stepped on this landmine. 

I believe other people submitted this text initially and I would like them to work through the details here. I remember it being @dhh1128 and @OR13 working on the initial PR here, but can't remember exactly.

I don't necessarily care about the outcome here as long as it's consistent. I do think @dhh1128 is raising good points, so I expect that anything that makes it in would address those concerns. To avoid a "too many cooks in the kitchen" scenario, I'm recusing myself from providing input.

If it helps, I can try to hunt down who initially worked on and refined this section of the specification.

kdenhartog: Did a quick lookup to save you the time @msporny 

The original text was merged via PR #569 with an alternative originally proposed by @OR13 in #548. This was originally intended to address issue #386 

It looks like @dhh1128 and @OR13 were able to come to agreement after some back and forth on the editing of this. Hopefully the discussion between those two will provide some additional context on what allowed the current text to be satisfying to all parties at that time.

OR13 (approved):

TallTed:
I think I figured out where my earlier confusion really came from:

> Rotation...enables the secret cryptographic material associated with an existing verification method to be deactivated or destroyed once a new verification method has been added to the DIDdocument.

I have't quite figured out how to reword this to eliminate the confusion.

Something about the new verification method being comprised of (not merely "associated with") both new secret cryptographic material and new "what is to be done with it", upon addition of which to the DID document, *both* the old secret cryptographic material and the old "what is to be done with it", which together comprise the old verification method, may (should? must?) be deleted....

dhh1128:
I don't love this text, but I am publicly clarifying that I don't object to merging it.

msporny:
Editorial, multiple reviews, some changes requested and made, others not concrete enough to apply, no objections, merging.

Issue #734, opened 2021-05-15
Examples might be out of date


msporny:
The following section might be out of date (and contains no prose explaining the examples): 

https://w3c.github.io/did-core/#examples 

We need to: 

1. Make sure the examples are correct.
2. Write some prose to explain each example.
3. Consider moving the examples to the implementation guide so that they can be updated more frequently (and then point to the implementation guide for those interested in examples.

If these items above are not done, we should move the entire appendix to the implementation guide.

@OR13, since you wrote this section, can you make a pass on it to make sure it's correct (as it'll probably be at least 6-12 months before we get a chance to fix it again once the REC is out).

OR13:
yes, examples are certainly out of date, we probably need w3id.org/security/suite/.... everywhere.

OR13:
@msporny looks like you got to many if not all of these.

msporny:
> @msporny looks like you got to many if not all of these.

I think the only ones I touched were the multibase ones... didn't touch many of the others. I wouldn't assume I got them, I was purely focusing on publicKeyMultibase when I did my pass (which I'm not sure was correct). At the very least, need a second set of eyes on the examples.

OR13:
cool, I can see a few missing from this example:

https://w3c.github.io/did-core/#example-31-did-document-with-many-different-verification-methods

which IMO is really an anti pattern example...

I will open a PR to clean it up.

OR13:
PR is up for this: https://github.com/w3c/did-core/pull/748

msporny:
PR #748 has been raised to address this issue. This issue will be closed once that PR has been merged.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2019/did-wg/Meetings/Minutes/2021-05-25-did#section3-5) on 2021-05-25

- no resolutions were taken

<details><summary><i>View the transcript</i></summary>

#### 3.5. Examples might be out of date 

_See github issue [#734](https://github.com/w3c/did-core/issues/734)._




_See github pull request [#748](https://github.com/w3c/did-core/pull/748)._




**Drummond Reed:** I submitted PRs on that and the next issue; they just need review  

**Orie Steele:** I removed the Ethereum address example  
… there is still a potential issue with having a public key multibase example first  
… so I hoped that providing several examples would solve the problem  

**Brent Zundel:** a link to the PR is in IRC so please review  

</details>

msporny:
PR #748 has been merged and addressed my concerns.

PR #748, opened 2021-05-22
Update DID Document examples

OR13:
Addresses: #734

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/748.html" title="Last updated on May 25, 2021, 10:47 PM UTC (242ea30)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/748/435a304...242ea30.html" title="Last updated on May 25, 2021, 10:47 PM UTC (242ea30)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -5045,103 +5045,149 @@ practice to avoid using the same <a>verification method</a> for multiple purpose
 }
       </pre>
 
-      <pre class="example" title="DID Document with many different verification methods">
+      <pre class="example" title="DID Document with many different key types">
 {
   "@context": [
     "https://www.w3.org/ns/did/v1",
-    "https://w3id.org/security/suites/jws-2020/v1",
-    "https://w3id.org/security/suites/ed25519-2020/v1"
+    "https://w3id.org/security/suites/jws-2020/v1"
   ],
-  "id": "did:example:123",
   "verificationMethod": [
     {
-      "id": "did:example:123#ZC2jXTO6t4R501bfCXv3RxarZyUbdP2w_psLwMuY6ec",
-      "type": "Ed25519VerificationKey2020", <span class="comment">// external (property value)</span>
-      "controller": "did:example:123",
-      "publicKeyMultibase": "zH3C2AVvLMv6gmMNam3uVAjZpfkcJCwDwnZn6z3wXmqPV"
-    },
-    {
-      "id": "did:example:123#zQ3shP2mWsZYWgvgM11nenXRTx9L1yiJKmkf9dfX7NaMKb1pX",
-      "type": "EcdsaSecp256k1VerificationKey2019", <span class="comment">// external (property value)</span>
-      "controller": "did:example:123",
-      "publicKeyBase58": "d5cW2R53NHTTkv7EQSYR8YxaKx7MVCcchjmK5EgCNXxo",
-    },
-    {
-      "id": "did:example:123#_Qq0UL2Fq651Q0Fjd6TvnYE-faHiOpRlPVQcY_-tA4A",
-      "type": "JsonWebKey2020", <span class="comment">// external (property value)</span>
+      "id": "did:example:123#key-0",
+      "type": "JsonWebKey2020",
       "controller": "did:example:123",
       "publicKeyJwk": {
-        "kty": "OKP", <span class="comment">// external (property name)</span>
-        "crv": "Ed25519", <span class="comment">// external (property name)</span>
-        "x": "VCpo2LMLhn6iWku8MKvSLg2ZAoC-nlOyPVQaO3FxVeQ" <span class="comment">// external (property name)</span>
+        "kty": "OKP", // external (property name)
+        "crv": "Ed25519", // external (property name)
+        "x": "VCpo2LMLhn6iWku8MKvSLg2ZAoC-nlOyPVQaO3FxVeQ" // external (property name)
       }
     },
     {
-      "id": "did:example:123#z6LSnjagzhe8Df6gZmroW3wjDd7XQLwAuYfwa4ZeTBCGFoYc",
-      "type": "JsonWebKey2020", <span class="comment">// external (property value)</span>
+      "id": "did:example:123#key-1",
+      "type": "JsonWebKey2020",
       "controller": "did:example:123",
       "publicKeyJwk": {
-        "kty": "OKP", <span class="comment">// external (property name)</span>
-        "crv": "X25519", <span class="comment">// external (property name)</span>
-        "x": "pE_mG098rdQjY3MKK2D5SUQ6ZOEW3a6Z6T7Z4SgnzCE" <span class="comment">// external (property name)</span>
-      },
-    }
-    {
-      "id": "did:example:123#4SZ-StXrp5Yd4_4rxHVTCYTHyt4zyPfN1fIuYsm6k3A",
-      "type": "JsonWebKey2020", <span class="comment">// external (property value)</span>
-      "controller": "did:example:123",
-      "publicKeyJwk": {
-        "kty": "EC", <span class="comment">// external (property name)</span>
-        "crv": "secp256k1", <span class="comment">// external (property name)</span>
-        "x": "Z4Y3NNOxv0J6tCgqOBFnHnaZhJF6LdulT7z8A-2D5_8", <span class="comment">// external (property name)</span>
-        "y": "i5a2NtJoUKXkLm6q8nOEu9WOkso1Ag6FTUT6k_LMnGk" <span class="comment">// external (property name)</span>
+        "kty": "OKP", // external (property name)
+        "crv": "X25519", // external (property name)
+        "x": "pE_mG098rdQjY3MKK2D5SUQ6ZOEW3a6Z6T7Z4SgnzCE" // external (property name)
       }
     },
     {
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -5045,103 +5045,149 @@ <h2>DID Documents</h2>
...
-        "y": "jq4QoAHKiIzezDp88s_cxSPXtuXYFliuCGndgU4Qp8l91xzD1spCmFIzQgVjqvcP" <span class="comment">// external (property name)</span>
+        "kty": "RSA", // external (property name)
+        "e": "AQAB", // external (property name)
+        "n": "UkhWaGJGOUZRMTlFVWtKSElBdENGV2hlU1F2djFNRXh1NVJMQ01UNGpWazlraEpLdjhKZU1YV2UzYldIYXRqUHNrZGYyZGxhR2tXNVFqdE9uVUtMNzQybXZyNHRDbGRLUzNVTElhVDFoSkluTUhIeGoyZ2N1Yk82ZUVlZ0FDUTRRU3U5TE8wSC1MTV9MM0RzUkFCQjdRamE4SGVjcHl1c3BXMVR1X0RicXhjU253ZW5kYW13TDUyVjE3ZUtobE80dVh3djJIRmx4dWZGSE0wS21DSnVqSUt5QXhqRF9tM3FfX0lpSFVWSEQxdERJRXZMUGhHOUF6c24zajk1ZC1zYU" // external (property name)
       }
+    }
+  ]
+}        
+      </pre>
+      <pre class="example" title="DID Document with different verification method types">
```
OR13, line 5143: @msporny I tried to hint at this being less than ideal by making it the 3rd example, and adding even more types.

index.html
```diff
@@ -5045,103 +5045,149 @@ <h2>DID Documents</h2>
...
       </pre>
 
-      <pre class="example" title="DID Document with many different verification methods">
+      <pre class="example" title="DID Document with many different key types">
 {
   "@context": [
     "https://www.w3.org/ns/did/v1",
-    "https://w3id.org/security/suites/jws-2020/v1",
-    "https://w3id.org/security/suites/ed25519-2020/v1"
+    "https://w3id.org/security/suites/jws-2020/v1"
```
OR13, line 5052: I updated this second example to mirror the first example. 

The ordering still implies that multibase is better to use, but since the section is informative, I am not concerned about that.

brentzundel (approved):

peacekeeper:
I'm not sure if I like splitting up the verification methods example into two examples - one that highlights only `JsonWebKey2020`, and one that actually shows different verification methods.

I don't think the DID Core specification should give the impression that one verification method is more "special" than others.

OR13:
@peacekeeper I agree, but it was worse before... it highlighted only `publicKeyMultibase`.

both `publicKeyMultibase` and `publicKeyJwk` are defined by DID Core....

The other examples are not.

I don't see a good way around this issue, given we were unable to agree to a single verification material representation, the best we can do is provide examples of the 2 presentations we did agree to include, and then a 3rd example which shows how bad this can get.

even ordering the examples starting with multibase could be argued as bias... there was clearly no consensus on this topic.

I might argue that `publicKeyJwk` should be listed first, since it relies on a real standard.

But I am willing to let the informative examples hint that a JSON only verification method format might not be the best foot to lead with for an abstract data model based spec.

peacekeeper:
> it highlighted only `publicKeyMultibase`

I don't understand. The current example https://www.w3.org/TR/did-core/#example-31-did-document-with-many-different-verification-methods shows 1x `publicKeyMultibase`, 1x `publicKeyBase58`, and 7x `publicKeyJwk`.

So `publicKeyJwk` is already given much more attention than the others. This PR here seems to highlight `publicKeyJwk` even more by giving it its own separate example block. Why?

index.html
```diff
@@ -5045,103 +5045,149 @@ <h2>DID Documents</h2>
...
+      "id": "did:example:123#key-2",
+      "type": "EcdsaSecp256k1VerificationKey2019",
+      "controller": "did:example:123",
+      "publicKeyBase58": "ns2aFDq25fEV1NUd3wZ65sgj5QjFW8JCAHdUJfLwfodt" // external (property name)
+    },
+    {
+      "id": "did:example:123#key-3",
+      "type": "EcdsaSecp256k1RecoverySignature2020",
+      "controller": "did:example:123",
+      "ethereumAddress": "0x89a932207c485f85226d86f7cd486a89a24fcc12" // external (property name)
```
peacekeeper, line 5176: -1 to adding an `ethereumAddress` example, which is [marked as deprecated](https://w3c.github.io/did-spec-registries/#ethereumaddress), and [has previously raised concerns about brand names](https://github.com/w3c/did-spec-registries/pull/73).

peacekeeper (changes requested):

index.html
OR13: @peacekeeper I removed the ethereum address example.

OR13:
@peacekeeper  the first example uses `publicKeyMultibase` only:

https://www.w3.org/TR/did-core/#example-30-did-document-with-1-verification-method-type

```
{
    "@context": [
      "https://w3id.org/security/suites/ed25519-2020/v1"
    ],
    "id": "did:example:123",
    "authentication": [
      {
        "id": "did:example:123#z6MkecaLyHuYWkayBDLw5ihndj3T1m6zKTGqau3A51G7RBf3",
        "type": "Ed25519VerificationKey2020", // external (property value)
        "publicKeyMultibase": "zAKJP3f7BD6W4iWEQ9jwndVTCBq8ua2Utt8EEjJ6Vxsf"
      }
    ],
    "capabilityInvocation": [
      {
        "id": "did:example:123#z6MkhdmzFu659ZJ4XKj31vtEDmjvsi5yDZG5L7Caz63oP39k",
        "type": "Ed25519VerificationKey2020", // external (property value)
        "publicKeyMultibase": "z4BWwfeqdp1obQptLLMvPNgBw48p7og1ie6Hf9p5nTpNN"
      }
    ],
    "capabilityDelegation": [
      {
        "id": "did:example:123#z6Mkw94ByR26zMSkNdCUi6FNRsWnc2DFEeDXyBGJ5KTzSWyi",
        "type": "Ed25519VerificationKey2020", // external (property value)
        "publicKeyMultibase": "zHgo9PAmfeoxHG8Mn2XHXamxnnSwPpkyBHAMNF3VyXJCL"
      }
    ],
    "assertionMethod": [
      {
        "id": "did:example:123#z6MkiukuAuQAE8ozxvmahnQGzApvtW7KT5XXKfojjwbdEomY",
        "type": "Ed25519VerificationKey2020", // external (property value)
        "publicKeyMultibase": "z5TVraf9itbKXrRvt2DSS95Gw4vqU3CHAdetoufdcKazA"
      }
    ]
}
```

peacekeeper:
@OR13 but what's the motivation for introducing a new example that focuses only on the `JsonWebKey2020` verification method, when there is already an existing example that shows several verification methods including `JsonWebKey2020`?

Is the idea that you want to highlight how a single verification method can support many different key types? If so, then I think that's useful, but maybe we could swap the two examples, so the one with several different verification methods comes first, and then the one that focuses only on `JsonWebKey2020`? Would that work for you?

kdenhartog (approved):

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2019/did-wg/Meetings/Minutes/2021-05-25-did#section3-5) on 2021-05-25

- no resolutions were taken

#### 3.5. Examples might be out of date 

_See github issue [#734](https://github.com/w3c/did-core/issues/734)._




_See github pull request [#748](https://github.com/w3c/did-core/pull/748)._




**Drummond Reed:** I submitted PRs on that and the next issue; they just need review  

**Orie Steele:** I removed the Ethereum address example  
… there is still a potential issue with having a public key multibase example first  
… so I hoped that providing several examples would solve the problem  

**Brent Zundel:** a link to the PR is in IRC so please review  

</details>

OR13:
@peacekeeper

I think the best we can do is:

1. example using `publicKeyMultibase`
2. example using `publicKeyJwk`
3. example using many types.

Thats what I did in this PR.

I could also live with:

1. example using many types.

I think its pretty obvious that:

1. example using `publicKeyMultibase`
2. example using many types.

is using non normative examples to bias against `publicKeyJwk` which is just as well defined in did core as `publicKeyMultibase`.

OR13:
@peacekeeper another option:

1. publicKeyMultibase + publicKeyJwk
2. example using many types

The problem with this is that it sorta sucks to consume keys are varying types from a did document... its better to have them all `publicKeyMultibase` or `publicKeyJwk` imo... even those a mix is legal according to the spec.

selfissued (approved):
I believe this is an improvement on what was there.

OR13:
Another note, `publicKeyJwk` and `publicKeyMultibase` are actually key types... they just represent the type information differently... whereas `publicKeyJwk` and `publicKeyBase58` are not both types.... you don't know the `crv` or `kty` values from `publicKeyBase58` but you do from `publicKeyMultibase`... imo this is a major reason why `publicKeyMultibase` is superior to `publicKeyBase58`

msporny:

msporny:
I can confirm that my concerns have been address with this PR for issue #734.

PR #753, opened 2021-05-24
Add advisement about signed DID documents.

peacekeeper:
As discussed in https://github.com/w3c/did-core/pull/738#pullrequestreview-662135911, this adds some language about signed DID documents (similar to what existed before merging https://github.com/w3c/did-core/pull/738).

### DIFF
--- a/index.html
+++ b/index.html
@@ -4061,6 +4061,17 @@ associated with the <a>verification methods</a> can be used to generate a
 cryptographic digital signature as a part of an authentication or authorization
 security protocol.
         </p>
+
+        <p class="advisement" title="Signed DID documents">
+Some <a>DID methods</a> allow digital signatures and other proofs to be
+included in the <a>DID document</a> or a <a href="#metadata-structure"></a>.
+However, such proofs by themselves do not necessarily prove control over a
+<a>DID</a>, or guarantee that the <a>DID document</a> is the correct one for
+the <a>DID</a>. In order to obtain
+the correct <a>DID document</a> and verify control over a <a>DID</a>, it is
+necessary to perform the <a>DID resolution</a> process as defined by the
+<a>DID method</a>.
+        </p>
       </section>
 
       <section class="notoc">


alenhorvat:
I agree with the statement and is great to add the description.
Other things to consider:
- Should we state: proofs (signatures) can only protect the DID Document integrity or serve as proof of private key ownership. Proof of key ownership is static and only proves the owner was capable to sign a self-issued challenge.

samu-gataca:
As you said in the update, the signature does not prove control over the DID, if we are talking about the creation, but in all the next operations about any update (rotating, revoking or adding new keys), it could help to demonstrate who is the owner for that DID Document (because, just the owner has the private key to be allowed to change de DID).
I agree that putting the signature on the document may not be the best option, it could be included in the metadata section. 

The main problem here is, if we want to avoid strong dependencies with the infra layer, we need a point of trust. It could be at the DID Registry level, or the DID Document level. At the level of the DID Registry, any party needs to trust in the vendor which is performing the DID Resolution. The same happens when the trust depends on the signature on the DID Document, it's necessary to trust in the DID Registry owner (if the DID Registry try to cheat the other party, it's easy to demonstrate based on the signature), but at the end the end user could check the validity of the DID Document, always basing this idea, not in the creation, however, we can demonstrate, all the updates over the same DID Document provides from the same source.
Once we have several keys for several different purposes, and each of them of any type (RSA, ed25519, secp, ...) we can not based our solution in any algorithm feature as correlate the DID with the hash of the public key, because the DID Document is a dynamic object, so we need any external help.

OR13 (approved):

kdenhartog (approved):

index.html
```diff
@@ -4061,6 +4061,17 @@ <h3>Proving Control of a DID and/or DID Document</h3>
 cryptographic digital signature as a part of an authentication or authorization
 security protocol.
         </p>
+
+        <p class="advisement" title="Signed DID documents">
```
msporny, line 4065:
```suggestion
        <p class="note" title="Signed DID documents">
```

msporny (approved):

msporny:
Editorial, multiple positive reviews, one editorial change requested an made (advisement -> note), no objections, merging.

Issue #718, opened 2021-03-29
FAQ Question: Can the DID Subject of a DID Document be updated/changed after the DID Document has been registered?


mwherman2000:
New FAQ Question for the https://w3c.github.io/did-core/#frequently-asked-questions-about-did-identification section:

Can the DID Subject of a DID Document be updated/changed after the DID Document has been registered?

I presume no?

msporny:
While not explicitly mentioned (this is a DID Method decision)... I have heard of no DID Method that would allow this.

OR13:
example would be DID Subject is the owner of house at this address....

OR13:
certainly the did controller can decide to change what the did identifies. +1 to @talltree comments on the wg call.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2019/did-wg/Meetings/Minutes/2021-03-30-did#section5-8) on 2021-03-30

- no resolutions were taken

<details><summary><i>View the transcript</i></summary>

#### 5.8. FAQ Question: Can the DID Subject of a DID Document be updated/changed after the DID Document has been registered? 

_See github issue [#718](https://github.com/w3c/did-core/issues/718)._




**Brent Zundel:** This would be another thing if drummond can in his refactoring, make sure this question ends up query  
… I think it's clear from the rest of the document... but making it clear would be valuable  

**Drummond Reed:** OK, I will take that one too.  
… Although the answer to earlier one seems obvious, I think there may be different answers for this one, depending on who you ask.  
… Do we have a well-documented reply to answer?  

**Manu Sporny:** This is not explicitly mentioned anywhere I think. It's really DID Method specific.  
… I don't know of any DID method that allows you do change the DID Subject.  
… Could imagine ... tombstoning... to make it seem like that happened.  

**Drummond Reed:** It might be theoretically possible to that, but we don't know of any. Or we could flat out say no.  

**Brent Zundel:** Or you could say whether or not it is possible is DID Method-specific.  

**Kyle Den Hartog:** Advocating for no. Huge can of worms  
… we did a good job avoiding.  

**Brent Zundel:** I know of at least one working group participant who would say yes. If we say DID method-specific, we could avoid that  

**Orie Steele:** I agree with what you just said, brent.  
… Example is DID subject is human owner at house. It changes over time. Does the subject change? I don't think so.  
… But depending on how you read English, you might come up with different ways of answering ti.  

[...]

Issue #719, opened 2021-03-29
FAQ Question: Can the DID Controller of a DID Document be updated/changed after the DID Document has been registered?


mwherman2000:
New FAQ Question for the https://w3c.github.io/did-core/#frequently-asked-questions-about-did-identification section:

Can the DID Controller of a DID Document be updated/changed after the DID Document has been registered?

I presume yes?

msporny:
Yes, it can.

OR13:
We describe wallet caching for offline use cases here: https://w3c-ccg.github.io/universal-wallet-interop-spec/#DIDResolutionResponse

I think generally there should be some warnings about caching and trusting network enabled resolvers.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2019/did-wg/Meetings/Minutes/2021-03-30-did#section5-7) on 2021-03-30

- no resolutions were taken

<details><summary><i>View the transcript</i></summary>

#### 5.7. FAQ Question: Can the DID Controller of a DID Document be updated/changed after the DID Document has been registered? 

_See github issue [#719](https://github.com/w3c/did-core/issues/719)._




**Brent Zundel:** Currently unassigned, and has one comment.  
… I don't know that more needs to happen  

**Manu Sporny:** There is a request to add a FAQ for it.  
… drummond, are you refactoring the end of the document to be a FAQ?  

**Drummond Reed:** I thought the goal was to refactor it to not be a FAQ  

**Manu Sporny:** My bad, the other way then  
… Never mind  
… I was going to say... can you speak to it ...  

**Drummond Reed:** Definitely. Feels like one of the things to discuss  
… If you want me to cover it, I'm happy to do that  
… All the update will do is change the questions into headings. It wasn't a FAQ originally ...  
… I'll take that action item.  

</details>

msporny:
@talltree ping on this issue, can you write a PR to address it?

talltree:
Yes, working on it.

msporny:
PR #754 has been raised to address this issue. This issue will be closed once that PR has been merged.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2019/did-wg/Meetings/Minutes/2021-05-25-did#section3-4) on 2021-05-25

- no resolutions were taken

<details><summary><i>View the transcript</i></summary>

#### 3.4. FAQ Question: Can the DID Controller of a DID Document be updated/changed after the DID Document has been registered? (issue did-core#719)

[...]

PR #755, opened 2021-05-25
add appendix text RE did subject change

talltree:

This addresses issue #718.

***
<a href="https://pr-preview.s3.amazonaws.com/talltree/did-core/pull/755.html" title="Last updated on Jun 10, 2021, 10:00 PM UTC (f77df37)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/755/443e944...talltree:f77df37.html" title="Last updated on Jun 10, 2021, 10:00 PM UTC (f77df37)">Diff</a>


index.html
```diff
@@ -5838,6 +5838,20 @@ <h3>Group Control</h3>
...
+    <section>
+      <h2>Changing the DID subject</h2>
+      <p>
+A <a>DID document</a> has exactly one <a>DID subject</a> identified by exactly
+one <a>DID</a> expressed as the value of the the <code><a>id</a></code>
+property. This property value is immutable for the lifetime of the
+<a>DID document</a>.
+      </p>
+      <p>
+However it is possible that the resource <em>identified</em> by the <a>DID</a>
```
TallTed, line 5850:
```suggestion
However, it is possible that the resource <em>identified</em> by the <a>DID</a>
```

index.html
```diff
@@ -5838,6 +5838,20 @@ <h3>Group Control</h3>
         </p>
       </section>
     </section>
+    <section>
+      <h2>Changing the DID subject</h2>
+      <p>
+A <a>DID document</a> has exactly one <a>DID subject</a>, identified by exactly
```
kdenhartog, line 5844:
```suggestion
A <a>DID document</a> should have one <a>DID subject</a>, identified by exactly
```

Not sure this is correct to be honest. Just wanting to offer some sort of definitive change.

This text seems to conflict with statements in the persistent section. In particular:

>It is technically possible for the controller to reuse a DID for different subjects and, more subtly, for the precise definition of the Subject to either change over time or be misunderstood. 

>Because of these potential ambiguities, DIDs should be considered valid contextually rather than absolutely. Their persistence does not imply that they refer to the exact same Subject, nor that they are under the control of the same controller. Instead, one needs to understand the context in which the DID was created, how it is used, and consider the likely shifts in their meaning, and adopt procedures and policies to address both potential and inevitable semantic drift. 

From this section I infer that it's not possible and shouldn't be done, but from the persistence section I infer that it's a consideration that's left contextually up to the party relying upon the DID to decide how they'll deal with it.

kdenhartog (commented):
This is subtle enough and articulated well enough in the second paragraph (while linking to the additional text). I just felt like it's potentially conflicting and so don't want to approve, but also don't believe it's big enough of a deal to warrant requesting changes. I'm +0 on the text as is and +1 on adding the section.

index.html
TallTed: @kdenhartog --- 

I think what's missing from your reading, and perhaps from the writing, is "at a time" --- i.e., `A <a>DID document</a> has exactly one <a>DID subject</a>[ at a time], ...`.  This allows for a DID subject to be a temporal or collective entity, which is identified by a single URI; but does not allow for a DID subject to be a collection of entities, each identified by its own single URI.

The other sections you quoted are not about having multiple URIs which identify multiple subject entities in a single DID document, but about having a single URI which identifies multiple subject entities *over time*.  In egregious pseudo-code, `did:example:xyz did:subject example:orgPresident` could be used to designate the subject of the DID as the President of an organization, which role might shift from one person to another over time.  The `example:orgPresident` is always the subject of that DID, but the person in that role will change periodically, so the effective *meaning* of the DID subject will change.

Perhaps that helps?

kdenhartog: +1 yeah that's a great nuance that makes them aligned. I'll update my suggestion with your text and approve pending the merging up that change.

kdenhartog (approved):
Approving pending the suggested text to align with the persistence text. Thank you @TallTed for the clarifying language to help bring out that important nuance.

TallTed:
My "at a time" suggestion was very off-the-cuff, and this segment of text would benefit from a bit more word-smithing, because as it stands (especially with that addition), it might be read to suggest that the *URI* of the DID Subject might change, and I don't believe *that* is what we mean to say; rather it's the *meaning* of the URI of the DID Subject that might change.

index.html
```diff
@@ -5838,6 +5838,20 @@ <h3>Group Control</h3>
...
+one <a>DID</a>, expressed as the value of the the <code><a>id</a></code>
+property. This property value is immutable for the lifetime of the
```
dlongley, line 5846: I think the trouble can be resolved like this:

```suggestion
A <a>DID document</a> has exactly one <a>DID</a>, which refers to
a <a>DID subject</a>. The <a>DID</a> is expressed as the value of the
<code><a>id</a></code> property. This property value is immutable for
the lifetime of the
```

index.html
```diff
@@ -5838,6 +5838,20 @@ <h3>Group Control</h3>
...
+<a>DID document</a>.
+      </p>
+      <p>
+However, it is possible that the resource <em>identified</em> by the <a>DID</a>
+may change over time. This is under the exclusive authority of the
+<a>DID controller</a>. For more details, see section <a href="#persistence"></a>.
```
dlongley, line 5852: I think this follow-on paragraph makes it clear that the DID subject may change over time. But editorially, we could add this to clarify:

```suggestion
However, it is possible that the resource <em>identified</em> by the <a>DID</a>,
the <a>DID subject</a>, may change over time. This is under the exclusive authority
of the <a>DID controller</a>. For more details, see section <a href="#persistence"></a>.
```

index.html
kdenhartog: Yeah I'd be good with this change as well.

kdenhartog:
Just commenting here. After re-reading this text again, I'm fine with it as is. My interpretation originally was that this text doesn't align well, but I now understand it differently. This reads to me now that the first paragraph is about the capabilities of how the DID Document, DID, and DID Subject are technically related. The second paragraph focuses on pointing out that the meaning is update-able by referencing the other text in persistence which gets to the heart of #718 

I'll leave my approval and no longer feel its necessary to update the text.

talltree:
I felt that @dlongley 's suggestions helped clarify any potential misunderstandings, so I accepted both of them.

TallTed (approved):

brentzundel (approved):

peacekeeper (approved):

talltree:
@msporny I was very puzzled as to why there would be a merge conflict until I realized (with @brentzundel 's help) that it was simply that I contributed two new sections of the Appendix (in responses to issues #718 and #719), and because both appeared at the end of the current Appendix text, GitHub thought they were in conflict with each other. So it was a quick fix to accept both. Should be ready to merge now.

iherman:
Administratively, this should not be merged until the CR snapshot is published (hopefully next Tuesday). Otherwise, we have to update the 'previous version' reference in the CR snapshot header (remember, merge means publishing a new version in /TR).

msporny:

PR #770, opened 2021-06-16
We're moving away from travis CI

plehegar:
This is using GitHub actions.

See also https://github.com/w3c/spec-prod

Note that Travis CI has been deactivated for the w3c/did-core repo so this needs to get merged asap.


iherman:
@msporny I did not want to merge this, 'cause I do not know whether you are doing some other merges right now. I let you do that. In theory (and in practice:-) this should work out of the box once merged.

msporny:
@iherman -- we have some merges to do, so let's take the new Github Actions workflows for a test run. :)

PR #799, opened 2021-10-04
Add section on Evaluating Competing Considerations.

msporny:
This PR adds a non-normative appendix section on resource usage trade-offs and points readers to the DID Rubric for further evaluation criteria.

***

### :boom: Error: connect ETIMEDOUT 128.30.52.89:443 :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Oct 6, 2021, 12:46 AM UTC)_.

<details>
<summary>More</summary>

PR Preview relies on a number of web services to run. There seems to be an issue with the following one:

:rotating_light: [Spec Generator](https://www.w3.org/2015/labs/) - Spec Generator is the web service used to build specs that rely on ReSpec.

:link: [Related URL](https://labs.w3.org/spec-generator/?type=respec&url=https%3A%2F%2Frawcdn.githack.com%2Fw3c%2Fdid-core%2Fe95f2b33d16606bb6a4ea211082f4b9e8e68f235%2Findex.html%3FisPreview%3Dtrue)

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20w3c/did-core%23799.)._
</details>

### DIFF
--- a/index.html
+++ b/index.html
@@ -5036,11 +5036,11 @@ This specification does not require or suggest the use of any specific type of
 requirements. Different requirements might suggest different considerations with different
 trade-offs. For example, trade-offs between computation (energy usage), trust
 (deference to authority), coordination (network bandwidth), or memory (physical
-storage) might or might not be appropriate, for any given use case. Other use cases
+storage) might or might not be appropriate for any given use case. Other use cases
 might not make the same trade-offs. Those that need to consider different criteria
 for their use case are directed to the <a
 href="https://www.w3.org/TR/did-rubric/">DID Method Rubric</a>, which provides
-evaluation criteria to help decision makers evaluate whether or not a particular
+evaluation criteria to help decision makers determine whether or not a particular
 <a>DID Method</a> is appropriate for their use cases.
       </p>
     </section>
--- a/index.html
+++ b/index.html
@@ -5026,6 +5026,24 @@ as described in <a href="#extensibility"></a>, and where <a
 href="#privacy-considerations"></a> is applicable for such extensions.
       </p>
     </section>
+
+    <section>
+      <h2>Resource Usage Trade-offs</h2>
+
+      <p>
+This specification does not require or suggest the use of any specific type of
+<a>verifiable data registry</a>. Different use cases might result in different
+requirements. Different requirements might suggest different resource usage
+trade-offs. For example, trade-offs between computation (energy usage), trust
+(deference to authority), coordination (network bandwidth), or memory (physical
+storage) might be appropriate for a certain threat model. Other threat models
+might not make the same trade-offs. Those that need to consider resource usage
+trade-offs for their threat models are directed to the <a
+href="https://www.w3.org/TR/did-rubric/">DID Method Rubric</a>, which provides
+evaluation criteria to help decision makers evaluate whether or not a particular
+<a>verifiable data registry</a> is appropriate for their use cases.
+      </p>
+    </section>
   </section>
 
   <section class="informative">
--- a/index.html
+++ b/index.html
@@ -5027,7 +5027,7 @@ href="#privacy-considerations"></a> is applicable for such extensions.
       </p>
     </section>
 
-    <section>
+    <section class="informative">
       <h2>Resource Usage Trade-offs</h2>
 
       <p>
--- a/index.html
+++ b/index.html
@@ -5028,20 +5028,20 @@ href="#privacy-considerations"></a> is applicable for such extensions.
     </section>
 
     <section class="informative">
-      <h2>Resource Usage Trade-offs</h2>
+      <h2>Evaluating Competing Considerations</h2>
 
       <p>
 This specification does not require or suggest the use of any specific type of
 <a>verifiable data registry</a>. Different use cases might result in different
-requirements. Different requirements might suggest different resource usage
+requirements. Different requirements might suggest different considerations with different
 trade-offs. For example, trade-offs between computation (energy usage), trust
 (deference to authority), coordination (network bandwidth), or memory (physical
-storage) might be appropriate for a certain threat model. Other threat models
-might not make the same trade-offs. Those that need to consider resource usage
-trade-offs for their threat models are directed to the <a
+storage) might or might not be appropriate, for any given use case. Other use cases
+might not make the same trade-offs. Those that need to consider different criteria
+for their use case are directed to the <a
 href="https://www.w3.org/TR/did-rubric/">DID Method Rubric</a>, which provides
 evaluation criteria to help decision makers evaluate whether or not a particular
-<a>verifiable data registry</a> is appropriate for their use cases.
+<a>DID Method</a> is appropriate for their use cases.
       </p>
     </section>
   </section>


index.html
```diff
@@ -5026,6 +5026,24 @@ <h2>Level of Assurance</h2>
...
+      <h2>Resource Usage Trade-offs</h2>
+
+      <p>
+This specification does not require or suggest the use of any specific type of
+<a>verifiable data registry</a>. Different use cases might result in different
+requirements. Different requirements might suggest different resource usage
+trade-offs. For example, trade-offs between computation (energy usage), trust
+(deference to authority), coordination (network bandwidth), or memory (physical
+storage) might be appropriate for a certain threat model. Other threat models
+might not make the same trade-offs. Those that need to consider resource usage
```
jandrieu, line 5040:
```suggestion
might not make the same trade-offs. Those that need to consider different criteria
```

index.html
```diff
@@ -5026,6 +5026,24 @@ <h2>Level of Assurance</h2>
...
+trade-offs for their threat models are directed to the <a
```
jandrieu, line 5041:
```suggestion
for their use case are directed to the <a
```

index.html
```diff
@@ -5026,6 +5026,24 @@ <h2>Level of Assurance</h2>
...
+href="https://www.w3.org/TR/did-rubric/">DID Method Rubric</a>, which provides
+evaluation criteria to help decision makers evaluate whether or not a particular
+<a>verifiable data registry</a> is appropriate for their use cases.
```
jandrieu, line 5044:
```suggestion
<a>DID Method</a> is appropriate for their use cases.
```

jandrieu (changes requested):
I'm surprised to see this PoW attack resurface in the guise of "resource usage".

I've rewritten as a generic pointer to the Rubric for evaluating competing considerations.

None of the existing VDRs that use PoW actually require PoW. Rather, they leverage existing PoW networks, to which they each add negligible marginal energy consumption.

Please stop amplifying this thread of disinformation. It's not doing anyone any good.

mprorock:
> I've rewritten as a generic pointer to the Rubric for evaluating competing considerations.

I am in line with @jandrieu on this one.  DID-CORE is a data model spec and is agnostic to any of the underlying mechanisms that may or may not be used by a particular DID method.  The impl guide deals with guidance to developers of the data model spec, and the rubric provides mechanism to evaluate various methods for a particular use case.  The data model itself doesn't have any horse in this race, and nor should it, otherwise i would have opened [pr #27](https://github.com/w3c/did-imp-guide/pull/27) here instead of against the impl guide.

OR13 (changes requested):
I'd like to see W3C create a policy for applying EWP, consistently to all specs.

Allowing big companies to leverage these documents to obstruct specs selectively should not be reinforced by applying a policy of appeasement.

See also: 

- https://github.com/w3ctag/ethical-web-principles/issues/52
- https://github.com/w3ctag/ethical-web-principles/issues/5
- https://github.com/w3ctag/ethical-web-principles/issues/42
- https://github.com/w3ctag/design-principles/issues/338

I agree with @jandrieu's suggestions and the spirit of this PR.

TallTed:
@jandrieu -- Does *every* possible future PoW-based DID solution use existing PoW deployments, and so cost only the marginal, rather than the initial (ahem) burn?

If not, then how about rewriting a little bit more, dropping your knee-jerk defense in the same way you're asking others to drop their knee-jerk condemnation, and making plain what's being treated as having negligible impact, and what could (new PoW networks *could* be built with care for all such impacts, but it's very easy to just plug a bunch of machines into the wall and pay no attention to the generation method of those electron circuits) have significant impact?

TallTed:
> Just because you can implement a JSON parser in an energy wasteful language on energy wasteful hardware, doesn't mean JSON wastes energy.

And vice versa.  Nothing's perfectly bad, and nothing's perfectly good.

The human ideal is to ethically and carefully _balance_ all the competing factors, not to declare once-and-for-all that anything is the be-all-and-end-all of good nor evil.

OR13:
> And vice versa. 

@TallTed what is the other side of this? 

Are you arguing that because some bad programs exist all languages that support them should have EWP disclaimers?

I could agree with that... we should start with JavaScript since developers have already written lots of malware that is indistinguishable from ad tech, and which violates user privacy...

Also turing machines... I have yet to see a turing machine implementation that was not an environmental catastrophe ; )

I'm not opposed to the section, I am opposed to selective enforcement of moral / ethical arguments which cannot be proved formally or "won" or "lost"... such arguments really ought to happen somewhere... and an appendix, implementation guide, or rubric are all fine places... perhaps W3C should formalize this as a requirement.

mprorock:
> perhaps W3C should formalize this as a requirement.

+1

This i think is part of the issue.  Objections of this nature are just now popping out in the wild in a real way, and unfortunately are blocking progress on this work item.  That does not mean that it is not a real concern, it very much is, but we as a community need to identify the right place, format, and policies, for how to cite and think about items such as environmental impacts on a consistent basis across the various outputs of working groups.

My bigger issue on this case, and why i think it needs some re-wording, is that this particular item is very much a data model spec, and that is very different than a specific method or implementation that might indeed have some highly negative impacts around environment, privacy, or key ethical items of note.

TallTed:
@OR13 --

> @TallTed what is the other side of this?

I thought that was obvious from context...  Since it wasn't: "Just because you can implement a JSON parser in an energy efficient language on energy efficient hardware, doesn't mean JSON efficiently uses energy."

OR13:
@TallTed thanks for spelling it out :)

So to generalize your argument, we SHOULD (MUST?) consider ESE regardless of spec type (data model, protocol, concrete implementation)?

dlongley (commented):
+1 to @jandrieu's improvements.

msporny:
I've applied all of @jandrieu's suggestions. Everyone please re-review, it's hard for me to tell from the discussion if there remain any objections, or if @jandrieu's changes have addressed everyone's concerns.

ChristopherA:
+1 from me. Balanced.

rxgrant:
I think Joe's rewrite, as currently applied, makes it tolerable.  

Another way of saying the issue is that *you're not incentivizing more than you're paying for*.

Let's take a pass on the [struggle session](https://en.wikipedia.org/wiki/Struggle_session) that is being offered.

peacekeeper (approved):
Very good.. Yes, there are trade-offs depending on the use case and requirements.

jandrieu:
> @jandrieu -- Does _every_ possible future PoW-based DID solution use existing PoW deployments, and so cost only the marginal, rather than the initial (ahem) burn?
> 
> If not, then how about rewriting a little bit more, dropping your knee-jerk defense in the same way you're asking others to drop their knee-jerk condemnation, and making plain what's being treated as having negligible impact, and what could (new PoW networks _could_ be built with care for all such impacts, but it's very easy to just plug a bunch of machines into the wall and pay no attention to the generation method of those electron circuits) have significant impact?

Your comment is an good illustration of why this PR as initially proposed is inappropriate: there is no justifiable argument for calling out resource usage or PoW as a particularly problematic issue. The PR itself doesn't say anything about PoW, you CLEARLY perceive that is is, and yet we have yet to see any legitimate arguments concerned about actual resource use by actual DID methods. As long as the language triggers anti-PoW responses, it's ceases to be a technical conversation and moves into political posturing.

Note that in the implementation guide text, I proposed language that does exactly what you propose (explaining the nuance of marginal use of existing infrastructure). This PR, however, is a different matter. As a pointer to the Rubric, it should, IMO be simply an introduction to the inevitable tradeoffs due to different considerations. It should not take a position that prioritizes any particular tradeoff as better or worse than any other.

In particular, there is no valid argument that any known DID Method, on its own, is a significant problem wrt resource usage. As such, calling out resource usage is inappropriate, regardless of whether or not you like cryptocurrencies or think they may bring the end of civilization as we know it. The spec doesn't require any particular approach and none of the known approaches actually have the flaw that is being attacked. It's like accusing automobile manufacturers of the environmental harms of railroad tracks just because some cars are occasionally transported by train. Pure fearmongering. Blame car makers for the impact of cars, not trains. Blame DID Method creators and users for the marginal impact of their DID Method, not the totality of harms that might be associated with any underlying infrastructure.

As such, the inclusion of language implying that resource usage is a problem is at best erroneous, and it is more aptly described as intentional disinformation.

Bad science shouldn't be used to justify political attacks in a technical specification.

jandrieu (approved):
This looks good.

mprorock (approved):

rxgrant (approved):

dlongley (approved):

TallTed:
> [@jandrieu] The PR itself doesn't say anything about PoW, yet you perceive that it is,

Note that I was replying to *you*, to [*your* comments about the PR](https://github.com/w3c/did-core/pull/799#pullrequestreview-770197216), where *you* raised PoW! To wit:

> I'm surprised to see this PoW attack resurface in the guise of "resource usage".
>
> [...]
>
> None of the existing VDRs that use PoW actually require PoW. Rather, they leverage existing PoW networks, to which they each add negligible marginal energy consumption.

In other words: _I_ did not perceive the PR to be about PoW.  _You_ did, and should perhaps take the rest of your latest comment into your own heart.

index.html
```diff
@@ -5026,6 +5026,24 @@ <h2>Level of Assurance</h2>
...
+      <h2>Evaluating Competing Considerations</h2>
+
+      <p>
+This specification does not require or suggest the use of any specific type of
+<a>verifiable data registry</a>. Different use cases might result in different
+requirements. Different requirements might suggest different considerations with different
+trade-offs. For example, trade-offs between computation (energy usage), trust
+(deference to authority), coordination (network bandwidth), or memory (physical
+storage) might or might not be appropriate, for any given use case. Other use cases
+might not make the same trade-offs. Those that need to consider different criteria
```
TallTed, line 5040:
```suggestion
might not make the same trade-offs. Those that need to consider different factors
```

index.html
```diff
@@ -5026,6 +5026,24 @@ <h2>Level of Assurance</h2>
...
+for their use case are directed to the <a
+href="https://www.w3.org/TR/did-rubric/">DID Method Rubric</a>, which provides
+evaluation criteria to help decision makers evaluate whether or not a particular
```
TallTed, line 5043:
```suggestion
evaluation criteria to help decision makers determine whether or not a particular
```

index.html
```diff
@@ -5026,6 +5026,24 @@ <h2>Level of Assurance</h2>
...
+    <section class="informative">
+      <h2>Evaluating Competing Considerations</h2>
+
+      <p>
+This specification does not require or suggest the use of any specific type of
+<a>verifiable data registry</a>. Different use cases might result in different
+requirements. Different requirements might suggest different considerations with different
+trade-offs. For example, trade-offs between computation (energy usage), trust
+(deference to authority), coordination (network bandwidth), or memory (physical
+storage) might or might not be appropriate, for any given use case. Other use cases
```
TallTed, line 5039:
```suggestion
storage) might or might not be appropriate for any given use case. Other use cases
```

TallTed (changes requested):
A few tiny wording and punctuation tweaks.

index.html
msporny: Skipping this suggestion since the DID Rubric calls these things "criteria": https://www.w3.org/TR/did-rubric/#categories-of-criteria

TallTed: I'll live with it. Constructions like "those that need x should look at y which provides x" just make me itch. You can mark this as resolved if Github lets you. (It doesn't give me that option.)

OR13 (approved):

tasdienes:
To address any concerns about DID methods which utilize Ethereum wasting energy, we would like to point out that Ethereum is in the process of transitioning from Proof of Work to Proof of Stake. This will [reduce its energy consumption by roughly 99.95%](https://blog.ethereum.org/2021/05/18/country-power-no-more/).  The [transition ](https://ethereum.org/en/eth2/merge/)has been in the works for a long time, and is expected to be completed within the next 6 months. We believe that POS is the future for public blockchains, and POW chains will eventually become obsolete.

Tas Dienes, Andreas Freund, and Dan Shaw - Ethereum Foundation

csuwildcat:
Proof of Stake is 1) not nearly as decentralized a foundation for these systems, 2) introduces a much larger attack/vulnerability surface, and 2) shifts its 'work' to mechanisms that lead to oligarchical capture/centralization that can't be unseated by unilateral operator competition. There is no free lunch, regardless of how many Rube Goldbergian cogs and widgets one attempts to bolt on.

> To address any concerns about DID methods which utilize Ethereum wasting energy, we would like to point out that Ethereum is in the process of transitioning from Proof of Work to Proof of Stake. This will [reduce its energy consumption by roughly 99.95%](https://blog.ethereum.org/2021/05/18/country-power-no-more/). The [transition ](https://ethereum.org/en/eth2/merge/)has been in the works for a long time, and is expected to be completed within the next 6 months. We believe that POS is the future for public blockchains, and POW chains will eventually become obsolete.
> 
> Tas Dienes, Andreas Freund, and Dan Shaw - Ethereum Foundation

OR13:
@csuwildcat 

>  3) shifts its 'work' to mechanisms that lead to oligarchical capture/centralization

I think any free to join network suffers from the "rich sybil attack"... its a fact of life.

The difference between having unlimited money to spin up PoW nodes and unlimited money to leverage PoS nodes... is pure energy savings... but the rich guys still own the network ; )

rxgrant:
> The difference between having unlimited money to spin up PoW nodes and unlimited money to leverage PoS nodes... is pure energy savings... but the rich guys still own the network ; )

@OR13 &lt;sigh&gt; there is so much to say about this.

OR13:
@rxgrant at least we can agree that more money, less problems, amirite? ;) 

From a cyber security perspective, its a fact that the "cost of an attack" is a factor in pulling it off... sometimes that cost is pure compute, sometimes its compromising a hardware supply chain, sometimes its joining a consortium and threatening folks... or tricking your opponent into overspending...  attacking free to join networks is always a thing, and always costs money... sometimes it has other costs, environment or political, etc... 

It's also a fact that winning a fight against an adversary with more money, power etc... involves indirect conflict.

> The conventional army loses if it does not win. The guerrilla wins if he does not lose.
- Henry Kissinger


I suppose the question is: is decentralization a standing army or a guerrilla unit?

Therecanbeonlyone1969:
@csuwildcat 
> Proof of Stake is 1) not nearly as decentralized a foundation for these systems, 

over 270k+ validators on the beacon chain https://beaconscan.com/ -> that is quite a few clients running.  How is that not decentralized?

> 2) introduces a much larger attack/vulnerability surface, 

While the game-theoretic security analysis of Casper FFG is not trivial, it is sound and turns out to be not less safe than Bitcoin aka both PoS and PoW suffer from the same type of long-range attacks. 
 
> and 3) shifts its 'work' to mechanisms that lead to oligarchical capture/centralization that can't be unseated by unilateral operator competition. There is no free lunch, regardless of how many Rube Goldbergian cogs and widgets one attempts to bolt on.

Where is your evidence? Check this out https://beaconscan.com/validators ... biggest validators all have ~ 66 Eth and the smallest ones have ~30 Eth (minimal stake) across 270k+ validators. That looks like a very flat distribution to me. Granted, one or few individuals could control tens of thousands of validators ... but since each validator is its own stack that is a non-trivial server farm, and a significant capital investment and given the node distribution, it is geographically localized (US and EU but loads of non-clustered nodes in those geos).

So, it would be awesome, if you could produce data that refutes what I just posted.

msporny:
Class 2, editorial, multiple reviews, changes requested and made, no objections, merging.

Issue #159, opened 2020-01-29
Remove matrix parameters from the DID specification


selfissued:
We should enable the use of standard URI parsers by removing matrix parameters from the DID specification.

As I wrote at https://github.com/w3c/did-core/issues/137#issuecomment-566604194, in my view, the working group should make every attempt to not introduce the matrix parameters syntax at all. There are already two mechanisms for passing parameters in URLs - query parameters and fragments. One or the other should suffice in all cases.

Even if it's necessary to do something like dedicate particular query and/or fragment names for DID purposes, that would arguably be preferable to introducing yet a third parameter passing mechanism that requires non-standard URL parsing to use.

I'd taken an action item during a recent WG call to file an issue to drive discussion on removing or defining the use matrix parameters.  This is that issue.

I'll also note that the specification currently appears to be inconsistent on whether matrix parameters are actually supported or not.  For instance, they're missing from the descriptions of DID portions in https://w3c.github.io/did-core/#terminology, but they're present at in the Generic DID Syntax at https://w3c.github.io/did-core/#generic-did-syntax.  The specification should eventually be made self-consistent in this regard.

peacekeeper:
Thanks for creating this issue. To help with the discussion, I recently shared some materials about matrix parameters in DID URLs:

- Slides, log, and recording from a recent DID Resolution spec weekly meeting where we discussed matrix parameters:
https://github.com/w3c-ccg/meetings/tree/gh-pages/2020-01-16-did-resolution

- My topic paper "Why Matrix Parameters?" for Rebooting-the-Web-of-Trust 10:
https://github.com/WebOfTrustInfo/rwot10-buenosaires/blob/master/topics-and-advance-readings/why-matrix-parameters.md

selfissued:
Note that the specification does not uniformly recognize matrix parameters as a component of a DID URL.  For instance, the definition of DID URL at https://w3c.github.io/did-core/#dfn-did-urls is:

> A DID plus an optional DID path, optional ? character followed by a DID query, and optional # character followed by a DID fragment.

This implies that, if present, any matrix parameters are part of the DID, because they are not enumerated in the list of DID URL components above.  Yet the DID definition at https://w3c.github.io/did-core/#dfn-decentralized-identifiers does not state that matrix parameters are part of the DID.

selfissued:
The intended semantics of this text in the Generic DID Parameter Names section at https://w3c.github.io/did-core/#generic-did-parameter-names is unclear:

[...]

PR #813, opened 2022-01-05
Fix improperly URI Encoded values in DID parameters.

msporny:
The DID parameter values that we had in the specification were never properly URI Encoded per the rules for query parameters in RFC3986 (which is what we normatively reference). This PR fixes that, and is an editorial change (because it only updates examples). We got the normative reference right, but flubbed on the DID parameter examples.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/813.html" title="Last updated on Aug 15, 2024, 6:45 PM UTC (aa14fcd)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/813/d20b7c7...aa14fcd.html" title="Last updated on Aug 15, 2024, 6:45 PM UTC (aa14fcd)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -929,7 +929,7 @@ did:example:123#agent
 
         <pre class="example nohighlight"
              title="A resource external to a DID Document">
-did:example:123?service=agent&amp;relativeRef=%2Fcredentials%23degree
+did:example:123?service=agent&amp;relativeRef=/credentials%23degree
         </pre>
 
         <p class="note"
@@ -963,12 +963,12 @@ identifier for a <a>resource</a>.
 
         <pre class="example nohighlight"
           title="A DID URL with a 'versionTime' DID parameter">
-did:example:123?versionTime=2021-05-10T17%3A00%3A00Z
+did:example:123?versionTime=2021-05-10T17:00:00Z
         </pre>
 
         <pre class="example nohighlight"
           title="A DID URL with a 'service' and a 'relativeRef' DID parameter">
-did:example:123?service=files&amp;relativeRef=%2Fresume.pdf
+did:example:123?service=files&amp;relativeRef=/resume.pdf
         </pre>
 
         <p>
--- a/index.html
+++ b/index.html
@@ -929,7 +929,7 @@ did:example:123#agent
 
         <pre class="example nohighlight"
              title="A resource external to a DID Document">
-did:example:123?service=agent&amp;relativeRef=/credentials#degree
+did:example:123?service=agent&amp;relativeRef=%2Fcredentials%23degree
         </pre>
 
         <p class="note"
@@ -963,12 +963,12 @@ identifier for a <a>resource</a>.
 
         <pre class="example nohighlight"
           title="A DID URL with a 'versionTime' DID parameter">
-did:example:123?versionTime=2021-05-10T17:00:00Z
+did:example:123?versionTime=2021-05-10T17%3A00%3A00Z
         </pre>
 
         <pre class="example nohighlight"
           title="A DID URL with a 'service' and a 'relativeRef' DID parameter">
-did:example:123?service=files&amp;relativeRef=/resume.pdf
+did:example:123?service=files&amp;relativeRef=%2Fresume.pdf
         </pre>
 
         <p>


brentzundel (approved):

OR13 (approved):

TallTed (approved):

dlongley (approved):

peacekeeper (changes requested):
Which "rules for query parameters in RFC3986" are you referring to?

I think the examples are correct. `/` and `:` are both allowed characters in a URI's query component and therefore don't have to be percent-encoded:

```
 query       = *( pchar / "/" / "?" )
 pchar       = unreserved / pct-encoded / sub-delims / ":" / "@"
```

Also see https://datatracker.ietf.org/doc/html/rfc3986#section-3.4.

With regard to this example:

```
did:example:123?service=agent&amp;relativeRef=/credentials#degree
```

Note that this appears in the section about the [DID fragment](https://www.w3.org/TR/did-core/#fragment). This is because `#degree` is the DID fragment and not part of the value of the `relativeRef` query parameter. Therefore this example is correct as well.

dlongley:
@peacekeeper,

Per RFC 3986, wherever characters in a URI component would cause a conflict with other characters used as delimiters, the conflicting data must be percent-encoded. In the example you highlighted, we have a `/` character appearing inside of a query component, which conflicts with the use of `/` to delimit the path. Therefore, the `/` character inside of the query must be percent-encoded.

See [Section 2.2](https://datatracker.ietf.org/doc/html/rfc3986#section-2.2), specifically (emphasis mine):

> URIs include components and subcomponents that are delimited by
   characters in the "reserved" set.  These characters are called
   "reserved" because they may (or may not) be defined as delimiters by
   the generic syntax, by each scheme-specific syntax, or by the
   implementation-specific syntax of a URI's dereferencing algorithm.
   ***If data for a URI component would conflict with a reserved
   character's purpose as a delimiter, then the conflicting data must be
   percent-encoded before the URI is formed.***

See also: https://stackoverflow.com/a/6017928/399274

Now, regarding the `#degree` in the example, if it is not percent-encoded, then it must be interpreted as a reference into the DID document. If it is percent-encoded, then it must be interpreted as a reference into the external resource (that would be retrieved via a process involving the rest of the `relativeRef` value). Based on the naming of the fragment here, I think the intent was for it to be a reference into the external resource, so it should be percent-encoded.

peacekeeper:
> wherever characters in a URI component would cause a conflict with other characters used as delimiters, the conflicting data must be percent-encoded

Yes, but a `/` inside a query component does not cause a conflict, because in that component it is not used as a delimiter. [Section 3.4.](https://datatracker.ietf.org/doc/html/rfc3986#section-3.4) explicitly says that `/` is valid in the query component. And [Section 2.1](https://datatracker.ietf.org/doc/html/rfc3986#section-2.1) says:

>    A percent-encoding mechanism is used to represent a data octet in a
>    component when that octet's corresponding character is outside the
>    allowed set or is being used as a delimiter of, or within, the
>    component.

In the case of query, the `/` is neither outside the allowed set, nor is it used as a delimiter of, or within, the component.

> Now, regarding the #degree in the example, if it is not percent-encoded, then it must be interpreted as a reference into the DID document. If it is percent-encoded, then it must be interpreted as a reference into the external resource

No that's not how it works. The fragment is used as a reference to something that is part of, or related to, the result of dereferencing the primary resource (without fragment). That primary resource may be the DID document, or it may be something else, such as an external resource. The `relativeRef` parameter is only one example where DID URLs can be dereferenced to something other than the DID document. Another example is the `resource` parameter. It is not up to us to "redefine" how fragments in URIs work.

dlongley:
Additionally, we would be asking for pain to deviate from common tooling here:

```js
const params = new URLSearchParams({relativeRef: '/test'});
console.log(params.toString());
```

Prints:

> 'relativeRef=%2Ftest'

Almost invariably, whenever people build query strings (especially using common tooling), they URI encode the component parts. This is inline with the interpretation of RFC 3986 I presented above.

TallTed:
@dlongley -- I'm afraid that common tooling is in error.

A `/` is parsed as part of the path segment *until* the first un-encoded `?`, which starts the query segment, after which point, `/` is no longer interpreted as part of the path segment.

Yes, these strings are confusing to humans, and when humans write software, because humans do a terrible job of internalizing the rules, they may persist their misunderstandings.

Please see [`3.4.  Query`](https://datatracker.ietf.org/doc/html/rfc3986#section-3.4) which explicitly states this --
```pre
      query       = *( pchar / "/" / "?" )
```
-- and incorporates these by reference --
```
      pchar         = unreserved / pct-encoded / sub-delims / ":" / "@"
      pct-encoded   = "%" HEXDIG HEXDIG
      unreserved    = ALPHA / DIGIT / "-" / "." / "_" / "~"
      sub-delims    = "!" / "$" / "&" / "'" / "(" / ")"
                    / "*" / "+" / "," / ";" / "="
```

The un-encoded solidus `/` is valid within the query and fragment components, which follow the first un-encoded `?` and `#`, respectively.

dlongley:
@TallTed,

In re-reading the relevant sections of RFC3986 again, I think both the percent-encoded and non-percent-encoded versions of the example are legal. Of course, whether the fragment part is encoded or not will change its interpretation, specifically, whether or not it is part of the `relativeRef` value. I recommend we offer both examples to resolve this.

dlongley:
As for the history here, this PR added `relativeRef`: https://github.com/w3c/did-core/issues/349

From Markus (emphasis mine):

> "Since matrix parameters were removed from the DID URL syntax, one consequence is that an extra DID URL parameter is needed that can contain an ***encoded*** service endpoint path and query string."

The example there clearly encodes the `/` and `#` characters. That PR also links to my comment here with more exposition:

https://github.com/w3c/did-core/issues/159#issuecomment-598286814

peacekeeper:
I agree with @dlongley that both an encoded and unencoded value of the `relativeRef` parameter would be valid. You're right that I have also used encoded examples a few times, see e.g.:

https://w3c-ccg.github.io/did-resolution/#example-5

And I can understand why it would be preferable to encode some characters such as `/` even if they don't have to be encoded. So I think I wouldn't strongly oppose making that change. I'm a bit worried however:

1. That this might be a breaking change, since we only referenced RFC3986 Section 2.1 and didn't further specify exactly which characters need to be encoded: https://www.w3.org/TR/did-core/#did-parameters
2. This might complicate things if we ever embed a DID URL inside e.g. an HTTP URL, then some characters may get double-percent-encoded, or there may be ambiguity on which level of the nested URLs a character was percent-encoded. But I guess that could be solved.

Regarding the fragment in the `relativeRef` case, I feel pretty strongly that this should not be percent-encoded, even though @dlongley is right that I also used examples of percent-encoded fragments in the past. The example I linked above at the beginning of this message shows this:

```
did:example:1234?service=files&relativeRef=%2Fmyresume%2Fdoc%3Fversion%3Dlatest#intro
```

The reason why I really prefer this is that processing of the fragment should happen after the primary dereferencing step. After processing `service` and `relativeRef`, the constructed service endpoint URL "inherits" the fragment, similar to how that works in HTTP redirects.

TallTed:
Unfortunately, encoding characters when they shouldn't be *shouldn't* produce the same result as leaving them unencoded in that location.  `%2F` is not the same as `/`.  Dereferencing a URL containing a parameter value of `%2Ftest` (where the actual target is `/test`) should result in a `4xx`.  Tragically, these errors are common enough that many implementers will check whether parameter values seem to be encoded, and deliver the results that are obtained after decoding.  Postel's robustness principle is the usual excuse.  I think that even if delivering payloads on that basis, some feedback should be provided that the encoding is a wrongness that should be corrected in future requests for that resource.

msporny:
@dlongley @peacekeeper @TallTed ... dusting this PR off after 2.5 years. :)

I think the request here is to show two versions of these URIs, one that is URIEncoded and one that is "raw", and say that both versions are valid? Am I getting that right? If so, this PR will have to change before it's merged.

TallTed:
Both versions are *not* correct, though they might both deliver the desired result in many tools (see: [Postel's Law](https://en.wikipedia.org/wiki/Robustness_principle)); only one version is correct.

`%2f` should only be used where `/` is a separator (e.g., within the path segment); not within the query (following first `?`) nor fragment (following first `#`) segments.

`%23` should only be used where `#` is found within the path (preceding first `?`) or query (preceding first `#`) segments; not when `#` is being used as the fragment delimiter.

Furthermore, important for many of the participants in this thread, In the world of RDF, URIs *SHOULD* (it might be *MUST* somewhere I'm not finding just now) have no unnecessarily encoded entities.

These strings have the extra fun that HTML reserved characters (e.g., `&`) must be HTML-escaped (e.g., `&amp;`).

Finally, it's worth noting that the resource targeted with `relativeRef=/...` is found relative to the *host server's root directory*. If it's intended to be found relative to the document in which the URI is found, it should be `relativeRef=./...`.

Correct if `#degree` represents the DID fragment: `did:example:123?service=agent&amp;relativeRef=/credentials#degree`
Incorrect: `did:example:123?service=agent&amp;relativeRef=/credentials#degree`

Correct if `#degree` is part of the path to the target resource: `did:example:123?service=agent&amp;relativeRef=/credentials%23degree`
Incorrect: `did:example:123?service=agent&amp;relativeRef=%2Fcredentials%23degree`

Correct: `did:example:123?versionTime=2021-05-10T17:00:00Z`
Incorrect: `did:example:123?versionTime=2021-05-10T17%3A00%3A00Z`

Correct: `did:example:123?service=files&amp;relativeRef=/resume.pdf`
Incorrect: `did:example:123?service=files&amp;relativeRef=%2Fresume.pdf`

dlongley:
We will want to check each of these against standard tools like those provided in JS:

> Correct: `did:example:123?versionTime=2021-05-10T17:00:00Z`
> Incorrect: `did:example:123?versionTime=2021-05-10T17%3A00%3A00Z`

Running this in node.js (and Chrome generates the same thing, likely so do other browsers):

```
> x = new URLSearchParams({versionTime: '2021-05-10T17:00:00Z'})
URLSearchParams { 'versionTime' => '2021-05-10T17:00:00Z' }
> x.toString()
'versionTime=2021-05-10T17%3A00%3A00Z'
```

This implies that the above is reversed.

TallTed:
>> Correct: `did:example:123?versionTime=2021-05-10T17:00:00Z`
>> Incorrect: `did:example:123?versionTime=2021-05-10T17%3A00%3A00Z`
>
> Running this in `node.js` (and Chrome generates the same thing, likely so do other browsers):
> ```
> x = new URLSearchParams({versionTime: '2021-05-10T17:00:00Z'})
> URLSearchParams { 'versionTime' => '2021-05-10T17:00:00Z' }
> x.toString()
> 'versionTime=2021-05-10T17%3A00%3A00Z'
> ```
>
> This implies that the above is reversed.

There's a widespread _incorrect_ belief that colons are disallowed in most if not all URL segments *except* the authority (where they're necessary to delimit the `host:port`), so many tools that output URLs URL encode all colons.

A careful review of the EBNF (which I've performed multiple times in the *decades* I've been arguing this one; see [*above in this thread, from January 2022*](https://github.com/w3c/did-core/pull/813#issuecomment-1009074373) for a quotation of the relevant segment) will confirm my assertion of the correct/incorrect quoted above. 

These colons are in the query segment which is partially defined as —
```
query       = *( pchar / "/" / "?" )
...
pchar         = unreserved / pct-encoded / sub-delims / ":" / "@" 
```
Note the explicit `":"` in the `pchar` expansion.

dlongley:
@TallTed,

> There's a widespread incorrect belief that colons are disallowed in most if not all URL segments except the authority (where they're necessary to delimit the `host:port`), so many tools that output URLs URL encode all colons.

To be clear, I'm not saying that you're incorrect about the interpretation of the grammar rules. I think you're right about this widespread incorrect belief wrt. RFC 3986. However, I think because it's so widespread and that everyone is using extremely common and widespread tooling that works the way it does (like WHATWG URL documents). We shouldn't and needn't fight a battle against it, a battle I'd expect us to quickly run out of steam fighting and lose.

The WHATWG spec, for better or worse (we don't need to debate those merits), was written to document common behavior (albeit "widespread incorrect belief") and cover more consistent ways to design tooling and make it compositional and reusable and so on. People followed this spec, if they weren't doing so already, and that's what the standard tools do now. I think it would be a mistake for us to say that people have to write new tools / can't use these in order to work with the DID spec.

So, I think we'll need to find a path through this that works with these common tools.

emg110:
I just wanted to mention and remind everyone about the first mission of the standards documents body and examples within that is to be human readable and therefore any URI formatting (less human-readable) of examples does not sound productive, IMHO. This PR does makes the DID-Core document and examples in it less human-readable, therefore It decreases the adoption and steepens the learning curve IMHO. 
Screenshot 2024-08-04 at 11 57 20

emg110:
Like the address bar content in browsers it will eventually be RFC3986 formatted in browsers or when sent by clients (no matter what software platform), IMHO. RFC3986 is mature and woven in http communications for the most, so in higehr level protocols and standards , IMHO, there is no need to re-implement that and just mentioning it as eventual format for URI fields and params would suffice the case.

decentralgabe:
Discussed on today's call. @TallTed please suggest what does not need to be encoded so that we can move forward with this PR.

index.html
```diff
@@ -963,12 +963,12 @@ <h2>DID Parameters</h2>
 
         <pre class="example nohighlight"
           title="A DID URL with a 'versionTime' DID parameter">
-did:example:123?versionTime=2021-05-10T17:00:00Z
+did:example:123?versionTime=2021-05-10T17%3A00%3A00Z
```
TallTed, line 966: don't need to encode colons in the query segment
```suggestion
did:example:123?versionTime=2021-05-10T17:00:00Z
```

index.html
```diff
@@ -963,12 +963,12 @@ <h2>DID Parameters</h2>
...
         </pre>
 
         <pre class="example nohighlight"
           title="A DID URL with a 'service' and a 'relativeRef' DID parameter">
-did:example:123?service=files&amp;relativeRef=/resume.pdf
+did:example:123?service=files&amp;relativeRef=%2Fresume.pdf
```
TallTed, line 971: Don't need to encode `/` in the query segment.
```suggestion
did:example:123?service=files&amp;relativeRef=/resume.pdf
```

index.html
```diff
@@ -929,7 +929,7 @@ <h2>Fragment</h2>
 
         <pre class="example nohighlight"
              title="A resource external to a DID Document">
-did:example:123?service=agent&amp;relativeRef=/credentials#degree
+did:example:123?service=agent&amp;relativeRef=%2Fcredentials%23degree
```
TallTed, line 932: Don't need to encode `/` in the query segment. 

Do need to encode `#` within the query segment (i.e., when it's not being the fragment delimiter).
```suggestion
did:example:123?service=agent&amp;relativeRef=/credentials%23degree
```

jandrieu:
+1 to @TallTed's adjustments.

However, I think the ampersand also should not be escaped:

Consider two interpretations. First, that the ampersand SHOULD be
```did:example:123?service=agent&amp;relativeRef=/credentials%23degree
In that case, the service is "agent&relativeRef=/credentials#degree", but this is probably incorrect.

Second, that the ampersand SHOULD NOT be
```did:example:123?service=agent&relativeRef=/credentials%23degree```
In this case the service query term = "agent" and the relativeRef query term is /credentials#degree

I think this second interpretation is what we are looking for. That is, the base URL is 
`did:example:123` the first query term is `service=agent` and the second query term is relativeRef=`/credentials#degree`

So, I think the edits should be ```did:example:123?service=agent&relativeRef=/credentials%23degree```

TallTed:
> However, I think the ampersand also should not be escaped:

[As covered earlier](https://github.com/w3c/did-core/pull/813#issuecomment-2223694999) in this rather long thread, the ampersand is HTML encoded as `&amp;` *for rendering by an HTML web browser*.  It is not URL encoded (which would be `%26`).

jandrieu:
> [As covered earlier](https://github.com/w3c/did-core/pull/813#issuecomment-2223694999) in this rather long thread, the ampersand is HTML encoded as `&amp;` _for rendering by an HTML web browser_. It is not URL encoded (which would be `%26`).

I missed that. Thanks for the clarification.

msporny:

pchampin:
This was discussed during the WG meeting on 2024-08-22:
https://www.w3.org/2024/08/22-did-minutes.html#t05

Issue #819, opened 2022-02-28
Controller confusion


mwherman2000:
In https://www.w3.org/TR/did-core/#verification-method-properties, it says that `controller` is _required_. Clicking on `controller` in this table links to https://www.w3.org/TR/did-core/#did-controller ...which says _controller_ is _optional_.  These need to be differentiated or, at least, unlinked from one another.

kdenhartog:
The controller property defined in https://www.w3.org/TR/did-core/#did-controller is for establishing the controller of the DID Document. The `controller` property defined in https://www.w3.org/TR/did-core/#verification-method-properties is about the controller of the verification method. They're two separate properties which follow different semantics because they're located in different objects of the JSON.

TallTed:
The link from the `controller` in the [Verification Method Properties](https://www.w3.org/TR/did-core/#verification-method-properties) table does need to change, from https://www.w3.org/TR/did-core/#dfn-controller (not from https://www.w3.org/TR/did-core/#did-controller as @mwherman2000 said, though `did-controller` is the fragment ID of the section header for the section which contains the `dfn-controller` fragment).

The current target of the link is within [§ 5.1.2 DID Controller](https://www.w3.org/TR/did-core/#did-controller)

The correct target of the link is within [§ 5.2 Verification Methods](https://www.w3.org/TR/did-core/#verification-methods) -- but [this definition](https://www.w3.org/TR/did-core/#:~:text=DID%2DSPEC%2DREGISTRIES%5D.-,controller,-The%20value%20of) does not now have a unique fragment identifier in the HTML.

msporny:
PR #820 has been merged to address this issue.

PR #820, opened 2022-03-01
remove link to wrong controller definition

pchampin:
addresses issue #819 by removing the wrong link from `VerificationMethod::controller` to the definition of `DIDDocument::controller`.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/820.html" title="Last updated on Mar 2, 2022, 11:57 AM UTC (decb9e8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/820/d20b7c7...decb9e8.html" title="Last updated on Mar 2, 2022, 11:57 AM UTC (decb9e8)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1512,7 +1512,7 @@ A <a data-cite="INFRA#string">string</a> that conforms to the rules in
             </td>
           </tr>
           <tr>
-            <td><code><a>controller</a></code></td>
+            <td><code>controller</code></td>
             <td>yes</td>
             <td>
 A <a data-cite="INFRA#string">string</a> that conforms to the rules in
--- a/index.html
+++ b/index.html
@@ -1504,7 +1504,7 @@ A <a data-cite="INFRA#ordered-set">set</a> of <a>Service Endpoint</a>
         </thead>
         <tbody>
           <tr>
-            <td><code>id</code></td>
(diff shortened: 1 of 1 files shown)


TallTed:
While this will remove the immediate cause of the confusion flagged in #819, I don't think it's the best nor a complete solution, not least because there remains a link from `controller` (line 1431) in the table preceding that where the link from `controller` (line 1515) is being removed, which may lead other readers to think the second refers to the same definition as the first (following the principle of linking from first occurrence), which is basically the same confusion as exists now.

Is there no way to make each of these two table occurrences link to their appropriate, different `dfn`?

pchampin:
I went for the minimal change, noticing that not all properties in the table are links anyway.

It would of course be possible to add a link to the correct definition of `controller`, but then why not add such a link for all properties in the table... (if only for the sake of consistency). If the editors think it is appropriate, I can extend this PR.

TallTed:
I hadn't realized that only some of the terms in the (informative! a/k/a non-normative!) tables are linked to their (normative!) definitions.  To my mind, because of the non-normative status of those tables, all the terms therein should be linked to their normative definitions.

pchampin:
done

msporny (approved):

msporny:
Class 1, editorial, reviewed, no changes requested, no objections, merging.

PR #826, opened 2022-07-16
remove PR-related paragraph from REC

pchampin:
@msporny I also took the liberty of applying this change directly
on the version hosted on w3.org/TR ,
as I notived this paragraph was not on the REC dated 2022-06-30.

I don't know why/how it popped back...


msporny (approved):
Agreed, this text should have been removed before REC. Thanks for the good catch, @pchampin !

PR #840, opened 2023-04-05
Fix link to INFRA#maps in index.html

FabioPinheiro:
Just fixing some links

***
<a href="https://pr-preview.s3.amazonaws.com/FabioPinheiro/did-core/pull/840.html" title="Last updated on Apr 5, 2023, 12:32 AM UTC (2c7886b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/840/e9959f9...FabioPinheiro:2c7886b.html" title="Last updated on Apr 5, 2023, 12:32 AM UTC (2c7886b)">Diff</a>

PR #851, opened 2024-03-05
Fix serviceEndpoints maps link

KendallWeihe:
Similar to https://github.com/w3c/did-core/pull/840

It was unclear to me if these should be maps (as they are) or [ordered maps](https://infra.spec.whatwg.org/#ordered-map)?

### DIFF
--- a/REC/2022-07-19/index.html
+++ b/REC/2022-07-19/index.html
@@ -2305,10 +2305,10 @@ A <a href="https://infra.spec.whatwg.org/#string">string</a> or a
             <td>yes</td>
             <td>
 A <a href="https://infra.spec.whatwg.org/#string">string</a> that conforms to the rules of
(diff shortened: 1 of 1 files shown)


davidlehn (approved):

decentralgabe (approved):

msporny:
This PR attempts to change a date-stamped Recommendation -- we can't do that (change documents published in the past).

Instead, this change has been made in the main spec and the fix will go out in v1.1 (within weeks).

msporny (changes requested):
This PR attempts to change a date-stamped Recommendation -- we can't do that (change documents published in the past). Our charter was just renewed a month ago, and so we're starting to process issues again, but we can't make the kind of change requested by this PR.

@KendallWeihe wrote:
> It was unclear to me if these should be maps (as they are) or [ordered maps](https://infra.spec.whatwg.org/#ordered-map)?

They are just maps, not ordered maps.

If this change hasn't already been made on the main spec, @KendallWeihe, can you please update this PR to do that instead?

decentralgabe:
Closing. Please open a follow up if there remains a problem in the core spec.

Issue #859, opened 2024-08-15
Find a path forward for contexts in v1.1


decentralgabe:
> My guess is we'll probably need a new context for 1.1 anyway

I am not so sure. I would not think we would add new properties for DID Documents (we may not even be allowed to do so), and that would be the only reason to change the context.

_Originally posted by @iherman in https://github.com/w3c/did-core/issues/850#issuecomment-2277216879_

jandrieu:
We should be open to it so that we don't constrain our work.

pchampin:
This was discussed during the did meeting on 2024-08-29:
https://www.w3.org/2024/08/29-did-minutes.html#t05

msporny:
PR #867 has been raised to address this issue. This issue will be closed once PR #867 has been merged.

Issue #850, opened 2024-02-05
Update JsonWebKey2020 to JsonWebKey


decentralgabe:
As per https://w3c.github.io/vc-jose-cose/

peacekeeper:
Makes sense to align with VC specs.

decentralgabe:
As @msporny mentioned today we can either update the existing context or create a new one.
My guess is we'll probably need a new context for 1.1 anyway, so we could have this issue follow that one.
Alternatively, we could create a new examples context and solve this quicker.

I'm fine with either path.

iherman:
> My guess is we'll probably need a new context for 1.1 anyway

I am not so sure. I would not think we would add new properties for DID Documents (we may not even be allowed to do so), and that would be the only reason to change the context.

msporny:
PR https://github.com/w3c/did-core/pull/867 has been raised to address this issue. This issue will be closed once PR https://github.com/w3c/did-core/pull/867 has been merged.

msporny:
PR #867 has been merged, closing.

PR #852, opened 2024-03-21
Use `publicKeyMultibase` instead of `publicKeyBase58` field in all examples

Patrik-Stas:
Hi, I believe that some examples were by mistake using `publicKeyBase58` rather than `publicKeyMultibase` - I've updated the examples. Let me know if I am missing something.

Cheers

***
<a href="https://pr-preview.s3.amazonaws.com/Patrik-Stas/did-core/pull/852.html" title="Last updated on Jul 7, 2024, 7:58 PM UTC (7edae25)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/852/a8ddc80...Patrik-Stas:7edae25.html" title="Last updated on Jul 7, 2024, 7:58 PM UTC (7edae25)">Diff</a>

PR #867, opened 2024-10-30
Add JSON-LD Context for v1.1

msporny:
This PR is an attempt to address issue #859 and #850 by adding a v1.1 JSON-LD Context and updating all examples in the specification to use the new context URL.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/pull/867.html" title="Last updated on Oct 31, 2024, 1:08 AM UTC (1dff914)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-core/867/b405914...1dff914.html" title="Last updated on Oct 31, 2024, 1:08 AM UTC (1dff914)">Diff</a>


contexts/v1.1
```diff
@@ -0,0 +1,103 @@
...
+          "@type": "@json"
+        },
+        "secretKeyJwk": {
+          "@id": "https://w3id.org/security#secretKeyJwk",
+          "@type": "@json"
+        }
+      }
+    }
+  }
+}
```
dlongley, line 103: Updates to match definitions in controller document v1 context:

```suggestion
{
  "@context": {
    "@protected": true,
    "id": "@id",
    "type": "@type",

    "alsoKnownAs": {
      "@id": "https://www.w3.org/ns/activitystreams#alsoKnownAs",
      "@type": "@id"
    },
    "assertionMethod": {
      "@id": "https://w3id.org/security#assertionMethod",
      "@type": "@id",
      "@container": "@set"
    },
    "authentication": {
      "@id": "https://w3id.org/security#authenticationMethod",
      "@type": "@id",
      "@container": "@set"
    },
    "capabilityDelegation": {
      "@id": "https://w3id.org/security#capabilityDelegationMethod",
      "@type": "@id",
      "@container": "@set"
    },
    "capabilityInvocation": {
      "@id": "https://w3id.org/security#capabilityInvocationMethod",
      "@type": "@id",
      "@container": "@set"
    },
    "controller": {
      "@id": "https://w3id.org/security#controller",
      "@type": "@id"
    },
    "keyAgreement": {
      "@id": "https://w3id.org/security#keyAgreementMethod",
      "@type": "@id",
      "@container": "@set"
    },
    "service": {
      "@id": "https://www.w3.org/ns/did#service",
      "@type": "@id",
      "@context": {
        "@protected": true,
        "id": "@id",
        "type": "@type",
        "serviceEndpoint": {
          "@id": "https://www.w3.org/ns/did#serviceEndpoint",
          "@type": "@id"
        }
      }
    },    
    "verificationMethod": {
      "@id": "https://w3id.org/security#verificationMethod",
      "@type": "@id"
    },
    "Multikey": {
      "@id": "https://w3id.org/security#Multikey",
      "@context": {
        "@protected": true,
        "id": "@id",
        "type": "@type",
        "controller": {
          "@id": "https://w3id.org/security#controller",
          "@type": "@id"
        },
        "revoked": {
          "@id": "https://w3id.org/security#revoked",
          "@type": "http://www.w3.org/2001/XMLSchema#dateTime"
        },
        "expires": {
          "@id": "https://w3id.org/security#expiration",
          "@type": "http://www.w3.org/2001/XMLSchema#dateTime"
        },
        "publicKeyMultibase": {
          "@id": "https://w3id.org/security#publicKeyMultibase",
          "@type": "https://w3id.org/security#multibase"
        },
        "secretKeyMultibase": {
          "@id": "https://w3id.org/security#secretKeyMultibase",
          "@type": "https://w3id.org/security#multibase"
        }
      }
    },
    "JsonWebKey": {
      "@id": "https://w3id.org/security#JsonWebKey",
      "@context": {
        "@protected": true,
        "id": "@id",
        "type": "@type",
        "controller": {
          "@id": "https://w3id.org/security#controller",
          "@type": "@id"
        },
        "revoked": {
          "@id": "https://w3id.org/security#revoked",
          "@type": "http://www.w3.org/2001/XMLSchema#dateTime"
        },
        "expires": {
          "@id": "https://w3id.org/security#expiration",
          "@type": "http://www.w3.org/2001/XMLSchema#dateTime"
        },
        "publicKeyJwk": {
          "@id": "https://w3id.org/security#publicKeyJwk",
          "@type": "@json"
        },
        "secretKeyJwk": {
          "@id": "https://w3id.org/security#secretKeyJwk",
          "@type": "@json"
        }
      }
    }
  }
}
```

dlongley: Also, I would recommend not publishing any v1.1 context until CR (at the earliest, and as a final change in CR prior to PR), instead publishing it as `https://www.w3.org/ns/did/v1.1rc1` ... with increasing `rcNNN` numbers as needed. This approach enables implementations to make changes as needed until the standard is stable, without locking potential mistakes into earlier versions of the context, which are meant to be statically loaded from local copies and treated as immutable.

danpape: Looks good, however, I did not check all your publicKey changes in index.html from base58 to MultiBase. Did you do those automatically somehow, or were you manually updating? If manual, can you describe your procedure so I can double check?

danpape: > Looks good, however, I did not check all your publicKey changes in index.html from base58 to MultiBase. Did you do those automatically somehow, or were you manually updating? If manual, can you describe your procedure so I can double check?

Oh, I see they may have come from #852. But in that PR the author just stuck a 'z' on the front of all the keys but in this PR it looks like you may have totally updated some of the values.

decentralgabe (approved):

contexts/v1.1
```diff
@@ -0,0 +1,103 @@
+{
+  "@context": {
+    "@protected": true,
+    "id": "@id",
+    "type": "@type",
+
+    "alsoKnownAs": {
+      "@id": "https://w3id.org/security#alsoKnownAs",
```
decentralgabe, line 8: non blocking, I am struggling to remember where we landed on w3id ... since we have registries in this WG could we migrate what's on w3id to one of the w3c managed registries?

msporny: Here's the latest discussion on w3id.org for the controller document: https://github.com/w3c/controller-document/issues/66#issuecomment-2311114405

Note that this URL https://w3id.org/security redirects to the security vocabulary maintained by the VCWG for the Data Integrity specifications. The only thing on w3id.org are redirect instructions to the VCWG document:

https://github.com/perma-id/w3id.org/blob/master/security/.htaccess

We need to add text to the following effect when we make the DID Core spec depend on the controller doc spec (effectively, "the hash of the base context is X... don't ever load it from the network"):

https://w3c.github.io/controller-document/#json-ld-context

Does that answer  your question @decentralgabe ? We've got a few more PRs we need to do to align the guidance in the controller document and DID Core (and hope to do those alignments in future PRs after we've based DID Core on the controller doc spec).

contexts/v1.1
msporny: @dlongley, weird, I copied the context from the controller doc spec directly. I'll need to go back through your changes to see where the misalignment was. Ack on using rc1..rcN, I'll make that change right after this comment.

@danpape, I copied the public keys from the controller document specification, which I remember generating from code. The algorithm for generating that stuff is documented here:

https://w3c.github.io/controller-document/#Multikey

iherman (approved):
I am a little bit worried by the proliferation of somewhat identical context files. We are clearly bypassing the DRY principle.

Specifically: by using the [Context import of JSON-LD](https://www.w3.org/TR/json-ld11/#imported-contexts), we could import, e.g., https://w3id.org/security/multikey/v1 or https://w3id.org/security/jwk/v1 and avoid repeating the information in this context file.

I realize that this may complicate the processing steps from a security and efficiency point of view (regarding caching, integrity checking, etc). But this approach may lead to maintenance issues down the line.

I approve the PR insofar as this is not a blocking comment.

contexts/v1.1
iherman: @decentralgabe, the issue was raised (several times) and, at the moment, there is no change (in agreement with the w3c management on the DI CR).

msporny:
@iherman wrote:
> I realize that this may complicate the processing steps from a security and efficiency point of view (regarding caching, integrity checking, etc). But this approach may lead to maintenance issues down the line.

Yes, that's the problem that was raised as a "security concern" a year or more ago, that imported contexts might be loaded from the network or trigger "extra processing" that "might be dangerous". All of that was analyzed as FUD -- if you cache contexts permanently and never load them from the network, then this isn't an issue. That said, people will still complain about "the complexity of not knowing what is being imported", which again is FUD, but it's easier to just repeat the declarations than import them (and have to deal w/ the FUD). 

There are two benefits in including all content in the context file:

* A human analyzing it to ensure it has what you think it has is easier, and 
* There is a slight performance boost by having everything local (though, it's negligible)

To put it another way, if we imported contexts, I expect that we'd see objections (again).

contexts/v1.1
msporny: > there is no change (in agreement with the w3c management on the DI CR).

Meaning, W3C Management didn't see an issue with the way w3id.org is operated today and don't see a reason that W3C needs to take over management of the website. This is because we tell people to not load these contexts from the network, provide static copies, and provide hashes for the contexts in the specs.

iherman:
> To put it another way, if we imported contexts, I expect that we'd see objections (again).

Ok, this is probably not worth the trouble. The is on us (maintainers of the spec), but we will live.

contexts/v1.1
decentralgabe: Thanks, yes that addresses my comment. The language you noted will be useful to add.

pchampin:
This was discussed during the [did meeting on 14 November 2024](https://www.w3.org/2024/11/14-did-minutes.html#7eb7).

<h4><a href="https://github.com/w3c/did-core/pull/867" rel="noopener noreferrer">w3c/<wbr>did-core#867</a></h4>
<p><cite>manu:</cite> Media types are an issue. Proposal of 3 part media types was not accepted.<br>
<span>… New context for the JSON-LD context to align with Controller Document</span></p>

<hr /></details>

TallTed (commented):
lgtm, fwiw

danpape (approved):
No issues from me.

msporny:
@dlongley wrote:
> Also, I would recommend not publishing any v1.1 context until CR (at the earliest, and as a final change in CR prior to PR), instead publishing it as https://www.w3.org/ns/did/v1.1rc1 ... with increasing rcNNN numbers as needed.

Fixed in 1dff9143decd20cdb0f085aac3c1554f410cac40.

PR #890, opened 2025-04-17
Add missing diagram source files.

peacekeeper:
@msporny Hope this is useful.


peacekeeper:
For exporting from LibreOffice to SVG, I found it necessary to do the following:

1. **Select all objects on a page.** (i.e. NOT select the whole page itself)
2. File -> Export, myfile.svg, and **check the "Selection" checkbox**, then click "Export"

This worked for me, as opposed to just exporting an entire page as a whole.

msporny (approved):

msporny:
Purely editorial, merging.

Issue #865, opened 2024-09-25
Question about query parameter ordering


SaulMoonves:
I am looking for guidance on query parameter ordering, specifically where a DID URL with a query parameter might be used as an ID

In many systems such as ActivityPub, an ID is basically an opaque string but in practice it is a URL whose structure gives a hint on how to resolve it, like using an HTTPS URL that resolves to the object.

I am in the process of trying to implement `did:web` for activitystreams object IDs inside of an ActivityPub project, but this would I think equally apply to service IDs in the DID document: How do I deal with query parameters being seemingly unordered if some things rely on IDs equally matching? Is anyone dealing with this in their projects?

Peeja:
(Someone with more knowledge should correct me if I'm wrong, but…)

A _DID_ can't contain query parameters. A _DID URL_ can. Only the _DID_ is the identifier. The DID URL refers to a resource. The DID is the [authority](https://datatracker.ietf.org/doc/html/rfc3986#section-3.2) of the DID URL. (That statement may not be exactly technically correct, but should be close to correct.)

So: for authentication purposes, you shouldn't be using identifiers which include query params. And if you really need to know if two DID _URL_s point to the same resource, then you've got the same problem you have with the equivalent HTTP URLs, and you'll have to decide how you want to solve that.

But it sounds like you're just talking about identifiers, in which case I believe you're going to have to rethink how your DIDs are produced to include whatever you're currently using query params for.

msporny:
Yes, effectively what @Peeja said. :)

If you are looking for id equivalence matching, using DID URLs with multiple query parameters is probably not the right thing to do. If you need to do that, your application/specification will have to contain some query parameter canonicalization rules to ensure equivalence holds in your application area.

If you could give us a more concrete use case, @SaulMoonves, we might be able to provide more guidance.

peacekeeper:
Also agree with @Peeja .

This should be handled in the same way as you would handle query parameters in HTTPS URLs. Would you order parameters in HTTPS URLs? Would you do exact string matching on HTTPS URLs with parameters?

trwnh:
> your application/specification will have to contain some query parameter canonicalization rules to ensure equivalence holds in your application area

If I'm reading this correctly, then you're saying that something like:

```json
{
  "@id": "did:example:abc123?service=foo&relativeRef=/"
}
```

[...]

PR #895, opened 2025-05-31
Add warning about comparing DID URLs containing query parameters.

msporny:
This PR is an attempt to address issue #865 by adding a warning about comparing DID URLs that contain more than a single query parameter.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did/pull/895.html" title="Last updated on Jul 10, 2025, 1:42 PM UTC (c3a8092)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did/895/d0be1f3...c3a8092.html" title="Last updated on Jul 10, 2025, 1:42 PM UTC (c3a8092)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1198,11 +1198,12 @@ did:example:123456?versionId=1
 
         <p class="note" title="Avoid comparison of DID URLs containing
           multiple query parameters">
-Implementers are urged to avoid using [=DID URLs=] that contain more than a
-single query parameter when comparing [=DID URLs=] for equivalence. There are
-no normalization rules defined for query parameters in this specification and
-any query parameter normalization rules defined at the DID Method specification
-or application layer are not universally recognized rules.
+Implementers are urged to avoid comparing [=DID URLs=] for equivalence when
+they have more than one query parameter without a specification specifically
+designed for that purpose. This specification defines no normalization rules
+for query parameters, and any query parameter normalization rules defined at
+the DID Method specification or application layer are not universally
+recognized rules.
         </p>
 
       </section>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1195,6 +1195,16 @@ <h2>Query</h2>
...
 did:example:123456?versionId=1
         </pre>
+
+        <p class="note" title="Avoid comparison of DID URLs containing
+          multiple query parameters">
+Implementers are urged to avoid using [=DID URLs=] that contain more than a
+single query parameter when comparing [=DID URLs=] for equivalence. There are
+no normalization rules defined for query parameters in this specification and
+any query parameter normalization rules defined at the DID Method specification
+or application layer are not universally recognized rules.
```
dlongley, line 1205: I could see another specification handling this for some ecosystem/use case in the future, so I have a slight suggestion. I also had some trouble following what the advice was here, so let me know if the additional rewording helps:

```suggestion
Implementers are urged to avoid comparing [=DID URLs=] with more than
one query parameter for equivalence without a specification specifically
designed for that purpose. There are no normalization rules defined for
query parameters in this specification and any query parameter normalization
rules defined at the DID Method specification or application layer are not
universally recognized rules.
```

TallTed: A light tweak to @dlongley's suggestion...

```suggestion
Implementers are urged to avoid comparing [=DID URLs=] for equivalence when
they have more than one query parameter without a specification specifically
designed for that purpose. This specification defines no normalization rules
for query parameters, and any query parameter normalization rules defined at
the DID Method specification or application layer are not universally
recognized rules.
```

ottomorac (approved):

TallTed (changes requested):
Approved given that https://github.com/w3c/did/pull/895#discussion_r2122009476 is applied

peacekeeper (approved):
@dlongley 's comment about future use cases makes sense to me, and @TallTed 's improvement also looks good, so I'm in favor of applying that and then merging.

Issue #898, opened 2025-06-21
Is percent encoding the fragment (#) really conformant with the URI RFC?


TheButlah:
Noob question here. I was using [example 6](https://www.w3.org/TR/did-1.1/#example-a-resource-external-to-a-did-document) in my test cases for parsing DIDs. The particular library I'm using ([fluent-uri](https://docs.rs/fluent-uri/latest/fluent_uri/index.html)) claims that there is no fragment in the URI - which is also what I would have expected.

To check, I went to [RFC3986 section 3.5](https://www.rfc-editor.org/rfc/rfc3986#section-3.5). Reading it, it says:

>  A fragment identifier component is indicated by the presence of a number sign ("#") character and terminated by the end of the URI.

Wouldn't this example in the DID-Core spec therefore be inconsistent with the RFC?

To test how browsers handle this, I tried the link https://www.w3.org/TR/did-1.1/%23example-a-resource-external-to-a-did-document (note the pct encoded fragment) in firefox and chromium. Neither scroll the page to the link, which to me indicates that like the RFC, they *also* don't consider this percent encoding to be a valid fragment identifier.

Is the DID-Core spec wrong? What should I as an application developer do with pct encoded fragments? Right now I'm probably going to consider them invalid.

w3cbot:
This was discussed during the [#did meeting on 26 June 2025](https://www.w3.org/2025/06/26-did-minutes.html#a253).

<details><summary><i>View the transcript</i></summary>

<h4>Is percent encoding the fragment (#) really conformant with the URI RFC? #898</h4>
<p><cite>&lt;ottomorac&gt;</cite> <a href="https://github.com/w3c/did/issues/898" rel="noopener noreferrer">w3c/<wbr>did#898</a></p>
<p><cite>manu:</cite> example 6 is correct and don't think it goes against RFC.  Changing to a hash mark would be wrong.  The fragment must % encode the fragment.  This is an esoteric example and will look at some improved descriptive text.</p>
<p><cite>manu:</cite> this would be a class 2 change</p>

<hr /></details>

peacekeeper:
We discussed this extensively in https://github.com/w3c/did/pull/813. I think that PR 813 was wrong and should be reverted. I don't think it's a class 2 change to revert that.

In other words, the current example:
`did:example:123?service=agent&relativeRef=/credentials%23degree`

Should be reverted back to:
`did:example:123?service=agent&relativeRef=/credentials#degree`

[...]

PR #904, opened 2025-08-28
Fix DID fragment examples.

msporny:
This PR is an attempt to address issue #898 by removing a confusing fragment example from the spec and fixing some of the fragment naming to make it easier to understand what the fragment is pointing to.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did/pull/904.html" title="Last updated on Aug 28, 2025, 10:18 PM UTC (4b92154)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did/904/3ab8b19...4b92154.html" title="Last updated on Aug 28, 2025, 10:18 PM UTC (4b92154)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1225,17 +1225,12 @@ identifiers are shown below.
 
         <pre class="example nohighlight"
              title="A unique verification method in a DID Document">
-did:example:123#public-key-0
+did:example:123#public-key-1
         </pre>
 
         <pre class="example nohighlight"
              title="A unique service in a DID Document">
-did:example:123#agent
-        </pre>
-
-        <pre class="example nohighlight"
-             title="A resource external to a DID Document">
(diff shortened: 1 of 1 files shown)


peacekeeper:
I guess removing this here is okay, if the relativeRef example is too difficult to understand. But the risk is that readers will think DID fragments are only used to identify something inside the DID document, which is not the case..

msporny:
> I guess removing this here is okay, if the relativeRef example is too difficult to understand. But the risk is that readers will think DID fragments are only used to identify something inside the DID document, which is not the case..

Once we get clarity on https://github.com/w3c/did-resolution/issues/181, I'm happy to put a clearer example back in (or point to an example in DID Resolution on how relativeRef fragments might work.

